PR #2, opened 2015-12-14
Got rid of the What's New text on the Contribute page

LHinson:
Not necessary to include what’s new text. removing redundancy.


LHinson:
@rhamilto CHECK IT!

rhamilto:
Nice job, @LHinson!

PR #19, opened 2016-04-17
Migrated navigation content to patternfly-toolkit.

dlabrecq:
Added vertical navigation example via seamless iframe, instead of linking to an external example (in a new window) as done on patternfly.org.
Added variables for brand icons so patternfly-toolkit may specify its own image path, product title, etc.
Created layout-fixed so vertical navigation may be displayed in a full page.
Removed !important from body padding-top style so vertical navigation is displayed properly.
Added the extra JS required to run examples in pretty print areas.
Fixed small layout issues.


jeff-phillips-18:
LGTM

dtaylor113:
LGTM

PR #21, opened 2016-04-19
Migrated all PF "Communication" Content Over to the Site.

dlabrecq:
This includes inline and toast notification as well as empty state.
Removed gray background surrounding examples and left justified.


dtaylor113:
LGTM

PR #32, opened 2016-04-27
PTNFLY-1391 Sticky footer.

cardosogabriel:
Used this approach: https://getbootstrap.com/examples/sticky-footer/
To achieve the result with pure CSS, I needed to move the <footer> one level up.
Tested on Safari, Chrome and FF for Mac and looks good. 
Did not test on IE. @dlabrecq which version are we supporting?


dlabrecq:
We support whatever Boostrap supports, which is IE 9 and above.

dlabrecq:
Downloaded, built and tested various pages with Firefox, Chrome and Safari -- looks good there.

If you have IE available, please test and let us know your results.

cardosogabriel:
Just downloaded VirtualBox. Trying to set up localhost to work on the VM.

cardosogabriel:
Tested and it works fine with IE 9,10, 11 and Edge.

dlabrecq:
FYI, It turns out that this change broke the home page. That is, the home page will not expand the entire height of the page -- it should look like the old site. It took a while to trace down, but the underlying problem is that the html selector was changed from "height:100%;" to "min-height:100%".

html {
  min-height: 100%; << Doesn't work on home page
  position: relative;
}

My solution is to override the html selector for the home page, only.

PR #41, opened 2016-05-12
Workaround for bug with hidden C3 charts in v0.4.11.

dlabrecq:
Note that an alternative solution is to have patternfly-org depend on C3 v0.4.10.


source/_includes/nav-tabs-pattern.html
```diff
@@ -14,3 +14,21 @@
...
+                    window.dispatchEvent(new Event('resize'));
+                }
+                // Special case for IE
+                if ($(document).fireEvent) {
+                    $(document).fireEvent('onresize');
+                }
+            }, 100);
+        }
+        $(document).on('shown.bs.tab', 'a[data-toggle="tab"]', resize);
+        $(document).ready(resize);
```
jeff-phillips-18: Are both of these necessary?

dlabrecq: Yes because the user can refresh the web browser. In that case, the previously hidden chart will look bad.

jeff-phillips-18: But I would think the shown event would always fire so no real need for the ready event.

dlabrecq:
You're right, it's not needed. I was also testing $('.nav-tabs').click(), which is why I added it. I'll take it out.

source/_includes/nav-tabs-pattern.html
dlabrecq: You're right, it's not needed. I was also testing $('.nav-tabs').click(), which is why I added it. I'll take it out.

PR #72, opened 2016-05-20
Issue-66: Fixing footer issue in homepage.

cardosogabriel:
@dlabrecq , the position relative attributed to the body in the homepage was breaking the way the footer fixes at the bottom. By removing it, footer works normally in this page.
I checked the homepage and didn't find any regression issues. It is good if you double check. Not sure what is the purpose of that class.


dlabrecq:
LGTM

Issue #105, opened 2016-06-03
Forms page modal example


andresgalante:
Remove the stye tag from the example. It was just meant to make the modal visible not to be part of the snippet. We don't want any user to use that style on modal in any case.

dlabrecq:
We can remove it. Thought it was needed to show a modal.

PR #116, opened 2016-06-03
Removes css from modal example

andresgalante:
@dlabrecq It solves #105 
We don't need to show the css, its just for our website to show the modal. This snippet is not meant to be use in production.


dlabrecq:
LGTM

PR #196, opened 2016-06-23
Update Overview and Color Palette page to match new design

kybaker:
@andresgalante @jeff-phillips-18 @dlabrecq - I updated the Overview page and the Color Palettes page under Styles to match the updated designs. Please note that all of the styles are inline which is consistent with the current format. Andres suggested we come back and move all of the new and old styling to the .less file. Please approve this request.


jeff-phillips-18:
Looks ok to me, though I did not try to build and run.

Please squash your commits, ping me if you need hand doing so.

srambach:
+1 on removing inline styles someday.

PR #203, opened 2016-07-07
Travis build setup to commit generated files to a travis-build branch.

dlabrecq:
Note that the publish.sh script depends on properties (i.e., AUTH_TOKEN and TARGET_BRANCH) being set via Travis CI configuration.


dlabrecq:
@dgutride @priley86 and/or @dtaylor113 please review.

publish.sh
```diff
@@ -0,0 +1,18 @@
+#/bin/sh
+
+# User info
+git config --global user.name "Admin"
+git config --global user.email "patternfly@redhat.com"
+
+# Add upstream authentication token
+git remote add upstream https://$AUTH_TOKEN@github.com/dlabrecq/patternfly-org.git
```
dgutride: Will this change to @github.com/patternfly/patternfly-org.git or do we need a bit more testing?

dlabrecq: Ah, yes. I need to change that. Thanks!

dgutride:
Looks good

dtaylor113:
LGTM

priley86:
@dlabrecq  - per our conversation we can also git add `_site` in `publish.sh`. I was testing this out in gh-pages and we will also need this directory. 

Looks good to me otherwise. Good job!

dlabrecq:
FYI, it is not necessary to add the _site directory. This directory built as part of OpenShift hot deploy and output directly to the php server. If we decided to use gh-pages, we can consider including it. For now, I think we should omit it.

PR #211, opened 2016-07-08
Deploying Travis build directly to Openshift repo for hot deploy.

dlabrecq:
In order for hot deploy to work, we must push to the OpenShift repo instead of creating a separate "releases" directory. OpenShift is only checking its own repo, not GitHub.


dlabrecq:
FYI @dgutride.

PR #218, opened 2016-07-14
Modified paths to patternfly widgets.

dlabrecq:
Patternfly changed the tests-src directory. The widgets directory is now located in tests/pages/_includes/widgets.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -17,7 +17,7 @@
     "animate.css": "~3.4.0",
     "clipboard": "~1.5.5",
     "jquery.scrollTo": "~2.1.2",
-    "patternfly": "3.6.0",
+    "patternfly": "f012b78c7f90fada15718705a6d99f72d7060d5c",
     "wowjs": "~1.1.2"
   }
 }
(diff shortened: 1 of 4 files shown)


dlabrecq:
Fairly simple change to get the build working with the latest Patternfly changes.

@dtaylor113, @andresgalante and/or @priley86 please review/merge.

jeff-phillips-18:
LGTM

bower.json
```diff
@@ -17,7 +17,7 @@
     "animate.css": "~3.4.0",
     "clipboard": "~1.5.5",
     "jquery.scrollTo": "~2.1.2",
-    "patternfly": "3.6.0",
+    "patternfly": "f012b78c7f90fada15718705a6d99f72d7060d5c",
     "wowjs": "~1.1.2"
```
priley86: @dlabrecq So is the idea to target a commit until we do another release?

dlabrecq: Right. These changes will cause the build to fail without the latest patternfly, so we either need to create a tag or reference a commit. We don't want to wait until the end of the sprint to cut a new patternfly release. Personally, I prefer a commit because we don't need to create anything new.

dlabrecq: When we cut a new release, at the end of the sprint, this will be replaced with 3.6.1 or whatever tag we decide to create

priley86: Sounds good! LGTM

PR #225, opened 2016-07-23
Cards no data available

LHinson:
Image and description added per request here:
https://github.com/cshinn/patternfly-design/blob/master/pattern_library/
Card/card.md


dlabrecq:
Looks good, but there are too many commits in this PR. Each PR typically has one commit so it's clear what needs to be reviewed -- there are three here. You also have a merge conflict.

To avoid merge conflicts, it's always best to pull/sync the latest source from the upstream repo. I believe SourcTree may help here.

PR #232, opened 2016-07-27
Create a Getting Started guide from Andres blog post.

mindreeper2420:
This is a new section of patternfly.org for those wanting to utilize the project.

screen shot 2016-07-27 at 4 42 39 pm


srambach:
I thought we needed to pull in both patternfly.less and patternfly-additions.less because of an IE limitation on file size. Is that no longer true?

LHinson:
- Instead of "Heads Up", we should probably just say "Note:" I don't think this should be all caps. 
- I know patternfly uses pretty print. Is that applicable here? 
- The page title and the item in the nav should match. "Setting Up PatternFly" might be a more succinct way of saying this. 
- Might be nice to include steps (e.g. Step One: Create a Project)
- @kybaker will you help review this from a visual design perspective? 
- looping in @bleathem for review as well. 
- Is there another screenshot so I can review the rest of it @mindreeper2420?

source/get-started/setup/index.md
```diff
@@ -0,0 +1,203 @@
...
+    <h2>Add Styles</h2>
+    <div class="row">
+      <div class="col-xs-12 col-sm-12 col-md-12">
+        <div class="well col-md-12">
+          <b>HEADS UP</b> PatternFly is based off of <a href="https://www.getbootstrap.com" target="top">Bootstrap</a> and it's use of the <a href="http://lesscss.org/" target="top">LESS</a> pre-processor.
+          If you prefer to use PatternFly with <a href="http://sass-lang.com/" target="top">SASS</a>, you can find that <a href="https://github.com/patternfly/patternfly-sass" target="top">here</a>.
+        </div>
+      </div>
+      <div class="col-xs-12 col-sm-12 col-md-12">
+        <p>Now create two folders - one for /less and one for /css.</p>
```
dlabrecq: When you say "/less" and "/css", that suggests the root file system. I think this should be something more like <install path>/less, for example?

source/get-started/setup/index.md
```diff
@@ -0,0 +1,203 @@
...
+  <div class="col-md-12">
+    <h2>Create A Project</h2>
+    <div class="row">
+      <div class="col-xs-12 col-sm-12 col-md-6">
+        <h3>Using Finder</h3>
+        <ul>
+          <li>Launch a new Finder window</li>
+          <li>Navigate to where you want to project to live</li>
+          <li>File -> New Folder (or Ctrl + Shift + N)</li>
+          <li>Name the folder</li>
```
dlabrecq: The create a project section only shows how to create a directory. It may be misleading to call this a project?

mindreeper2420:
@srambach I'm not sure about the patternfly-additions.less file, but I will look into that. @LHinson - I've attached full screenshots of the updated work.
screen shot 2016-07-28 at 10 04 59 am
screen shot 2016-07-28 at 10 05 10 am
screen shot 2016-07-28 at 10 05 18 am
screen shot 2016-07-28 at 10 05 30 am
screen shot 2016-07-28 at 10 05 41 am
screen shot 2016-07-28 at 10 05 50 am

source/get-started/setup/index.md
mindreeper2420: @dlabrecq How about Create A Project Directory?

dlabrecq:
The HTML file appears to be missing quite a few JavaScript files. In addition to patternfly.min.js, developers will need to include jQuery, Bootstrap, etc. Depending on what component they plan to use, they may also need C3, D3, and more.

I believe it would be best to show a small component example in the HTML. Just adding the CSS isn't going to get you very far.

source/get-started/setup/index.md
dlabrecq: That would be fine. Although, we probably don't need to show multiple variations of how to create a directory?

source/get-started/setup/index.md
```diff
@@ -0,0 +1,202 @@
...
+<div class="row">
+  <div class="col-md-12">
+    <h2>Step 2: Use Bower To Install PatternFly</h2>
+    <div class="row">
+      <div class="col-xs-12 col-sm-12 col-md-12">
+        <div class="well">
+          <b>Note:</b> Bower is a package manager. We are going to use Bower to install the PatternFly framework.
+        </div>
+        <p>The first thing needs to be done is to install Bower globally. This is done using the command <kbd>npm install -g bower</kbd></p>
+        <p>Occasionally, you will have to run this command as an Administrator. To do so, precede <kbd>npm install -g bower</kbd> with <kbd>sudo</kbd> and run the command again.</p>
```
dlabrecq: Might want to show the command as "sudo npm install -g bower" just to be clear.

source/get-started/setup/index.md
mindreeper2420: I believe the only reason for showing both is for those that prefer command line over GUI. Although, the GUI method is only listed for MacOS... @LHinson do we need to show GUI guides, or can we just stick with the command line? If so for GUI guides, should we have more than just MacOS?

dlabrecq: I don't feel strongly about it. It's just odd to see so much space devoted to explaining how to create a single directory. I think most folks know how to create a directory.

dlabrecq:
Also need patternfly-additions.min.css.

dlabrecq:
Looking good. I still feel that we should also include patternfly.min.js as this overrides a lot of functionality in jQuery, Boostrap, C3, etc.

dlabrecq:
Maybe I'm missing something, but why are we instructing folks to compile the patternfly less files? The files below are already compiled and available in the dist directory. I believe patternfly.less also imports bootstrap, font awesome, etc. Shouldn't we just add the files below using a link tag?

patternfly/dist/css/patternfly-additions.min.css
patternfly/dist/css/patternfly.min.css
patternfly/dist/js/patternfly.min.js

LHinson:
@mindreeper2420 thanks for attaching those additional screenshots! For the last sentence, consider modifying to something like "Now that you have everything configured, enjoy exploring and using PatternFly. For any questions, reach out to us on our mailing list, gitter or on Freenode at #patternfly."  <<< All those things should link to the appropriate places.

mindreeper2420:
@LHinson I've updated the screenshots to match the current commit.

screen shot 2016-07-26 at 9 23 40 am
screen shot 2016-07-29 at 11 37 47 am
screen shot 2016-07-29 at 11 37 58 am
screen shot 2016-07-29 at 11 38 07 am
screen shot 2016-07-29 at 11 38 14 am
screen shot 2016-07-29 at 11 38 28 am
screen shot 2016-07-29 at 11 38 36 am
screen shot 2016-07-29 at 11 38 44 am
screen shot 2016-07-29 at 11 38 51 am

LHinson:
Thanks for doing that. Looks great @mindreeper2420

dlabrecq:
Built and ran your branch and the instructions look much cleaner/simpler.

Now that the CSS files are included directly in the HTML, can we remove steps 4 and 5 to create the cs folder and styles.css file? It appears to be an empty file now.

srambach:
Sorry, I think I might be too late, but if you only include the .css files, the project doesn't have access to the Less variables and mixins to use and extend. Don't we want users of Patternfly to be able to build their custom styles on top of it?

dlabrecq:
Is this level of customization something we want to show in getting started? Most folks don't need to setup their apps like this.

Bootstrap shows a very basic HTML template/example in their getting started doc. They have a section on compiling CSS and JavaScript, but it's not part of their basic example.

srambach:
Because of the split patternfly and patternfly-additions, I think it's worth explaining the best way to get everything included properly.

PR #237, opened 2016-08-03
fix drunk html special character on setup page

mwcz:
Hi Andres, Hi Brian!


srambach:
LGTM. We don't want those special characters drinking. :-)

dgutride:
Looks good to me.

PR #241, opened 2016-08-03
changed index.html version reference

priley86:
* also noted that `source/assets/css/patternfly-site.css.map` was no longer needed in source control


dlabrecq:
LGTM

PR #248, opened 2016-08-04
This file should be removed and ignored by .gitignore.

dlabrecq:
The patternfly-site.css.map file was added by Gabriel's commit and is now causing a merge conflict with Leslie's PR. This file should be removed and built by Travis.


dgutride:
Looks good to me - does the .gitignore change also need to be included so this file won't be added in the future?

dlabrecq:
There is already an "source/assets/css" entry in .gitignore, but this file must also be removed so it Git can ignore it in the future.

dgutride:
Ok - lgtm

PR #255, opened 2016-08-11
remove merge comments from markdown page, which render on patternfly.org

mindreeper2420:
The previous work on adding the migration guides left remnants of the merges. This PR fixes those.


dlabrecq:
LGTM

PR #257, opened 2016-08-11
removed detail that only one toast should appear at a time

LHinson:
updated toast page so that "one toast at time" was removed. It will remain unspecified at this time.


jeff-phillips-18:
LGTM

dlabrecq:
LGTM

Issue #259, opened 2016-08-12
Angular code for Inline Notifications in listed under Toast Notifications


rhwood:
The Angular code that matches the design for Inline Notifications is instead shown for [Toast Notifications](https://www.patternfly.org/pattern-library/communication/toast-notifications/#/angular).

rhwood:
Fixed in #261.

PR #260, opened 2016-08-12
Show Angular code for Inline Notifications on correct page

rhwood:
Addresses #259 by showing the Angular code for Inline Notifications on the [Inline Notifications](https://www.patternfly.org/pattern-library/communication/inline-notifications/#/_code) page and hide the same code on the [Toast Notifications](https://www.patternfly.org/pattern-library/communication/toast-notifications/#/_code) page.

---
This change is [Reviewable](https://reviewable.io/reviews/patternfly/patternfly-org/260)

PR #261, opened 2016-08-13
Move Angular code for Inline Notifications to correct page.

rhwood:
Addresses #259 by showing the Angular code for Inline Notifications on the [Inline Notifications](https://www.patternfly.org/pattern-library/communication/inline-notifications/#/_code) page and hide the same code on the [Toast Notifications](https://www.patternfly.org/pattern-library/communication/toast-notifications/#/_code) page.

---
This change is [Reviewable](https://reviewable.io/reviews/patternfly/patternfly-org/261)


rhwood:
This is the same content as #260, but done in a single commit.

dlabrecq:
LGTM. Thanks for squashing.

PR #279, opened 2016-08-31
Fix links that pointed to incorrect files/locations.

mindreeper2420:
This fixes the 404 errors that appeared due to incorrect file paths.


dlabrecq:
LGTM

PR #284, opened 2016-09-10
Consume design markdown from patternfly-design

bleathem:
This PR introduces the `patternfly`, `patternfly-design`, and `angular-patternfly` repositories as submodules of this repository.  The submodules are setup to track master, master, and master-dist respectively, and are pulled every time a grunt build is executed.

When a new design is meant to be included in the site, a markdown file must be created in the site repo with front matter pointing to the overview, design, and code source.  See the `source/pattern-library/data-visualization/heat-map/index.md` in this PR as an example.

The patternfly-page layout includes javascript to set the page title to the title in the md file.  The title in the md file is then removed.


.gitmodules
```diff
@@ -0,0 +1,12 @@
+[submodule "submodules/patternfly-core"]
+	path = submodules/patternfly-core
+	url = https://github.com/patternfly/patternfly.git
+	branch = master
```
dlabrecq: We should pull from master-dist, similar to angular-patternfly, so pf.org can ultimately remove the patternfly dependency.

source/_layouts/page-pattern.html
```diff
@@ -0,0 +1,57 @@
...
+            {% include nav-tabs-pattern.html designtab=false %}
+          {% else %}
+            {% include nav-tabs-pattern.html %}
+          {% endif %}
+        </div>
+      </div>
+    </header>
+    <section class="post_content">
+      <div class="tab-content">
+        <div role="tabpanel" class="tab-pane active" id="overview">
```
dlabrecq: I know there is an overview.md file for heat map, but this may be a bad example to follow? On pf.org, heat map does not have any examples on the overview page; however, this is the exception -- there isn't an example for core patternfly, only an angular implementation.

Every other overview page on pf.org has examples, included from patternfly. This is in addition to the examples for the code tab.

Perhaps the designers decided to remove the examples from overview? Or, the heat map is a bad example to follow?

source/_layouts/page-pattern.html
```diff
@@ -0,0 +1,57 @@
...
+          {% capture my_include %}{% include {{page.overview}} %}{% endcapture %}
+          {{ my_include | site_replace | markdownify }}
+        </div>
+        <div role="tabpanel" class="tab-pane" id="design">
+          {% capture my_include %}{% include {{page.design}} %}{% endcapture %}
+          {{ my_include | site_replace | markdownify }}
+        </div>
+        <div role="tabpanel" class="tab-pane" id="code">
```
dlabrecq: This template only includes an angular example, not an html/css example from patternfly core. By setting html=false, it disables that tab.

Each set of pages throughout pf.org are unique. For example, some code pages have html/css examples only, while others only have an angular example and a few have both html/css and angular examples.

source/_layouts/page-pattern.html
```diff
@@ -0,0 +1,57 @@
...
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    </section>
+  </article>
+</div>
+<script>
+  var pageHeader = document.querySelector('.page-tabs-header h1'),
```
dlabrecq: I don't understand why the h1 tags are being hidden?

source/pattern-library/data-visualization/heat-map/index.md
```diff
@@ -1,60 +1,7 @@
 ---
-title: Heat Map
-author: dlabrecq
-layout: page-tabs
+author: lhinson
+layout: page-pattern
+overview: patterns/data-visualization/heat-map/overview.md
+design: patterns/data-visualization/heat-map/design.md
+code: /components/angular-patternfly/dist/docs/partials/api/patternfly.charts.directive.pfHeatMap.html
```
dlabrecq: There may be both angular and html/css (core) code here.

dlabrecq:
There are many images added under the source/pattern-library/card/img and source/pattern-library/data-visualization/heat-map/img directories, which doesn't seem correct? We already have a source/pattern-library/cards (plural) directory.

Shouldn't these images be located under the submodules?

source/_layouts/page-pattern.html
bleathem: @dlabrecq the only pattern that has been converted to the patternfly-design repo so far is the heatmap.  This PR addresses including the heatmap mardown in the site.  As more patterns are converted our approach can be adjusted pragmatically.

source/_layouts/page-pattern.html
bleathem: We can adjust the template pragmatically when a new pattern is available in the patternfly-design repo for us to consume.

source/_layouts/page-pattern.html
bleathem: I explained this in the PR desccription:
> The patternfly-page layout includes javascript to set the page title to the title in the md file. The title [from] the md file is then removed

Where by *title* I mean the *h1* tag. 

Does that clarify the situation sufficiently for you?

source/pattern-library/data-visualization/heat-map/index.md
bleathem: I'll change the frontmatter to `code_angular` to allow for the addition of `code_core` later.

bleathem:
When the images are located in the `_includes` folder, they are not served along with the rendered html.  I had to copy the images under the `pattern-library` folder in order for them to be served.  I would be happy to hear an alternate solution for this.

The `card` pattern was an initial PoC pattern introduced in the patternfly-design repo.  It either needs to be removed from the design-repo or completed (and renamed to cards) (cc: @LHinson).  For now I'm considering it a WIP.  No site makrdown file was created for it, so none of the mardown will be converted to html.  We can simply ignore the images until the pattern is either removed or completed.

source/_layouts/page-pattern.html
dlabrecq: I'm more curious if this was an intentional decision that affects the entire site (i.e., to no longer include examples in the overview)? Or, perhaps that was an over-site when including the heat map overview page in the design repo?

If that's the plan for future overview pages, I'm fine with that. Essentially, our website pages now live in only in the deign repo.

It would be good to know that so we stop adding examples to overview pages. I recently merged a couple last sprint.

source/_layouts/page-pattern.html
dlabrecq: My thought is that we continue to set the title as a variable in the template. Then, use a style to hide the first h1 each tab-pane selector. Otherwise, the JavaScript removes all h1 tags in the markup, not just the first title in each pane -- it's possible, although unlikely they would use multiple h1 tags.

For example, we could use this instead:
 
.tab-pane:not(.nested) > h1:first-child {
  display: none;
}

source/_layouts/page-pattern.html
dlabrecq: It is more the case that the code tab displays multiple examples. Typically, there is only one angular example, but many html/css (core) examples are shown in that tab.

dlabrecq:
Can we not copy them directly to the _site folder, instead? The will be picked up there and we don't need to add the images to another repo.

Don't feel strongly about the "card" pattern name, especially if we copy the images. However, there could be name collisions if we're not careful -- debugging could become tricky.

source/pattern-library/data-visualization/heat-map/index.md
```diff
@@ -1,60 +1,8 @@
 ---
-title: Heat Map
-author: dlabrecq
-layout: page-tabs
+author: lhinson
+layout: page-pattern
+overview: patterns/data-visualization/heat-map/overview.md
+design: patterns/data-visualization/heat-map/design.md
+code_html: false
+code_angular: /components/angular-patternfly/dist/docs/partials/api/patternfly.charts.directive.pfHeatMap.html
```
dlabrecq: Should this use angular-partials instead of /components/angular-patternfly/dist/docs/partials? It looks like the GruntFile copies the partials there?

source/_layouts/page-pattern.html
bleathem: We may have to revisit how we package the examples, for instance a single html file with all examples in it.  Let's solve that problem when we add a pattern that has such a set of examples.

source/_layouts/page-pattern.html
bleathem: I like having a single source of truth for the design title.  If the title gets adjusted in the markdown, it's nice that the page gets updated automatically, without having to remember a manual task to change the title.

The current design templates in the patternfly-design repo suggest only a single h1 header (https://raw.githubusercontent.com/patternfly/patternfly-design/master/pattern-template/design/design.md and https://raw.githubusercontent.com/patternfly/patternfly-design/master/pattern-template/design/overview.md).  If this ever becomes a problem, we can adjust our implementation accordingly.

source/pattern-library/data-visualization/heat-map/index.md
bleathem: 1) the _includes folder isn't sourced to the web server directly
2) I tried using a jekyll include instead of the current angular include so we could use the submodule code, but it didn't work.  I plan to track getting that to work with a new story.

bleathem:
> Can we not copy them directly to the _site folder, instead?

This is an excellent suggestion.  I've added a commit making this change.  The result is a much cleaner.  Thanks!

source/_layouts/page-pattern.html
dlabrecq: IDK. I'm afraid this template is only going to work for heat map, which isn't a typical use case. If we introduce a new template, these use cases should be considered.

A good example to look at would be any of the cards (e.g., utilization trend card). There are multiple Patternfly examples here, included via templates. These particular  examples render multiple charts, which require unique IDs.

Some examples (e.g., aggregate-status-card) also include JavaScript to initialize the examples properly. Other pages include extra javaScript in the pretty print to show the required includes.

source/_layouts/page-pattern.html
dlabrecq: Unfortunately, the only way to set the title is via these documents. There are cases where there the design tab is disabled (e.g., timepicker) because not all documents are available.

source/pattern-library/data-visualization/heat-map/index.md
dlabrecq: I see. I understand that Dave wrote the ng-include directive to include the angular partials. You might want to have a chat with him?

source/_layouts/page-pattern.html
dlabrecq: Looking at the new page.code_html include and thinking we should be able to create any custom example pages we want. It's just that those pages now must live under the _includes directory, which is getting crowded. Not sure there is much we can do about that? I'd prefer additional files be located with the heat map index.md file, for example.

LHinson:
@bleathem @dlabrecq regarding the examples on the overview page. The site is iterative... first the visual design happens and that is posted. Once the code is available, it replaces the examples on the overview. So, the heat map doc is a valid example as there are instances where just images are shown. I guess the question is... is there a way to handle including the code on the overview once that is available?

bleathem:
We've created a new story to track examples in the overview page: https://patternfly.atlassian.net/browse/PTNFLY-1835

IMO all concerns in this PR that effect the currently developed patternfly-design impls have been addressed.  The remaining concerns can be addressed when we come to merge patternfly-design impls make those concerns immediate.

Are you ok if I merge this PR @LHinson and @dlabrecq ?

LHinson:
Thanks for reviewing the specifics of this in the PF des/dev meeting @bleathem This approach LGTM.

PR #287, opened 2016-09-12
fix type on bower install command on patternfly setup page

sjd78:
On the "Setting Up PatternFly" page, the bower install command was missing the npm bit.


dlabrecq:
Good catch!

Issue #293, opened 2016-09-13
Find dropdown is not opening in Toolbar component


cardosogabriel:
In toolbar page, when clicking on the search icon, the dropdown does not open.
http://www.patternfly.org/pattern-library/forms-and-controls/toolbar/#_code

dlabrecq:
I see the drop down open on Firefox and Chrome. However, there is no JavaScript added to make this functional. That is, selecting an option won't change the drop down state.

cardosogabriel:
@dlabrecq I found the issue. Somehow this javascript is not working inside patternfly-org: https://github.com/patternfly/patternfly/blob/master/tests/pages/_includes/widgets/layouts/toolbar.html#L104
When you click on the .btn-find button, instead of adding display:block in .find-pf-dropdown-container, it adds display:none. In patternfly it works properly.
Any clue? @jeff-phillips-18 @bleathem @andresgalante

dlabrecq:
Looking at the JavaScript below.

$(".btn-find").click(function () {
  $(".find-pf-dropdown-container").toggle();
});

The code intended to find an element with a find-pf-dropdown-container class is returning undefined. It may work in the Patternfly test page because there is only one element with this class. In pf.org, there are multiple elements with this class. My guess is that the code needs to be written differently (e.g., use a unique ID instead of a style class).

cardosogabriel:
Yes, I was suspecting it was this also. I tried this code:

$(".btn-find").click(function () {
  $(this).parent().find(".find-pf-dropdown-container").toggle();
});

It worked on the patternfly page, but not on patternfly-org. What would be a code that would work?

dlabrecq:
If you obtain the parent div tag, then this code should work.

$(".btn-find").click(function () {
$(this).parent('div').find(".find-pf-dropdown-container").toggle();
});

cardosogabriel:
Thanks Dan.
The code I wrote works. Just sent a PR fixing the issue. Actually there were two problems: the code itself, and the fact that the javascript was being loaded 2 times.

LHinson:
I just check this and it is working. Looks like PR #301 did the trick so I'm closing.

PR #301, opened 2016-10-05
Find dropdown is not opening in Toolbar component #293

cardosogabriel:
Fix for issue #293.
Problem is that the javascript was inside the toolbar.html, that was being imported twice. Thus, the duplicated javascript was not working.
I removed the javascript from the toolbar.html and added it to the page code.
This depends on an update in patternfly project.


dlabrecq:
@cardosogabriel Can you please fix the merge conflicts?

cardosogabriel:
@dlabrecq , it's rebased.

PR #302, opened 2016-10-05
Added a script to publish to github pages

bleathem:
This PR introduces a script to publish the contents of the `source/_site` folder to the patternfly.github.io repository.  This makes the site available via the URL:
http://patternfly.github.io/

@dlabrecq please review this PR.  We can merge it with minimal impact, as the site will be dual hosted on openshift (patternfly.org) and github pages (patternfly.github.io).  Once we are satisfied with the github pages we can transfer the CNAME to point to github pages, and remove the script that publishes to openshift (as a separate story/PR).


scripts/publish-ghpages.sh
```diff
@@ -0,0 +1,27 @@
+#/bin/sh
+
+set -o errexit -o nounset
+
+# User info
+git config user.name "patternfly-build"
+git config user.email "patternfly-build@redhat.com"
+git config --global push.default simple
+
```
dlabrecq: Will need to check for patternfly/patternfly-org and the master branch. The travis.yml file does this via the deploy properties. Otherwise, any clone can push to GitHub and potentially corrupt the site.

Take a look at the publish.sh script for patternfly. It tests $TRAVIS_REPO_SLUG, $TRAVIS_TAG and $TRAVIS_BRANCH.

scripts/publish-ghpages.sh
```diff
@@ -0,0 +1,27 @@
...
+## Deploy to github pages
+if [ -d "patternfly.github.io" ]; then
+  rm -rf patternfly.github.io
+fi
+git clone git@github.com:patternfly/patternfly.github.io.git
+rsync -av --delete --exclude .git source/_site/ patternfly.github.io
+cd patternfly.github.io
```
dlabrecq: Curious about the need for a separate patternfly.github.io repo. Is that a GitHub Pages requirement? Can we not use another branch (e.g., master-dist) in patternfly/patternfly.org?

My concern is that if patternfly, angular-patternfly, and RCUE move to GitHub, will we need to create 4 additional repos?

scripts/publish-ghpages.sh
```diff
@@ -0,0 +1,27 @@
...
+find components -type f -not -regex ".*/.*\.\(html\|js\|css\|less\)" -print0 | xargs -0 rm
```
dlabrecq: What file types are being omitting from the patternfly.github.io repo?

scripts/publish-ghpages.sh
```diff
@@ -0,0 +1,27 @@
...
+git add . -A
+if [ -z "$(git status --porcelain)" ]; then
+  echo "Working directory clean, no changes to commit."
+else
+  # Commit generated files
+  git commit -a -m "Added files generated by Travis build"
+  git push origin master:master
+fi
+cd ..
+rm -rf patternfly.github.io
```
dlabrecq: Don't need to rm as the Travis workspace will be deleted.

scripts/publish-ghpages.sh
```diff
@@ -0,0 +1,27 @@
...
+set -o errexit -o nounset
+
+# User info
+git config user.name "patternfly-build"
+git config user.email "patternfly-build@redhat.com"
+git config --global push.default simple
+
+## Deploy to github pages
+if [ -d "patternfly.github.io" ]; then
+  rm -rf patternfly.github.io
```
dlabrecq: Don't need to rm as the Travis workspace is already clean.

bleathem: I put this in so I could run the script locally, and we can circumvent travis when we need to manually publish the site.

scripts/publish-ghpages.sh
bleathem: Yeah, I've given this some though, and I feel we could accomplish this with a single repo (consolidating on the patternfly.github.io repo).  I'll re-factor my approach and update this PR.

scripts/publish-ghpages.sh
bleathem: All non web resources.  The bower components brings in things like exe binaries etc.  This is a hack.  When we move to consuming the git submodules and npm dependencies instead of a blind bower components copy this can be removed.

scripts/publish-ghpages.sh
bleathem: Same response as above, I want to be able to run this script locally when we need to manually push the site. (eg. travis is unavailable)

dlabrecq:
I see that this was merged, but don't see my concerns being addressed. Was anything changed?

dlabrecq:
FYI, we typically have two developers review changes.

PR #304, opened 2016-10-07
Added a script to publish to github pages

bleathem:
This is a follow-on PR to https://github.com/patternfly/patternfly-org/pull/302 that was accidentally merged and subsequently reverted (cc @dlabrecq).

I believe I've addressed all the concerns raised in that PR, barring the dual repo question.  It turns out we will need to use 2 repos, as github only allows organisational pages to be published from the master repo (see: https://help.github.com/articles/user-organization-and-project-pages/).  We can simply think of the patternfly.gihub.io repo as the hosting location for the site.


dlabrecq:
LGTM

PR #305, opened 2016-10-12
Updated the build to use an up-to-date ruby gem bundle

bleathem:
This PR:
*  Adds a grunt jekyll option to use the 'bundle exec' prefix when running jekyll
* Updates the gems in our Gemfile.lock file
* Updates the README to include bower and bundle install steps


dlabrecq:
The Travis build is failing due to this error. This is similar to what I was seeing with saas.

Gem::InstallError: rack requires Ruby version >= 2.2.2.

bleathem:
Thanks @dlabrecq.  I've updated the travis.yml file to use a recent ruby version, and the build succeeds.

andresgalante:
if @dlabrecq is ok with it, then I am more than ok with it.

dlabrecq:
LGTM

PR #310, opened 2016-10-19
Updating nav and pattern names and adding errors and validation pattern

LHinson:


bleathem:
These changes look good, will merge.

PR #328, opened 2016-11-25
Consume site.md files from the patternfly-design repo

bleathem:
This introduces changes to allow consumption of `site.md` files from the patternfly-design repo.  These file are used to generate the index.html pages of pattern-library pages, and enables designers to initiate posting of designs to the website from the patternfly-design repo. ([PTNFLY-1884](https://patternfly.atlassian.net/browse/PTNFLY-1884)).

To enable this, an intermediate step was introduced into the site build.  A new folder `_build` was introduced to be used as the target for jekyll.  Jekyll generates the site from this `_build` folder to the `_site` folder (now located in the project root).

The `source` folder is now a "human edited" folder, with all generated source now targeting the `_build` folder.  The `site.md` files are copied from the patternfly-design folder into the `_build` folder as `index.md` files, to be consumed by jekyll.

Note: to test this PR, one should first run:
```
git clean -f -x -d
npm install
bower install
```


dlabrecq:
I built and ran the changes locally and the build looks good. The only problem is that Travis is failing.

Although, I don't believe this is related to the PR, it appears that publish-ghpages.sh is executing for pull requests. We might add something like this to to publish-ghpages.sh to prevent the error?

# Skip publish for pull requests
if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
 exit 0
fi

bleathem:
Thanks for the tip Dan.  I've incorporated the check into the publish-ghpages script.

dlabrecq:
LGTM

Issue #338, opened 2017-01-09
Notification drawer - code tab is broken


michpetrov:
[Link](http://www.patternfly.org/pattern-library/communication/notification-drawer/#/_code)

The code being shown is for Angular and the HTML code is not included. Switching to Angular and back to HTML results in a blank page.

The [layout page](https://github.com/patternfly/patternfly-org/blob/master/source/_layouts/page-pattern.html) is not set up correctly, I'm working on a fix.

PR #339, opened 2017-01-09
Fix page-pattern layout

michpetrov:
Fix for #338. The HTML code page is now properly included. Currently Notification drawer is the only pattern that uses this layout and has a HTML code page.

The index page in patternfly-org is being overwritten by a page from patternfly-design so I assume it's safe to delete.


bleathem:
Thanks @michpetrov! (I had this in my local clone, I forgot to include it in #334.)

PR #341, opened 2017-01-14
Change slack link to point at new autoenroll page

dgutride:
The rhcloud site is live on openshift now - this is ready to go in once approved.

@bleathem @dlabrecq


bleathem:
We manage the patternfly DNS ourselves on CloudFlare now. I'll set up this sub domain, and make a story to write up a doc on how we manage DNS.

bleathem:
I setup the CNAME: https://slack.patternfly.org/.

@dgutride you will have to configure openshift.com to be aware of that CNAME.

I would also suggest we add an "Already a member?" notice to the signup page, with a link to patternfly.slack.com.

PR #345, opened 2017-01-19
Introduce a static menu generator

bleathem:
In this PR I've introduced a jekyll plugin that scans the `site.pages` to dynamically create the _pattern-library_ menu items for the site.

The plugin executes and the menus are generated on every build.  No special action is required to enable this functionality.  Simply run: `grunt serve`


LHinson:
bitmoji

PR #349, opened 2017-01-26
Fixed the font download links

bleathem:


cshinn:
Both of those links seem to work. LGTM

PR #350, opened 2017-01-27
Change the markdown parser from kramdown to redcarpet.

bleathem:
This PR configures jekyll to use the redcarpet markdown parser, rather than kramdown.

This change is introduced to fix the bug where the markdown rendered on patternfly.org [1] doesn't match what's shown on Github [2]:
[1] http://www.patternfly.org/pattern-library/application-framework/masthead/#/_design
[2] https://github.com/patternfly/patternfly-design/blob/master/pattern-library/application-framework/masthead/design/design.md

By switching to the redcarpet markdown parser, we achieve better consistency in rendering markdown between gitHub and patternfly.org, as redcarpet is what Github uses to render markdown (see https://github.com/blog/832-rolling-out-the-redcarpet).

Since this change affects the entire site, I've staged the site built with redcarpet for a thorough review:
https://bleathem.github.io/

@LHinson @beanh66 if you could please review the staged site to check for any rendering errors


beanh66:
@bleathem I noticed a few pages with issues. The Toolbar pattern isn't displaying correctly and the widgets section of the site seems to have problems as well.

bleathem:
Thanks for the review @beanh66.

I tried all day to get the redcarpet processor to work.  In the end I discovered that github had switched to using kramdown itself, so I'm dropping this PR.

I was able to find a simpler fix for the masthead, as can be seen in the PR:
https://github.com/patternfly/patternfly-design/pull/178

PR #351, opened 2017-01-28
Allow for setting the jekyll baseurl

bleathem:
This PR introduces the necessary changes to allow the site to be served from different baseurls.

To try it out, set the `baseurl` property of the `source/_config.yml` file to something like `/username` and run `grut serve`.

Your review of this would be much appreciated @dlabrecq

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
   - TRIGGER_REPO_BRANCH: "master"
 
 before_install:
+  - nvm install 4
   - gem install jekyll -v 3.1.6
   - if [[ `npm -v` != 3* ]]; then npm i -g npm@3; fi
   - npm install -g bower grunt-cli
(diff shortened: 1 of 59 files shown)


bleathem:
Travis build is failing, I'll need to configure travis to use node 4.

.travis.yml
```diff
@@ -10,6 +10,7 @@ env:
   - TRIGGER_REPO_BRANCH: "master"
 
 before_install:
+  - nvm install 4
   - gem install jekyll -v 3.1.6
   - if [[ `npm -v` != 3* ]]; then npm i -g npm@3; fi
   - npm install -g bower grunt-cli
```
dlabrecq: Can you explain what's happening regarding jekyll and npm? Not sure what versions are being installed under nvm 4?

bleathem: jekyll is a gem and the version used is controlled by the Gemfile.lock file present int he repo.  nvm is the node version manager, and is used to install node 4 (which we use to build patternfly and angular patternfly).  We subsequently update npm to be version 3.  Note: nvm != npm.

PR #357, opened 2017-02-08
Table View - Selected Row color not being applied

dabeng:
This PR is used to fix the [bug](https://github.com/patternfly/patternfly/issues/577).

Additionally, there are still two issues we need to consider:

1.  Maybe we should add datatables.net-select as a dependency in bower.js file of patternfly repo. Then we can introduce datatables.net-select with the form of ``` <script src="{{ "/components/...```

2. I found that the following code snippets of patternfly-site.js have a conflict with the page resize feature of table-view. @rhamilto, could you re-verify the practical use of the following sentence. As far as now, I can not  find any changes or issues if I delete this statement.
````js
  // Initialize Bootstrap-select
  jQuery('.selectpicker').selectpicker();
````

### DIFF
--- a/source/_includes/head.html
+++ b/source/_includes/head.html
@@ -38,6 +38,12 @@
   <script src="{{ "/components/c3/c3.min.js" | prepend: site.baseurl }}"></script>
   <script src="{{ "/components/d3/d3.min.js" | prepend: site.baseurl }}"></script>
   <script src="{{ "/components/datatables/media/js/jquery.dataTables.js" | prepend: site.baseurl }}"></script>
+  <script src="//cdn.datatables.net/select/1.2.0/js/dataTables.select.min.js"></script>
+  <script src="{{ "/components/patternfly/dist/js/patternfly.dataTables.pfEmpty.min.js" | prepend: site.baseurl }}"></script>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by bleathem** [comment]
This PR introduces the js dependencies in the head.html include file.  This means these dpendencies will be loaded on every patternfly.org page request.

Rather, you will want to enumerate these dependencies (as I suggested in [PTNFLY-2141](https://patternfly.atlassian.net/browse/PTNFLY-2141) in the file:
https://github.com/patternfly/patternfly/blob/master-dist/dist/tests/table-view.html

eg. right now that file looks like:
```
...
    <script src="../../dist/js/patternfly.dataTables.pfEmpty.js"></script>
    <script src="../../dist/js/patternfly.dataTables.pfFilter.js"></script>
    <script src="../../dist/js/patternfly.dataTables.pfPagination.js"></script>
    <script src="../../dist/js/patternfly.dataTables.pfResize.js"></script>
    <script src="../../dist/js/patternfly.dataTables.pfSelect.js"></script>
    <script src="../../dist/js/patternfly.min.js"></script>
...
```

You will want to change the relative paths `../../dist/js/patternfly.dataTables.pfEmpty.js` to absolute paths like `https://cdn.rawgit.com/patternfly/patternfly/master-dist/dist/js/patternfly.dataTables.pfEmpty.js`.

With the change localised to that file, those javascript dependencies will only be loaded on the single example page that needs them, rather than over page of the site.

PR #358, opened 2017-02-09
Support for staging deployments

bleathem:
This PR introduces a number of changes to support staging the site to patternfly-org forks:

- The build has been updated to accept either a `staging` or `production` target, where `staging` is the default target.  The target can be invoked explicitly as: `grunt build:production` or `grunt build:staging`
- Grunt configs have been changed to behave differently for `staging` and `production` builds:
  - The CNAME file is only included for `production` builds
  - The _config-staging.yml file is included only for `staging` builds
- "Non-web" files have been excluded in the bower_components copy at build time (simplifying the deployment script)
- The dependency on _patternfly-eng-publish_ has been updated to the latest node-based release.

@dlabrecq your expertise with the patternfly-org site would be greatly appreciated in reviewing these changes.


.travis.yml
```diff
@@ -21,5 +21,5 @@ script:
   - sh -x ./node_modules/patternfly-eng-release/scripts/_build.sh -o
 
 after_success:
-  - grunt build
+  - grunt build:production
```
dlabrecq: The build script from patternfly-eng-release also performs a build. Basically, we're building the repo twice. Do we want the patternfly-eng-release script to use grunt build:production instead?

bleathem: +1 for not building twice.  We would the `:production` target for anything travis buiids.

Can we pull the build step out of the release plugin so it's better exposed?  IMO it should be an explicit step in the travis config.  Scripts executed by travis should be single purpose for optimal clarity.

dlabrecq: The release still needs to build because we do a lot after the build completes. We modify files first, then build, shrinkwrap, verify installs, publish npm, push to master, master-dist, etc. Adding the build step in Travis config would introduce a timing issue.

This would be a problem for a non-release build as well. For most repos, we also push generated files to master-dist.  We don't have a master-dist for pf.org; however, It's easier if the scripts do this because they decide if we're performing a release build or not and publish to master-dist.

dlabrecq: I had moved this logic to the scripts mainly to keep the Travis config clean and it was common to each repo.

dlabrecq: Added the :production target to the patternfly-eng-release scripts

bleathem: The travis scripts are executed sequentially, so it would look like:
- <build script>
- <release script>
- <publish script>

These single purpose scripts more clearly indicate what travis is accomplishing.

dlabrecq: That would be ideal. Unfortunately, there are many changes that must occur prior to a build. Similarly, there is a bit of logic to determine what we're building; normal build, release, and alpha.

PR #360, opened 2017-02-09
Copied the component dependencies from node_modules rather than bower_components

bleathem:
This PR introduces a change where the component dependencies are copied from node_modules rather than bower_components.  This makes it unnecessary to run `bower install` before building the site.

Please review @dlabrecq

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -98,14 +98,30 @@ module.exports = function (grunt) {
       }
     },
     copy: {
-      bower: {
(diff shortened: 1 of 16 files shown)


bleathem:
I've staged the site built with this branch here:
https://bleathem.github.io/patternfly-org/

A non-exhaustive review doesn't indicate any breakages.

bleathem:
Thanks for the review @dlabrecq

PR #363, opened 2017-02-11
Split the build step in two, to allow external scripts to modify the _build folder prior to the jekyll build

bleathem:


bleathem:
Simple fix, merging.

PR #373, opened 2017-03-01
Show/hide functionality does not work on Vertical Nav

dabeng:
Fix the bug: [Show/hide functionality does not work on Vertical Nav example code when accessed from web site.](https://github.com/patternfly/patternfly/issues/607)

### DIFF
--- a/source/_includes/code/cards/aggregate-status-card/code.md
+++ b/source/_includes/code/cards/aggregate-status-card/code.md
@@ -39,7 +39,7 @@
           </div>
         </div><!-- /row -->
       </div><!-- /container -->
-      <script src="/components/jquery-match-height/jquery.matchHeight-min.js"></script>
+      <script src="/components/jquery-match-height/dist/jquery.matchHeight-min.js"></script>
       <script>
       $(function() {
         // matchHeight the contents of each .card-pf and then the .card-pf itself
@@ -95,7 +95,7 @@
           </div>
         </div><!-- /row -->
       </div><!-- /container -->
-      <script src="/components/jquery-match-height/jquery.matchHeight-min.js"></script>
+      <script src="/components/jquery-match-height/dist/jquery.matchHeight-min.js"></script>
       <script>
       $(function() {
         // matchHeight the contents of each .card-pf and then the .card-pf itself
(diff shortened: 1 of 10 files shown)


michpetrov:
@dabeng  I was about to submit a PR and then I noticed yours. Anyway, there are 12 links for jquery.matchHeight-min.js, you should change all of them.

dabeng:
Hi @michpetrov, thank you for your reminding 😊 . While I think minimal modification for a particular issue is a good practice and then we can get the limited testing scope 😊

michpetrov:
I don't think that's a good practice, especially in this case where the only problem is the wrong/outdated path, the widget itself has been already tested.

bleathem:
Why are we making this change?  It seems the `jquery-match-height` project ships the minified js file in both the root and dist folders.

```
$ tree _build/components/jquery-match-height/
_build/components/jquery-match-height/
├── bower.json
├── CHANGELOG.md
├── dist
│   ├── jquery.matchHeight.js
│   └── jquery.matchHeight-min.js
├── gulpfile.js
├── jquery.matchHeight.js
├── jquery.matchHeight-min.js
...
```

dabeng:
Hi @bleathem , I can't see the minified version of jquery.matchHeight.
screen shot 2017-03-09 at 2 04 44 pm

michpetrov:
@bleathem @dabeng it was removed a while ago - liabru/jquery-match-height@f530833 as far as I can tell npm doesn't clean up extraneous files.

bleathem:
Thanks for pointing that our @michpetrov.  I removed the node_modules folder, and ran `npm install`, and indeed the minified file is not present.

@dabeng can you please expand the scope of this PR to replace all incorrect file references?  We might as well fix them in one swoop.

dabeng:
OK. no question.

LHinson:
@dabeng has Brian's comment about replacing all incorrect file references been addressed?

dabeng:
@LHinson , current PR has fixed all incorrect references. Don't worry. @michpetrov , would you review again?

LHinson:
OK thanks @dabeng @michpetrov! I'll merge it. 

Do we have issues filed for the bad links? Sorry I'm so lost on this one @michpetrov If you point them out to me, I'm happy to file the issues too. Thanks!

michpetrov:
@LHinson you have now - patternfly/patternfly-design#237

LHinson:
Appreciated @michpetrov!

PR #381, opened 2017-03-06
Added a search icon to the icon-search input

bleathem:
because I can't leave well enough alone.  cc: @LHinson


LHinson:
It's tough being a perfectionist :)

PR #383, opened 2017-03-07
fix jQuery selector

michpetrov:
Fix for URL handling in Pattern Library.

jQuery requires special characters in quotes, i.e. `a[href="#api"]`, instead of just `a[href=#api]` (which doesn't work and unlike other situations throws an error).


bleathem:
Thanks @michpetrov !

PR #384, opened 2017-03-08
Fixes the window.location tab handling

bleathem:
This is staged at:
https://bleathem.github.io/patternfly-org

Test it by navigating around pattern-library pages, selecting overview/design/code tabs, and refreshing the page.  Notice that the active tab is correctly activated. 

cc: @LHinson


bleathem:
@dabeng @michpetrov if either of you would like to review this PR, that would be much appreciated

LHinson:
LGTM! +1

PR #389, opened 2017-03-23
fixed typo in readme.me

andrralv:
original readme has "grunt serve" to start up the server and it should be "grunt server"


LHinson:
Thank you @andrralv!

Issue #361, opened 2017-02-10
Datepicker is covered by navigation bar of patternfly website


dabeng:
![screenshot from 2017-02-10 14-35-05](https://cloud.githubusercontent.com/assets/5689075/22816663/eda8b914-ef28-11e6-9d6a-8fd1c1ef4051.png)

As a result, it's unavailable currently.

LHinson:
Closing this since #430 has been merged

PR #430, opened 2017-07-11
Date picker year behind navbar z-index fix

### LINKED ISSUE #388: datepicker bug (can not pick time)
The bug is the datepicker can not pick year and month because the part of it hidden in the back of navigation.http://www.patternfly.org/pattern-library/forms-and-controls/datepicker/#/code

amarie401:
Added a z-index to navbar to fix date picker year issue which would close #388 and #361 

screen shot 2017-07-11 at 4 19 14 pm

I also played around in dev tools with moving the top to 70px for the date picker since it was absolute positioned but that would put it too close to the "Reference Markup" text and it would need to be moved over more to the right in that case. 

Can you please review @bleathem and @andresgalante :)

### DIFF
--- a/source/_less/patternfly-site.less
+++ b/source/_less/patternfly-site.less
@@ -384,6 +384,7 @@ header {
     font-size: @font-size-base + 1;
     margin-bottom: 0;
     filter: progid:DXImageTransform.Microsoft.gradient(enabled = false) !important;
+    z-index: 1;
     .home & {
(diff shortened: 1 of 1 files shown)


andresgalante:
I like this, it fixes the bug on the website and leaves the component alone. 

Great job @amarie401 ! 👏🏼👏🏼👏🏼👏🏼

Issue #451, opened 2017-08-23
App launcher not displaying correctly


LHinson:
The html/css version of the app launcher is not working as expected here: http://www.patternfly.org/pattern-library/application-framework/launcher/#/code/html-css

Issues: 
* The logo does not render properly 
* The launcher icon does not even appear 

The implementation is working correctly in the test page seen here:  https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/application-launcher.html

PR #453, opened 2017-08-25
Includes css to overwrite ng-docs.css for app launcher

jgiardino:
CSS in ng-docs.css was causing the issue noted in #451 related to the logo. I found that the horizontal nav example had additional css to fix this. I added the same class `example-navbar`, and then included additional css for this class to fix the display within the vertical nav as well.


LHinson:
bitmoji

PR #485, opened 2017-10-24
Updated the Dockerfile to use centos7 as the base image

bleathem:


bleathem:
@cliffpyles I've made further adjustments to fix the livereloading and  further improve the container build time.

cliffpyles:
Looks great! 👍

PR #486, opened 2017-10-26
Fix PatternFly JS URL in setup page

tadayosi:
I found a tiny typo in Setting Up PatternFly page. It's a trivial but beginners may stumble at this by their copy&paste first trial.

Thank you!


tadayosi:
@bleathem Could you take a look at it please?

PR #489, opened 2017-10-28
Implented the Pattern Status page

bleathem:
The page is staged at:
http://bleathem.github.io/patternfly-org/stage/status-page/pattern-library/

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -183,6 +183,14 @@ module.exports = function (grunt) {
         }
       }
     },
+    http: {
+      pattern_status: {
+        options: {
+          url: 'http://www.patternfly.org/patternfly-design/status/pattern-status.json'
+        },
+        dest: '_build/_data/pattern-status.json'
+      }
+    },
     jekyll: {
       options: {
         dest: '<%= config.site %>',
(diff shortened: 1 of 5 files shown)


bleathem:
@mcarrano are you sure the web components work is being discontinued?

mcarrano:
I believe that was what was decided, @bleathem . @LHinson can you confirm this regarding Web Components?

LHinson:
Web Components is not being "discontinued" per say, but it is being put on hold and we will revisit at a much later date to see if it's something that we want to pursue in the future. It will not be part of our roadmap to date. 

The only js implementations that should be visible from the site at this time are: Angular 1, patternfly-ng and patternfly-react. That means we should remove JQuery, Web Componets and CSS Next. 

Also, would it be possible to link the pattern names to the corresponding web page? 

@bleathem @mcarrano

mcarrano:
Thanks for the clarification @LHinson .  My only question is whether we should keep a link for the jQuery implementation. This is essentially the PF Core implementation of HTML/CSS/jQuery found on the test pages, not the jQuery showcase. I think this is useful to reflect components with/without test pages currently.

bleathem:
I've made the pattern names links, dropped and renamed the columns as requested.  I've also updated the stage.

PR #501, opened 2018-01-11
fix(secondary-tabs): refactor the secondary tabs into nested structure

dabeng:
This PR is targeted to fix [issue](https://github.com/patternfly/patternfly/issues/905).

Please review the [Demo](https://rawgit.com/dabeng/patternfly-org/sec-tabs-dist/_site/pattern-library/widgets/#tabs).

### DIFF
--- a/source/pattern-library/widgets/index.md
+++ b/source/pattern-library/widgets/index.md
@@ -2116,19 +2116,59 @@ Bootstrap JavaScript modular. PatternFly also uses <a href="http://c3js.org/" ta
     <p><strong>Note:</strong> the markup semantics used here are sub-optimal in order to easily achieve a responsive layout.</p>
     <div class="pf-example">
       <ul class="nav nav-tabs">
-        <li class="active"><a href="#">Tab One</a></li>
-        <li><a href="#">Tab Two</a></li>
-        <li><a href="#">Tab Three</a></li>
-        <li><a href="#">Tab Four</a></li>
-        <li><a href="#">Tab Five</a></li>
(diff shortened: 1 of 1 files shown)


mcarrano:
@dabeng @jgiardino I tried to follow the discussion of this in the React repo, but have to admit to having a hard time understanding what we are changing and why. Can someone please give a quick summary? I would also like @andresgalante to review.

If we are changing the code in the PF core repo, should a PR be sent against the PatternFly repo or the website? Looks like this PR is no change the example on the website? @LHinson FYI

LHinson:
I think this is meant to be a change against core (patternfly-patternfly) vs against patternfly-org.

jgiardino:
@mcarrano We should only need a change to the css files in patternfly core for this and can close this PR. 

The way the css is currently written is tightly coupled with how the html is structured. But the react component cannot reproduce this html exactly the way the css selectors expect it to be. We don't need to change the current html. We only need to add an additional class to the existing styles for the secondary navigation so that we can apply those styles by adding this class to the react component.

I added comments to this PR to describe the changes that we need: https://github.com/patternfly/patternfly/pull/906

dabeng:
Hi @mcarrano, the quick summary is shown in the following picture. Could you tell me which is your favorite "secondary tabs"?
screen shot 2018-01-12 at 1 31 16 pm

mcarrano:
@dabeng Nesting the tabs like this was not what was intended. I could see where this may be useful, but to me it becomes more like a nav bar and I don't think we would want to couple the tabs that way unless a specific application is calling for this design.

LHinson:
I'm going to close this PR as changes should be introduced to core. Thanks!

PR #488, opened 2017-10-28
Added links to the showcase implementations

bleathem:
This resolves a portion of [PTNFLY-2599](https://patternfly.atlassian.net/browse/PTNFLY-2599)

I have staged the site built with this PR, you can see the result here:
https://bleathem.github.io/patternfly-org/stage/showcase-links/pattern-library/communication/inline-notifications/#code

Some pages do not have a HTML/CSS (jQuery) example, and so end up looking a little odd:
https://bleathem.github.io/patternfly-org/stage/showcase-links/pattern-library/data-visualization/heat-map/#code

PR #502, opened 2018-01-11
Adding links to showcase implementations w/ updates

amarie401:
A continuation of #488 to include https://patternfly.atlassian.net/browse/PTNFLY-2670 and new design updates from Gina https://redhat.invisionapp.com/share/X6EO0P77J#/screens

Notes: 

- The message for when there is no pf code available has been changed due to issues with adding the title/name of each pattern to the message dynamically from the markdown file (title isn't available) @LHinson is this the message you wanted to go with?

- Could you review and suggest any changes to the styling of this? Tried to mirror the design as closely as I could :) @andresgalante  @matthewcarleton 

screen shot 2018-01-11 at 12 55 19 pm

screen shot 2018-01-11 at 12 53 51 pm

screen shot 2018-01-11 at 1 25 04 pm


LHinson:
@amarie401 nice!

I would like to request a change for the message from "Help create a core example here" to "To contribute the code for this example, visit our GitHub repository." and the "GitHub repository" will be the words that link.

amarie401:
@LHinson will do!

source/_less/patternfly-site.less
```diff
@@ -1425,7 +1425,7 @@ th {
 }
 
 .page-tabs-header {
-  margin-bottom: (@grid-gutter-width + 5);
+  margin-bottom: 25px;
```
matthewcarleton: Any reason why you dropped the variable here?

source/_less/patternfly-site.less
```diff
@@ -1442,3 +1442,41 @@ th {
...
+  padding: 0 0 25px 0;
+  display: flex;
+  flex-wrap: wrap;
+
+  border-bottom: 1px solid #D5D5D5;
+}
+
+.pforg-code__showcase-link-item {
+  padding-right: 4rem;
+  display: block;
```
matthewcarleton: What is the display:block; for?

source/_less/patternfly-site.less
amarie401: I believe it was 45px with the variable and that leaves too wide of a gap between the header and the links and made it look odd, so I replaced it to 25px to make the gap less

source/_less/patternfly-site.less
amarie401: hmmm I actually forgot to remove that. I don't need it since I have display: flex; on the parent.  Will go ahead and remove.

source/_less/patternfly-site.less
amarie401: @matthewcarleton should I keep the variable and just do ```(@grid-gutter-width - 15);```?

matthewcarleton: 👍,That's a good question for @andresgalante I'm not sure what the approach is with the pf site.

dtaylor113:
Hi @amarie401, I hope to review this soon!

amarie401:
The design has it being linked to patternfly/patternfly but that would make sense to having it link to the contribute link. @LHinson thoughts?

I'm guessing when there is no code available for anything (core, or other repos) either we could keep the disabled code tab or have a separate "no example" message for when there are none at all. 

When there are no core examples it makes sense to have the tab enabled since there are other links to the other repos that may have those examples.

mcarrano:
Looks good to me also @amarie401 . Is there a way to stage and test this before we make this live on the site?

amarie401:
@mcarrano I'm not sure I'll be able to provide a stage link like in #488 since I have been having issues trying to do so. If you run it locally following the steps @dtaylor113 provided, then you'll be able to test it that way.

mcarrano:
I did manage to get this running locally on my MacBook.  Yay! Thanks, @dtaylor113 for the instructions. There were a few false starts but I did manage to make this work.

The changes look great @amarie401. I think this will be a big benefit for developers who want to find JS implementations of patterns listed on the website. I did not check every page, but just randomly tried several links and made sure I could get to Angular, NG, and React showcase examples.

amarie401:
Awesome! Thanks for reviewing @mcarrano :)

LHinson:
@dtaylor113 I see what you are saying regarding the empty state pages. For now, let's hold off and get this PR merged as the changes seen here will be a good improvement and I'd like to see it done sooner than later. That said, I'll get a story filed to look into this and address it as another PR. Thanks for your review and help! 

FYI @amarie401 @mcarrano

amarie401:
@LHinson link and corresponding message have been updated for contributing to PF

screen shot 2018-01-15 at 10 41 35 am

amarie401:
@andresgalante could you give this a quick review? :)

source/_less/patternfly-site.less
```diff
@@ -1442,3 +1442,40 @@ th {
...
+}
+
+.pforg-code__showcase-link-list {
+  list-style: none;
+  margin: 0;
+  padding: 0 0 25px 0;
+  display: flex;
+  flex-wrap: wrap;
+
+  border-bottom: 1px solid #D5D5D5;
```
jeff-phillips-18: Are the patternfly color variables not available here?

amarie401: Hmm..looking through I don't see a color variable for this. Can double check.

andresgalante: black-300 is `#d1d1d1`, pretty close, I'd say that `d5d5d5` is a mistake

amarie401: @andresgalante should I leave the `d5d5d5` or change them all to black-300 because `d5d5d5` is being used in a few different places.

amarie401: @andresgalante I'm not sure what the approach is styling wise when it comes to the site. I'm thinking since the site has already been using #d5d5d5 in the ``patternfly-site.less`` then perhaps I should keep it as is to remain consistent rather than change them all and perhaps deviating from the design?

amarie401:
If all else is good, I believe this is ready for a merge.

source/_less/patternfly-site.less
jeff-phillips-18: Please file an issue to fix instances where the palette variables are not used.

LHinson: Issue has been filed here #507

PR #522, opened 2018-02-12
fix variable names in launcher example

### LINKED ISSUE #520: App launcher examples not showing icons or grid format
The examples are meant to show both a grid and list view. The grid is shown by adding a class of `applauncher-pf-block-list` on the `applauncher-pf` list item. The icons are not in the code example at all so they will have to be added. For reference here is the [test page](http://rawgit.com/patternfly/patternfly/master-dist/dist/tests/application-launcher.html) that shows the right examples

michpetrov:
Renaming variables to align with PatternFly master.

Close #520


matthewcarleton:
LGTM thanks @michpetrov !

PR #532, opened 2018-03-08
Adding context selector example to pf-org

amarie401:
Related to this jira story: https://patternfly.atlassian.net/browse/PTNFLY-2786
and this issue: https://github.com/patternfly/patternfly-design/issues/589

Changes introduced:

1. ~~Creating a `doc.md` file for dev notes (which will also need to be included as a variable (`code_doc` in the design repo in a separate PR)~~ Dev notes can be added in a separate PR.
2. Include `contextselector: true` to the current vertical navigation on patternfly and add that to `code.md` file example as an iframe.

* [Preview](https://rawgit.com/amarie401/patternfly-org/context-selector-example-dist/_site/pattern-library/navigation/context-selector/#code)

Issues/Questions: 

1. The navbar brand logo for pf disappears when `contextselector:true` is present. I think it may have something to with this part of the code in [patternfly](https://github.com/patternfly/patternfly/blob/master/tests/pages/_includes/widgets/layouts/navbar-vertical.html#L11). Is this okay for the site?

2. In the prettyprint reference markup I tried including patternfly's `vertical-navigation.html` for the code snippet but the code doesn't contain the relevant context selector classes that are added. Not sure if it's okay to keep it this way or if there needs to be another file added into pf core that org can consume with the relevant classes already added to the vertical navigation.

Thoughts? @matthewcarleton @dgutride @cliffpyles @dabeng

### DIFF
--- a/source/_includes/code/navigation/context-selector/code.md
+++ b/source/_includes/code/navigation/context-selector/code.md
@@ -1,0 +1,12 @@
+<h2 id="example-code-1">Example</h2>
+<div class="example-pf">
+    <iframe src="{{ site.baseurl}}/pattern-library/navigation/context-selector/context-selector-vertical-nav.html"
+      width="100%" height="650px;" scrolling="no" seamless></iframe>
+</div>
+<p>
+  <a href="https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/context-selector-vertical-nav.html" target="_blank">View full page example</a>
+</p>
+<p class="reference-markup"><a class="collapse-toggle" data-toggle="collapse" aria-expanded="true" aria-controls="markup-1" href="#markup-1">Reference Markup</a></p>
+<div class="collapse in" id="markup-1">
+  <pre class="prettyprint">{% capture markup_include %}{% include widgets/framework/context-selector.html %}{% endcapture %}{{ markup_include | xml_escape }}</pre>
+</div>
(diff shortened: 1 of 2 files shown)


amarie401:
@matthewcarleton this currently what I have for the context selector for org based off of what I can include from the `_includes/widgets` directory in pf-core.

[Context selector](https://rawgit.com/amarie401/patternfly-org/context-selector-example-dist/_site/pattern-library/navigation/context-selector/#code)

I am pulling from  pf-core's vertical-navigation with `contextselector:true` for the example but cannot include it for the code snippet since it wont render the html text for the context selector inside of the vertical-nav, so I am using the `_includes/widgets/framework/context-selector.html` as the code snippet instead.

Would you be able to take this from the test pages: https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/context-selector-vertical-nav.html and create a file in ` _includes/widgets/navigation` in pf-core which could make it simpler to consume and display the correct code snippet for the context selector inside of the vertical nav?

If using what I have now for the code snippet is okay then in the future dev notes could be added to note where to add the context selector class if we want to go that route.

matthewcarleton:
@amarie401 and I just talked about this. I think it makes more sense to just show the code for the context selector itself rather than bury it in the entire page's markup. With the addition of dev notes we can clarify that it belongs in the vertical navbar. @LHinson @mcarrano are you ok with this?

Issue #538, opened 2018-03-21
abstract all third-party dependencies to external link


amarie401:
There has been issues where the pf-org site does not deploy/update due to the deploy size being exceeded. More recently this has been caused by a moment.js version update. See Travis: https://travis-ci.org/patternfly/patternfly-org/builds/355880364#L1900

The proposed solution is removing the thirdy-party dependencies and using a CDN instead. 

This was done recently for bootstrap slider in this PR: https://github.com/patternfly/patternfly-org/pull/531

 We're pushing unused source as well:

./bootstrap-switch/src
./bootstrap-touchspin/src
./c3/src
./d3/src
./eonasdan-bootstrap-datetimepicker/src
./google-code-prettify/src
./jquery/src
./moment/src
./patternfly-bootstrap-treeview/src
./wowjs/src

@dlabrecq @dgutride @dabeng

dabeng:
Great idea :+1: Amiee

PR #540, opened 2018-03-22
fix(build): Abstract third-party dependencies to external cdn links

dabeng:
## Description
This PR is targeted to fix issue #538

The addition of third-party dependencies has put the site over the size our deploy scripts can handle. This moves those packages out and uses the cdn instead.

## Additional Info

- google-code-prettify is not published on cndjs
- wowjs is used in [index page](https://github.com/patternfly/patternfly-org/blob/master/source/_layouts/home.html#L11) of patternfly.org site
So I didn't move the above packages out from the package.json.


dgutride:
This looks good @LHinson

Issue #553, opened 2018-04-19
Update Login Code Example to Account for Updated Styling


mindreeper2420:
The Login code example currently on PatternFly.org will need to be updated to reflect the changes put forth by PR https://github.com/patternfly/patternfly/pull/1024. This moves the class `.login-pf` from the `<html>` element to the `<body>` element.

amarie401:
@matthewcarleton you want me to take this to update the dev notes?

matthewcarleton:
@amarie401 if you have the time. Just let me know if you need any help with it. Thanks!

amarie401:
@matthewcarleton sooo I'm not sure exactly what needs to be changed since in the current dev notes we only list `login-pf-page` and not `login-pf` does this need to be added? Might be better if you or @mindreeper2420 made the update to `doc.md` table.

matthewcarleton:
Hey @amarie401 this would be in addition to `login-pf-page`. @mindreeper2420 are you able to take care of this, I think you are more familiar with the specifics of it. Let me know if you want me to jump in though.

amarie401:
Since I'm not familiar with the specifics of it then it might be better if you took this one @mindreeper2420 and make the necessary changes in the `doc.md` table of the login-page.

mindreeper2420:
@amarie401 Sure, I'll take a stab at it. According to the `login.less` file, this is for the older pattern, so I'll clarify that in the `doc.md` update :)

amarie401:
Also FYI because the pf core build is currently broken after your login changes, org cannot pick up the latest changes so it will still show`.login-pf` class on the `<html>` element.

mindreeper2420:
PR submitted: https://github.com/patternfly/patternfly-org/pull/557. My first to patternfly-org, so please let me know if I missed something :)

PR #557, opened 2018-04-25
fix(login): Update login docs

mindreeper2420:
Update the login docs to account for the use of `.login-pf` and applying it to a page's `<body>` element

fixes https://github.com/patternfly/patternfly-org/issues/553

### DIFF
--- a/source/_includes/code/application-framework/login-page/doc.md
+++ b/source/_includes/code/application-framework/login-page/doc.md
@@ -2,6 +2,25 @@
   <thead>
     <tr>
       <th>Selector</th>
+      <th>.login-pf</th>
+  </thead>
+  <tbody>
+  <tr>
+    <td>Applied to</td>
+    <td><code>body</code></td>
+  </tr>
+  <tr>
+    <td>Required</td>
+    <td><strong>Yes</strong></td>
+  </tr>
+  <tr>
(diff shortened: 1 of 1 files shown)


source/_includes/code/application-framework/login-page/doc.md
```diff
@@ -1,4 +1,23 @@
...
+      <th>.login-pf</th>
+  </thead>
+  <tbody>
+  <tr>
+    <td>Applied to</td>
+    <td><code>body</code></td>
+  </tr>
+  <tr>
+    <td>Required</td>
+    <td><strong>No</strong></td>
```
matthewcarleton: wouldn't this be required?

amarie401: I would think it would be required..?

mindreeper2420: It is only required if they want the background image...should it be required, as that is the design pattern?

matthewcarleton: ya that's my thought. If it's removed it's white box on a white background, which looks pretty strange.

amarie401:
@matthewcarleton can you give another review so we can get this merged in?

Issue #495, opened 2017-11-22
Bootstrap-switch link is broken


ethan-rambacher:
On http://www.patternfly.org/pattern-library/widgets/#bootstrap-switch, the link to http://www.bootstrap-switch.org/ should be http://bootstrapswitch.com/

LHinson:
I'll take this one. Thanks for reporting.

PR #566, opened 2018-05-17
fix(link): Bootstrap-switch link is broken

dabeng:
This PR is targeted to fix issue #495

However, bootstrapswitch.com is also broken, we should use latest bootstrapswitch.site

@dgutride Please review this PR whenever you get the chance 😊


amarie401:
Since this was a simple fix, I'm going to go ahead and merge this. Thanks @dabeng

Issue #648, opened 2018-11-19
FAQ typos & dead links


katierik:
FAQ page, The "What does patternfly offer?" question. The links for "widget library" and "pattern library" are both broken.
<img width="889" alt="screen shot 2018-11-15 at 1 57 34 pm" src="https://user-images.githubusercontent.com/23172038/48728104-07d09500-ec02-11e8-8d22-7cb8f81d39d8.png">

The "how is patternfly addressing accessibility?" question there is a typo in the link for Styles: Color *Pallete* page"
<img width="1039" alt="screen shot 2018-11-15 at 5 30 54 pm" src="https://user-images.githubusercontent.com/23172038/48728155-26cf2700-ec02-11e8-9402-67178a24a2cf.png">

PR #666, opened 2018-12-11
fix(faq): fix links to widget and pattern library

mcoker:
fixes https://github.com/patternfly/patternfly-org/issues/648


LHinson:
Another straight forward one. Will go ahead and merge once check pass.

PR #683, opened 2019-01-30
Add Alerts+notifications

mcarrano:
Adds initial version of Alerts and Notifications design guidelines documentation for website.

Closes https://github.com/patternfly/patternfly-next/issues/299


mcarrano:
@katierik Would also welcome your comments on this one.

mcarrano:
@stacymcauliffe Thanks for the feedback. Definitely agree on the first point. That also occurred to me so I will change.

The terminology here is a tough one. I agree that it's a bit confusing, but in PF3 we talk about inline notifications and toast notifications, but the thing is an alert. We also have a Notification Drawer (that holds alerts!?). I was working from the premise that the component is an Alert, but it's used to present notifications. Does that make sense? I would be glad to discuss further.

mcarrano:
@stacymcauliffe I did not get to fixing the images, but I did reorganize and change the terminology as we discussed (Notifications => Alerts). Let me know if this is better and I will continue working on it.

mcarrano:
@stacymcauliffe I made another round of updates. This may not be perfect, but hopefully good enough for initial training. Please review and merge if you think it is ready.

PR #685, opened 2019-02-01
refactored forms images

stacymcauliffe:
Refactored forms images to align with new image creation standards


packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/forms/design.md
```diff
@@ -41,78 +42,67 @@ The PatternFly components listed in the following sections can be used in a numb
...
 
-### Choose the right data input
-There are multiple components that enable a user to make a simple selection. To learn more about data input, visit the [text and data input design guidelines](/design-guidelines/usage-and-behavior/data-input).
+Learn more about data input in the [text and data input design guidelines](/design-guidelines/usage-and-behavior/data-input).
 
 ### Provide the right kind of help
-There are multiple ways to provide help on forms.
+There are a few different ways to provide help on forms. The type of help you choose will depend on the content you need to relay to your user.  
 
 **Field level help**
```
mcarrano: I'm not sure this approach is still valid. I know we had multiple discussions about accessibility of putting the Help icon to the right of the field. @jgiardino @mceledonia can you remind me of where we ended up on this?

jgiardino: The recommendation from deque is that if a form field includes a button that provides help for that form, then the button should precede the input field in the DOM, and the label for the button should clearly associate it with the input (e.g. repeat the input label in the button label). The reason for this is so that screen reader users know that help is available for an input before they get to the input.

stacymcauliffe:
@katierik @mcarrano - made additional updates to forms docs based on Katie's review comments.

PR #703, opened 2019-02-07
Adding toolbar documentation

bmignano:
Add documentation for toolbar.

Closes patternfly/patternfly-next#435


packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
```diff
@@ -0,0 +1,94 @@
...
+2. **Filters**: PatternFly supports a variety of common filter types that can be used alone or in combination to construct a custom filter.
+3. **Sort (optional)**: A sort component may to added when this functionality is needed. Tables have built in sorting, but card views and data lists may require toolbar sorting.
+4. **Global actions**: Any global actions that apply to selected data items are included left-aligned following filter and sort controls. These may be exposed as buttons or collapsed into a kabob menu. No more than two items should be exposed as buttons. Button labels should be limited in length to preserve horizontal space.
+5. **Pagination**: When pagination is supported, a pagination component will be included. Pagination will always be right aligned in the toolbar.
+
+## Related components and demos
```
mcarrano: I like the way you've formatted this. @stacymcauliffe should we adopt this as the standard or list inline like we've been doing? In any case I think we should be consistent.

packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
```diff
@@ -0,0 +1,94 @@
...
+
+**Toolbar with filter chips**
+![filterchips](img/toolbar-filter-chips.png)
+When the current settings of the controls are not sufficient to convey applied filters, filter chips should be used.
+1. **Filter chips**: When filters are applied the toolbar will expand and filter chips will be displayed on the second row. Users can clear all filters by clicking the “Clear filters” link, and the toolbar will collapse back into one row.
+
+**Custom toolbar**
+![custom](img/custom-toolbar.png)
+When the current settings of the controls are not sufficient to convey applied filters, filter chips should be used.
+1. **Filter chips**: When filters are applied the toolbar will expand and filter chips will be displayed on the second row. Users can clear all filters by clicking the “Clear filters” link, and the toolbar will collapse back into one row.
```
mcarrano: This feels redundant. Maybe it is not needed since it is stated above?

packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
```diff
@@ -0,0 +1,94 @@
...
+
+Expanded
+
+![open](img/mobile-expanded.png)
+
+Footer with pagination
+
+![footer](img/mobile-footer.png)
+
+The toolbar adapts to smaller viewport sizes by collapsing or hiding elements that take significant space, including filters and actions. On mobile devices, when used, pagination will be available from the footer, only.
```
mcarrano: I would move this so it appears above the mockups. Maybe show this in the context of a mobile screen?

packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
bmignano: Oops, you're absolutely right. This was an accidental repetition. I can replace this with a more accurate description for "Custom toolbar".

packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
bmignano: There's a mobile screen in the examples section -- should I add another here, or perhaps a schematic?

packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
```diff
@@ -0,0 +1,94 @@
+# Toolbar
+The toolbar contains multiple components that allow a user to manage and manipulate a data set. Data can be presented in a variety of views (table, list, card, etc.). PatternFly supports multiple toolbar variations that address a variety of use cases.
+
+## Elements
+![toolbar](img/toolbar.png)
+
+The following elements may be included in a typical toolbar.
```
stacymcauliffe: You can substitute "can" for "may" and use contractions to make it feel less formal. So, "The following elements can be used in a typical toolbar." 

Also, I wasn't sure what a data presentation component would be, so maybe clarifying with an example? "Bulk selection can be added when the functionality isn't available from the component you're using to present the data. For example, you might need to include a bulk selector for a card view or data list, but a table will enable bulk selection by default." 

"Bulk selection allows the user to select or deselect all items on a page and provide an indeterminate state that notifies the user when only some items are selected. Bulk selectors should always be the left-most element in a toolbar. When providing a bulk selector in a list view, it should be aligned with the checkboxes.

packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
stacymcauliffe: Agree. It's easier to read, so I think that makes up for the potential scroll-factor. I can update all of the docs to be formatted like this. 
That being said, how would you feel about moving this section to the very bottom of the documentation? Let me know. I just feel like it could be secondary information for a designer.

packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
```diff
@@ -0,0 +1,94 @@
...
+
+**React**
+* [Badges](/documentation/react/components/badge)
+* [Buttons](/documentation/react/components/button)
+* [Chip](/documentation/react/components/chip)
+* [Dropdowns](/documentation/react/components/dropdown)
+* [Toolbar demo](/documentation/react/demos/toolbar)
+
+## Usage
+### When to use
```
stacymcauliffe: Toolbars enable users to manipulate data in views where built-in controls aren't already available, like list views or card views. You can choose from multiple toolbar variations to suit your use case, and further customize your toolbar using any of the components listed under [Related components and demos](#related-components-and-demos).

packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
```diff
@@ -0,0 +1,94 @@
...
+The toolbar should be used when users may need to manage and manipulate a data set, whether it is displayed in a table view, list view, card view, or any other view that requires manipulation and does not already offer controls to do so. There are multiple toolbar variations in order to address a variety of use cases. The toolbar may be further customized by using a unique set of components from the related components list.
+
+### When not to use
+The toolbar should not be used when the data view already includes the necessary controls to manipulate the data set.
```
stacymcauliffe: Don't provide a toolbar for any data view that already includes the controls needed to manipulate a given data.

packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
stacymcauliffe: Agree

packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
bmignano: One thing we could potentially do to minimize the scroll-factor is use a 2x2 table so that the Core and React lists are side by side rather than stacked. Just an idea!

mcarrano: I think that we considered that earlier and decided not to do a table because it called to the mis-alignment between Core and React. But I'll defer to @stacymcauliffe on this point.

packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/toolbar/design.md
mcarrano: Yeah, or maybe this does not even need to be in both places. The responsive behavior is baked into the component so I think the only value here is to show a designer what to expect if they are on mobile. The example does that pretty well. You could add any text or explanation there.

bmignano:
@mcarrano @stacymcauliffe Thanks for your comments! I've updated the documentation accordingly.

PR #704, opened 2019-02-11
Images, links, and button content

### LINKED ISSUE #700: Image fixes across all docs
Currently, there are a few places on the site with broken images. These should be resolved.

stacymcauliffe:
Made updates to images and content for nav and buttons docs. 

Closes #700


stacymcauliffe:
@mcarrano - made additional updates to the buttons doc based on our convos yesterday and also took a stab at adding that rationale for why buttons are placed differently depending on content. Let me know what you think. Thanks!

mcarrano:
This looks great @stacymcauliffe . I like it much better. The only thing is it seems like we lost links to the related components?

stacymcauliffe:
@mcarrano - added components - I think it's just the button component for both react and core

PR #711, opened 2019-02-14
updated getting started content based on review from a. glass

stacymcauliffe:
Updated get started content with review comments from Amy Glass. Also updated URLs.

### DIFF
--- a/packages/patternfly-4/src/content/get-started/about-pf.md
+++ b/packages/patternfly-4/src/content/get-started/about-pf.md
@@ -2,7 +2,7 @@
 path: "/get-started/about"
 ---
 # About PatternFly 4
-PatternFly is an open source design system that was created to enable consistency and usability across a wide range of applications and use cases. PatternFly provides clear standards, guidance, and tools to help designers and developers work together more efficiently to build better user experiences.
+PatternFly is an open source design system created to enable consistency and usability across a wide range of applications and use cases. PatternFly provides clear standards, guidance, and tools that help designers and developers work together more efficiently and build better user experiences.
 
 ## Basic structure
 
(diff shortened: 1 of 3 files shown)


packages/patternfly-4/src/content/get-started/about-pf.md
```diff
@@ -2,43 +2,40 @@
...
 
-### Components
-Components, like buttons and alerts, can be assembled together to build applications.
-[View Core components](/documentation/core/) | [View React components](/documentation/react/)
+### Demos
+Demos use components and layouts in combination to show you how to build more complex structures and application views.
+
+[View components, layouts, and demos in HTML/CSS](/documentation/core) | [View components, layouts, and demos in React](/documentation/react)
 
 ## Style and design
```
mcarrano: Should we align this with the way that the website is now organized? So we have Design Guidelines as a top level category and under that Styles, Usage and Behavior, and Content.

packages/patternfly-4/src/content/get-started/gettingstarted-design.md
```diff
@@ -2,7 +2,7 @@
 path: "/get-started/designers"
 ---
 # Getting started for designers
```
mcarrano: I think this is OK for now, but could be made much more robust. Maybe some of the materials being developed now for training can be dropped in here. Also, there is no mention of the Sketch library.

stacymcauliffe: Agree, although I dont think we can mention the sketch library yet bc it's not ready to be made available - that's the feedback I got from Gina and Kyle.

stacymcauliffe:
@mcarrano - I made your recommended changes to the design guidelines section. I have some calls set for later today and early next week to chat with devs about the CSS variables work and also about building our the react section a bit further. I'd be comfortable with merging this now since I have an open issue for the additional dev get started work, but I am also fine with waiting. 

Thanks!

PR #732, opened 2019-03-01
Adding content about header alignment

katierik:
As per a question that came up, I added a sentence describing that column headers should align with the content (aka if your content is center aligned, center align the header as well). 
Open to suggestions of where it could go - I didn't include a lot of detail because there was already a lot going on in that list item.

### DIFF
--- a/packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/lists-and-tables/design.md
+++ b/packages/patternfly-4/src/content/design-guidelines/usage-and-behavior/lists-and-tables/design.md
@@ -77,7 +77,7 @@ Two variants are available for Data Tables.
 
 1. **[Toolbar:](../toolbar/design.md)** The toolbar sits above the list and contains controls for manipulating table data. Common actions include filtering, sorting, and pagination.
 2. **Select all:** When present, selects all items in a table. If pagination is being used, this will only select items on the current page.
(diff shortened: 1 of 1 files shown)


mcoker:
The [compact table example](https://pf4.patternfly.org/components/Table/examples/) shows center aligned content in a table. We currently have a `.pf-m-center` class that you can add to elements that will center contents in a table cell. However that specific example (the column of checks) is changing a bit in https://github.com/patternfly/patternfly-next/pull/1472 because the checks are center aligned on desktop and left aligned on mobile, so we have a `.pf-c-table__icon` class that we apply to those table cells that aligns the content differently between desktop and mobile. We're leaving the `.pf-m-center` class in the component so it can still be used. However, we don't have a `.pf-m-right` or `.pf-m-left` class to align content left or right - should we?

mcarrano:
Thanks for the clarification on this @mcoker . So I take it that right now content will left-align by default but can be center-aligned by applying the .pf-m-center class, is that right? So in that case I assume we could not right align content. I could see rare cases where we might want this, like a column of numbers, but don't feel strongly about it.

As for this PR, I will go ahead and merge since it only affects the documentation. If we decide to update the CSS, this can be opened as a separate issue. Thanks for catching this @katierik !

mcoker:
@mcarrano correct!

PR #734, opened 2019-03-05
Added images to a folder and updated path in text file.

jknilans:
@mcarrano I cleaned up the contents by moving the images into a folder. Is this okay now? Thanks


mcarrano:
Actually, looks like a conflict so I can't merge. @jschuler can you take a look at this? I don't see what the problem is. @jknilans was just trying to move image files to a folder.

jschuler:
It's because the folder location has changed, it is no longer in packages/patternfly-4/src/content/... but moved up to packages/patternfly-4/content/...
Need to rebase first, should be something like
```
git fetch origin
git merge origin/gatsby2
```
There will be some rename conflicts because file locations had changed, but the login markdown file should have merged correctly in patternfly-4/content.
At this point you can move the img folder into the correct location under patternfly-4/content and manually delete the other images that are still hanging around.
Then delete your patternfly-4/src/content folder as it is no longer in use. Then you should be good to do another 
```
git add patternfly-4/content/*
git commit -m 'rebase'
git push
```

jknilans:
@mcarrano is this okay to merge now?

mcarrano:
Yep. Looks good. Thanks @jknilans ! Will merge.

PR #744, opened 2019-03-06
Adding filters documentation

bmignano:
Adds documentation for filters.

Closes https://github.com/patternfly/patternfly-next/issues/571


packages/patternfly-4/content/design-guidelines/usage-and-behavior/filters/design.md
```diff
@@ -0,0 +1,217 @@
+# Filters
```
jschuler: Can you add this at the top of the file
```
---
path: "/design-guidelines/usage-and-behavior/filters"
---
```
Then in the usage-and-behavior folder you will also see a json file, need to add the new navigation item in there
```
{
        "text": "Filters",
        "path": "/design-guidelines/usage-and-behavior/filters"
      }, 
```

bmignano: All set 👍

packages/patternfly-4/content/design-guidelines/usage-and-behavior/filters/design.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### When to use
+Use a complex faceted filter when you have limited space and multiple attributes with a variety of input types, for example, and search field and checkbox selects.
+
+#### Behavior
+Filter queries are applied when the user selects or deselects an item in the list, or types a query in the input field and presses Enter.
+
+
+
+## Related components and demos
```
mcarrano: We should include Toolbar in the Core list. Looking at this raises a couple or questions to me: 

* Is it helpful here to list every component that can be used? The list here is mostly correct, but for example does a user looking at this page want the exhaustive list of all sub-components or would I just want to have a link to say the Filter Demo where I can see live examples and then from there see what's required to implement?
* This is the first instance where the documentation is actually out ahead of implementation. On the React side especially, there are definitely more components used than are listed here. Many of the required components don't exist yet. There may be many cases like this in the future. Should we be listing components that are still WIP? If not, it gives the impression that this is a complete list when it's not. How to we make it painless to keep this list updated when components are released? Another option is to not publish these guidelines in the nav until development work is done.

@rachael-phillips @stacymcauliffe interested in your thoughts.

rachael-phillips: Hi @mcarrano I agree, I don't think we should list any components that haven't been released. Also, perhaps we'll leave the design guidelines off the site until the component is ready which may help clarify what to list. I have less insight on the first bullet. I think linking to the Filter Demo is the most simple way to provide pointed guidance.

mcarrano:
@jschuler can you give this a look? Seems that I cannot merge due to a conflict in the navigation JSON file, but I don't see what the problem is.

mcarrano:
OK. I think I was able to resolve the conflict so I will proceed with merging.

PR #751, opened 2019-03-13
Add initial short descriptions for demo and testing.

mcarrano:
These are the initial set of short descriptions (I just did the A's) to test how these can be inserted into the related component pages.


mcarrano:
@jschuler and @stacymcauliffe As we discussed yesterday, I created a few short descriptions in the repo to test integration with component pages. Let me know if these work or if we should do something different. @stacymcauliffe also let me know if the formatting and capitalization makes sense. I can adjust, if needed.

jschuler:
Thanks @mcarrano , i'll merge these in so i can start playing with these

Issue #739, opened 2019-03-05
HTML/CSS: Implementing the Component page


rachael-phillips:
Adam will implement the visual designs for the live site when Liz completes the high fidelity mockups.

rachael-phillips:
Adam and Joachim will meet to go over site structure.

mindreeper2420:
@rachael-phillips I believe that this is complete - can you verify?

rachael-phillips:
Yes, we can close this issue.

PR #752, opened 2019-03-18
fix(typography): update site typography to use Overpass

mindreeper2420:
Update the typography of the website to use Overpass, rather than relying on the user's local font family (ie San Francisco for macOS, Arial for Windows 10, etc.). Force the cascading font family settings, per PatternFly 4 specs on typography.


mindreeper2420:
related to Issue https://github.com/patternfly/patternfly-org/issues/739

PR #766, opened 2019-03-21
Update component links in Filters doc

mcarrano:
As was discussed, we want to find a more effective method for displaying component cross-links on the website. These are intended to help readers of a design guideline document know where to find related components for implementation. Previously we had a long list at the end of the document that did not seem very helpful.

This is an initial attempt at this using the Filters document that leverages multiple core and React components. Links have been moved to be more inline with the elements section.

### DIFF
--- a/packages/patternfly-4/content/design-guidelines/usage-and-behavior/filters/design.md
+++ b/packages/patternfly-4/content/design-guidelines/usage-and-behavior/filters/design.md
@@ -48,13 +48,19 @@ The attribute-value filter gives the user the ability to specify an attribute-va
 
 ![image7](img/attribute-value-date-picker.png)
 
-The following elements can be used in an attribute-value (textbox) filter.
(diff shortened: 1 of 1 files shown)


mcarrano:
@stacymcauliffe @jschuler @rachael-phillips This is a follow-up on our discussion from last week. I'm open to suggestions about how well you think this approach works. It was difficult to find an ideal solution. Components are placed inline using the "quote" tag in Markdown. Formatting options seem limited without writing custom HTML. I was not sure the best way to link components where thee were different Core and React versions with the same name. @jschuler thoughts? In our meeting you mentioned the idea of a badge, but was not sure how to do that.

mcarrano:
I am actually going to close this pull request. We need to find a better way to connect/link design guidelines with components and demos, but I'm not sure this is the right approach. We can revisit again in the future.

PR #768, opened 2019-03-25
added date and time standards

stacymcauliffe:
Added date and time information to the content standards

### DIFF
--- a/packages/patternfly-4/content/design-guidelines/content/grammarandterminology.md
+++ b/packages/patternfly-4/content/design-guidelines/content/grammarandterminology.md
@@ -7,6 +7,7 @@ path: "/design-guidelines/content/grammar-and-terminology"
 [Abbreviations](#abbreviations)
 [Acronyms](#acronyms)
 [Capitalization](#capitalization)
+[Date and time](#date-and-time)
 [Punctuation](#punctuation)
 [Truncation](#truncation)
 
@@ -48,16 +49,6 @@ path: "/design-guidelines/content/grammar-and-terminology"
 | N/A           | Refers to data not available. Applies to tables, list views, card views, etc. |
 | --            | Refers to data not applicable. It is different than N/A that data would never be available for the object. |
 
-### Date and time
-
-When referring to time zones, use [http://www.timeanddate.com/time/zones/](http://www.timeanddate.com/time/zones/):
-
-| Symbol | Period |
-| --- | --- |
-| AM | Ante meridiem (before noon) |
-| PM | Post meridiem (after noon) |
-
-
 ## Acronyms
 
 ### General rules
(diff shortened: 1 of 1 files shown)


packages/patternfly-4/content/design-guidelines/content/grammarandterminology.md
```diff
@@ -73,6 +64,63 @@ Use sentence-style capitalization for all text elements, **except:**
...
+Whenever possible, it is best to enable users to customize the format in which date and time are displayed in the interface.
+
+These date and time standards are based in American English, but date and time is formatted differently depending on language.
+
+### Date
+Use the default US style of **Month DD, YYYY**. Use **DD Month YYYY** for non-US locations.
+
+Whenever possible, write out the full name of the month. If you have space constraints, use the 3 letter abbreviation for the name of the month.
+
+Jan
```
mcarrano: I think you intend line breaks here and below but they are not showing in the GitHub preview.

packages/patternfly-4/content/design-guidelines/content/grammarandterminology.md
```diff
@@ -73,6 +64,63 @@ Use sentence-style capitalization for all text elements, **except:**
...
+9:38 PM
+
+**Absolute and relative time**
+The use of absolute or relative times stamps will depend on your user needs and use case. If your user is primarily interested in understanding the exact date and time that an event occurred, use an absolute timestamp.
+
+If users are interested in how long ago an event occurred, use a relative timestamp.
+
+For consistent timestamp formatting, we recommend using the following examples:
+
+| Timeframe  | Usage |
```
mcarrano: I'm not exactly sure what this is recommending. Are you saying that for an event that occurred within the past hour, I would use relative time, but then switch to absolute time after an hour passes? I agree that this is a tough call regarding the right approach.

packages/patternfly-4/content/design-guidelines/content/grammarandterminology.md
```diff
@@ -73,6 +64,63 @@ Use sentence-style capitalization for all text elements, **except:**
...
+| ------------- | ----- |
+| 0-60 seconds | Just now |
+| 1-60 minutes | 30 mins ago |
+| 1-24 hours | 9:30 AM |
+| Yesterday | Yesterday, 9:30 AM |
+| In the last week | Thursday, 9:30 AM |
+| 1 week - 1 year | 21 Jan, 9:30 AM |
+| 1 year + | 21 Jan 2019 |
 <!--## Links
```
katierik: Can we add something about time zones?

Time can be displayed in the user's time zone or in UTC. 
UTC is recommended for cases where people may be collaborating across time zones (for instance, tracking incidents on a server). 
The user's time zone is recommended when the user has the option of switching to/from any time zone and UTC in settings.

packages/patternfly-4/content/design-guidelines/content/grammarandterminology.md
katierik: I'm plugging my modification to this relative time here! It is more similar to [momentjs](https://momentjs.com/)
https://docs.google.com/document/d/1hivSfe48717VhlLRtjyqReS5twbSLuPC6MISJt3MTG4/edit?usp=sharing

Xnip2019-03-28_15-52-48

stacymcauliffe:
Made requested changes - let me know if I missed anything! Thanks friends

packages/patternfly-4/content/design-guidelines/content/grammarandterminology.md
KendraMar: PROBABLY a dumb question, but...

when Cloud Services apps indicate that a rule failed on a server in france and a server in China, I'm assuming it reports the finding using each server's time zone and not the time zone of the Cloud Services SaaS app (wherever it lives?). Just curious.

KendraMar: And I am trying to understand how important that is, use case/customer goal wise...could be important for Rules, where we have performance hits people might use for troubleshooting (eg, the problem happened during this time window when we were running reports on this server...)

packages/patternfly-4/content/design-guidelines/content/grammarandterminology.md
```diff
@@ -115,12 +129,13 @@ For consistent timestamp formatting, we recommend using the following examples:
 | Timeframe  | Usage |
 | ------------- | ----- |
```
mcarrano: Does this table apply specifically for relative timestamps only? I think it does. Maybe the text above should say something like, "When reporting relative time, follow these examples:"

packages/patternfly-4/content/design-guidelines/content/grammarandterminology.md
mcarrano: Agree that it would be good to explain how to represent this. Either UTC or adding the time zone as a suffix, like "EDT, PDT, etc"

katierik: ASAIK - All the cloud services stuff is saving their time on the backend in UTC. So to display it in the user's timezone, there's some transformation that would need to happen.

stacymcauliffe:
Made additional updates based on feedback. @mcarrano - I added an example of a time stamp that uses EST. I'm not sure what additional information you want me to include about how to represent this. Let me know if I need more details. Thanks.

mcarrano:
This looks great @stacymcauliffe I will go ahead and merge.

Issue #738, opened 2019-03-05
HTML/CSS: Implementing the homepage


rachael-phillips:
Adam will complete the homepage based on Liz's high fidelity mockups.

rachael-phillips:
Closing this because work has been completed!

PR #773, opened 2019-03-27
fix(homepage): update homepage from reviews

mindreeper2420:
Updates to the homepage based off of reviews.

- Adjust positioning of the individual SVG elements on in-between screen sizes
- Fix background image bleed when resizing the screen
- Apply a solid background on smaller screen to eliminate visual clutter and improve load times
- Set inline SVGs to use `<object>` inside of empty `<div>` containers
- Fix footer wrapping on smaller screens


mindreeper2420:
related to https://github.com/patternfly/patternfly-org/issues/738

PR #786, opened 2019-04-02
add component descriptions

jschuler:


dgutride:
@redallen  - can we close this in favor of the new description work you did in react?  we'll need to replicate that in pf-next documentation also

PR #805, opened 2019-04-08
added empty state docs

stacymcauliffe:


stacymcauliffe:
@mcarrano - updated the images and removed the variations section of this doc

mcarrano:
@stacymcauliffe your latest changes look so much better! I like seeing the different empty state examples in context. Just one question - is there a reason we would use a card in the background or not? I'm seeing that in the last example (success) you are placing the empty state inside of a card with a white background and in other cases you do not. What's the difference?

stacymcauliffe:
Good question @mcarrano - I guess I was thinking that the success message might be displayed on whatever the user just finished interacting with.

mcarrano:
Yes, I guess that's true @stacymcauliffe . Maybe instead of this include a completion screen as in my Wizard doc here: https://docs.google.com/document/d/1rgAXIBFGfeDwHTjVvHvKzD5Zqtro6kQkpt3PJvK0kkg/edit?usp=sharing  I can send that image.

stacymcauliffe:
Creating new PR in master

Issue #801, opened 2019-04-06
Bug: line appearing on HTML examples components page


rachael-phillips:
There is a line appearing on the HTML example components page that does not appear in Liz's wires. Here is an example of the line we should remove: <img width="564" alt="Screen Shot 2019-04-05 at 10 23 21 PM" src="https://user-images.githubusercontent.com/40275806/55663812-75681100-57f1-11e9-8bb6-eee2d6154697.png">

mindreeper2420:
Related to PR https://github.com/patternfly/patternfly-org/pull/764

PR #810, opened 2019-04-10
fix(docs): remove border between sections

mindreeper2420:
fixes https://github.com/patternfly/patternfly-org/issues/801


jschuler:
Looks like you need to rebase @mindreeper2420

PR #813, opened 2019-04-11
add pf4 banner

christiemolloy:
Not sure if there is an issue to close here?

@jschuler I put the banner at the top of the page but it is still second to that top navigation banner, but I couldn't work out where  it was coming from. If you could help me here that would be great.

Also had to add the stylesheet for the breakpoint styles.

### DIFF
--- a/packages/patternfly-4/src/components/layout.js
+++ b/packages/patternfly-4/src/components/layout.js
@@ -5,6 +5,7 @@ import Header from './header';
 import Footer from './footer/footer';
 import {
   Brand,
+  Button,
   Nav,
   NavItem,
   NavList,
@@ -22,6 +23,8 @@ import {
 } from '@patternfly/react-core';
 import { Location } from '@reach/router';
 import brandImg from '../images/PatternFly_logo.svg';
+import {TimesIcon} from '@patternfly/react-icons';
+import Styles from '../components/banner.scss';
 
 class Layout extends React.Component {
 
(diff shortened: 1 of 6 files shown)


jschuler:
So actually it looks like this should be added to
packages/patternfly-4/src/components/layout.js
Line ~120 over the <Header>
Can you add it there? You can create a new banner.css file in the same location and import it at the top of the file.
Once there, the banner will show up on every page, I will add a conditional after you make the change to make it only show on the home page.

christiemolloy:
@jschuler awesome just moved it over!

packages/patternfly-4/src/pages/index.js
```diff
@@ -11,9 +11,10 @@ import {
   Text,
   TextVariants
 } from '@patternfly/react-core';
-import { ArrowRightIcon } from '@patternfly/react-icons';
+import { ArrowRightIcon} from '@patternfly/react-icons';
 import orb from '../images/PF4_logo.svg';
 import principles from '../images/PF4_WIREFRAME.png';
+import { bold } from '../../../../node_modules/ansi-colors';
```
jschuler: Is this needed? And can probably undo the change on line 14

packages/patternfly-4/src/components/layout.js
```diff
@@ -22,6 +23,8 @@ import {
 } from '@patternfly/react-core';
 import { Location } from '@reach/router';
 import brandImg from '../images/PatternFly_logo.svg';
+import {TimesIcon} from '@patternfly/react-icons';
+import Styles from '../components/banner.scss';
```
jschuler: could just be
```suggestion
import '../components/banner.scss';
```

jschuler:
Included in https://github.com/patternfly/patternfly-org/pull/816

PR #827, opened 2019-04-16
accessibility documentation

jschuler:
Just straight up copied from pf4.patternfly.org

Needs any edits or styling updates? @rachael-phillips @stacymcauliffe 

Preview:
http://pf-a11y.surge.sh/get-started/accessibility

### DIFF
--- a/packages/patternfly-4/content/get-started/accessibility.md
+++ b/packages/patternfly-4/content/get-started/accessibility.md
@@ -1,0 +1,169 @@
+---
+path: "/get-started/accessibility"
+---
+# Accessibility Guide
+
+*Please note, this guide is a work in progress and will be updated regularly. We welcome your comments and feedback.*
+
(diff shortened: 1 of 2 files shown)


stacymcauliffe:
My main comment about this doc would be the readability of the tables. The tag formatting is gone and it's a bit hard to follow content across each line item. I'd leave it to @jgiardino  to comment on the completeness of the content, but always happy to review for style.

jgiardino:
Thanks for commenting on that, @stacymcauliffe. I agree the styles on the table need some help. I'm open to suggestion on that, but also happy to make suggestions. 

The list styles are a little odd too (e.g. sub-bullets are closer to the next parent bullet than to their parent bullet).

But the contents all seem to be there.

rachael-phillips:
@christiemolloy is this something we can get up with styling on for May 3? You and I can pair to give this page some help early next week if you are open.

jschuler:
Looks like A11y guide was merged in a different PR

PR #815, opened 2019-04-12
chore(dialogs): add dialog docs

jknilans:
This PR adds docs related to modals and dialogs.

This should help close https://github.com/patternfly/patternfly-next/issues/1179

Updates include:

- screenshots now reflect left-aligned buttons
- added message dialog content to modal dialog documentation
- updated path

PR #839, opened 2019-04-23
Update Modal to Modal Message Dialogs

mcarrano:
This PR replaces #815 to update the Modal design guidelines with message dialog content.


mcarrano:
Need to do this over. File content is not correct.

PR #840, opened 2019-04-23
rebased modal msg dialog

jschuler:


stacymcauliffe:
@jschuler @mcarrano - I'm seeing a few things I might update, but Josh is on paternity leave. What would be the simplest path forward? Do we want to merge this and then I can submit a new PR afterwards to address my changes?

mcarrano:
Yes, @stacymcauliffe I think it will be easiest to just merge this and then just send a new PR. I also wondered if we should change the label in the nav. Right now it's still "Modal." But I think we can do that as part of a follow-up PR.

If you can approve this, then I can go in and merge.

stacymcauliffe:
Approved, thanks Matt

Issue #802, opened 2019-04-06
Implement Styles Pages


rachael-phillips:
There are 4 styles pages that we will need to create for launch

- Spacers
- Typography
- Colors
- Iconography

There pages have all been added to the Drive: https://drive.google.com/drive/u/1/folders/1MaI3Sewes8h4fxD2h4P6R6cZuRQIBYBf

mindreeper2420:
Once the four referenced PRs are merged, this story will be complete.

mindreeper2420:
All PRs merged - marking as complete.

PR #838, opened 2019-04-23
All chart stuff

gdoyle1:
Close #720 #747 

Added the chart color system to the styles section and added charting design guidelines

PR #846, opened 2019-04-24
feat(colors): implement Colors page

mindreeper2420:
Implements the Light Background portion of the Colors page.

download

----

related to issue https://github.com/patternfly/patternfly-org/issues/802


christiemolloy:
For these cards, in the Sketch file it looks like the size of the card is determined by the content, not sure what you did for this but looks like there is a lot of extra space on the top and bottom.

Screen Shot 2019-04-25 at 4 01 41 PM
Screen Shot 2019-04-25 at 4 01 45 PM

christiemolloy:
This is the sketch version for this section, content inside looks different  than what you have currently:
Screen Shot 2019-04-25 at 4 04 28 PM

christiemolloy:
Just spoke with @gdoyle1 about how the width of these cards should be. Can the content in the first colummn be 50% of the page, and then the content in the second column be 50% of the page. That way all of the "white cards" have the same width and the content just fits inside.

Screen Shot 2019-04-25 at 4 10 29 PM

mindreeper2420:
> For these cards, in the Sketch file it looks like the size of the card is determined by the content, not sure what you did for this but looks like there is a lot of extra space on the top and bottom.
> 
> Screen Shot 2019-04-25 at 4 01 41 PM
> 
> Sketch:
> Screen Shot 2019-04-25 at 4 01 45 PM

@christiemolloy the card sizing is based off of the grid in this implementation, I can change the wrapper to only go around the content, rather than the block.

mindreeper2420:
> This is the sketch version for this section, content inside looks different than what you have currently:
> Screen Shot 2019-04-25 at 4 04 28 PM

@christiemolloy A lot of the content didn't actually match what the real values are in the code. I spoke with @gdoyle1 yesterday about that, and the values will be updated to match the Sketch file once PatternFly code is updated with the new colors and variables.

mindreeper2420:
> Just spoke with @gdoyle1 about how the width of these cards should be. Can the content in the first colummn be 50% of the page, and then the content in the second column be 50% of the page. That way all of the "white cards" have the same width and the content just fits inside.
> 
> Screen Shot 2019-04-25 at 4 10 29 PM
> 
> I drew up what this would look like from @gdoyle1 sketch file, and we could have padding around the content.
> 
> Screen Shot 2019-04-25 at 4 12 46 PM

@christiemolloy They are broken into 6 columns each, but I don't have them set at the same width all the way through each one. I'll update that and give them room to flow better.

mindreeper2420:
@christiemolloy @gdoyle1 Here is the updated version with the split columns and adjusted spacing in the right-hand column boxes. 

download (1)

mindreeper2420:
Updated screenshot with latest fixes:

download (2)

gdoyle1:
@mindreeper2420 This all looks great!! Where it says "this copy needs to be updated" under the typography section, can you replace it with "Lorem ipsum". And where the chart colors are presented, can you add "Learn more about [colors for charts](patternfly.org/v4/design-guidelines/styles/colors-for-charts)." underneath the palette? (Like you see with icons section)

mindreeper2420:
@gdoyle1 Updated with "Lorem ipsum" and I've added the "Learn more about colors for charts.", but that link does not exist yet. I believe your PR (https://github.com/patternfly/patternfly-org/pull/838) needs to be merged first?

gdoyle1:
@mindreeper2420 Yup you're right!

PR #992, opened 2019-04-29
Attempt to find matching component page when switching frameworks

### LINKED ISSUE #812: Switching frameworks (React <> HTML) in Documentation defaults back to first component
If I am viewing Chip Group in React (https://v2.patternfly.org/documentation/react/components/chipgroup) and then switch the toggle to HTML, it defaults back to the first component in the left nav (About Modal) instead of to https://v2.patternfly.org/documentation/core/components/chipgroup

### LINKED ISSUE #965: Documentation section's framework toggle button is confusing
Scenario:
On the [Forms design guideline page](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms) there is a view forms demo link that redirects the user to the [Core Forms page](https://www.patternfly.org/v4/documentation/core/demos/basicforms). If the user wanted to see the equivalent React demo version, if they try and click on the "React | HTML" toggle button to see the React version of a forms demo, they are navigated to the [About modal](https://www.patternfly.org/v4/documentation/react/components/aboutmodal) react demo I was expecting to see the [react demo of forms](ttps://www.patternfly.org/v4/documentation/react/components/form/)

The demo link isn't all that confusing in that I now see that there is only a core version of the forms demo (might want to explicitly call that out in the guidelines page you are going to core version), but what is confusing is that when I clicked on the react toggle, I was expecting to land on a page that is about forms (the components version if no demo was available).

I think the real issue is that the toggle button makes me think that it will toggle between core vs. react versions of the same component/demo/layout we are on.  Maybe its a visualization or layout issue of where the toggle button is, but never really thought of the toggle button as a top level navigational link that won't persist selections made below it.

mturley:
Fixes #812 
Fixes #965
(Duplicate issues)

This PR changes the framework switcher so that when it is rendered, it takes the current path and replaces `/core/` with `/react/` or vice versa, then checks to see if the result is a valid path. If so, the rendered link will take the user to the page matching their current component, otherwise it will still take them to the about modal page.

As long as the paths to our react and core component pages are consistently named, this should result in an intuitive framework toggle that can switch between frameworks without switching components.

4nZIOF4onR

### DIFF
--- a/packages/patternfly-4/src/components/switcher/index.js
+++ b/packages/patternfly-4/src/components/switcher/index.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { graphql, StaticQuery } from 'gatsby';
 import {
   Button,
   Bullseye,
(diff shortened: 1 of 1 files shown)


mturley:
@jschuler @redallen @dgutride can you review?

mturley:
I'm not sure why the CircleCI build is failing here, it works locally.

redallen:
It failed because there is no `.circleci/config.yml` on this branch, but there is on other branches. Rebase or merge master to get it to work. It doesn't matter anyway, though, since it's just for the deploy previews, which are broken anyway because the site is over 250MB and can't upload to surge.sh :(

packages/patternfly-4/src/components/switcher/index.js
```diff
@@ -8,21 +9,30 @@ import { Link } from 'gatsby';
 import { Location } from '@reach/router';
 import './styles.scss';
 
-const Switcher = () => (
+const trimTrailingSlash = str => str.replace(/\/$/, '');
+const verifyPath = (path, pages) =>
+  pages.edges
+    .map(edge => trimTrailingSlash(edge.node.path))
+    .find(queryPath => queryPath === trimTrailingSlash(path));
```
redallen: Won't this not work for aboutmodal vs aboutmodalbox? Maybe compare up to the shorter of the two?

mturley: I think ideally we would want to rename one or the other of those two to make them consistent.

mturley: Conveniently enough for now, those two happen to be the defaults, so toggling between them will still work despite not matching at all :)

Yes, I'm a little ashamed of this.

PR #1024, opened 2019-05-01
feat: Table of contents

jschuler:
Add table of contents to component pages


dgutride:
@jschuler  - the props anchor seems incorrect - https://1024-pr-patternfly-org-patternfly.surge.sh/v4/documentation/react/components/aboutmodal/#props doesn't work for instance

PR #1030, opened 2019-05-01
fix framework size and padding for documentation

christiemolloy:
- fixes Framework title size and spacing from top
- override padding in the page main section

Screen Shot 2019-05-01 at 5 08 43 PM


rachael-phillips:
Wow, this looks so much better!

PR #1033, opened 2019-05-02
icons: tweak open/closing

jschuler:
make the open/close behavior a little smoother


jschuler:
@redallen @dgutride should be good to merge now after build

PR #1046, opened 2019-05-02
search input hide on outside click

jschuler:


packages/patternfly-4/src/components/layout.js
```diff
@@ -55,7 +55,7 @@ class Layout extends React.Component {
         apiKey: '06941733239da4f8617d272cf2ed4d5c',
         indexName: 'patternfly',
         inputSelector: '#global-search-input',
-        debug: true // Set debug to true if you want to inspect the dropdown
+        debug: false // Set debug to true if you want to inspect the dropdown
```
redallen: Maybe use `process.env.NODE_ENV`?

jschuler: debug mode is useful when updating the search dropdown css. don't think we need to make it more than this :)

PR #1079, opened 2019-05-07
Code toggle editor styling between React and HTML

jschuler:
- HTML code toggle uses same editor now that react-live for React uses under the hood. This is to align the way they look and behave.
- Remove unused prism imports

HTML:
html

REACT:
react


dgutride:
Should hold off on merging this unless it fixes a critical bug - @rachael-phillips  - thoughts?

rachael-phillips:
Hi Dana, yes sounds good - let’s hold off

Issue #983, opened 2019-04-29
Colors in Colors Page Breaking


christiemolloy:
At a certain viewport this is happening:

<img width="408" alt="Screen Shot 2019-04-29 at 9 12 36 AM" src="https://user-images.githubusercontent.com/20118816/56898418-fd42e300-6a5e-11e9-8b3f-cb87609f15b9.png">

mindreeper2420:
@christiemolloy I believe the color blocks just need to be updated to dynamically resize on the smaller viewports - right now, they are a fixed size.

mindreeper2420:
Closing with merge of PR.

PR #1081, opened 2019-05-08
fix(colors): adjust mobile alignments for colors page

mindreeper2420:
Update the Colors page to properly render on mobile devices (smaller viewports). Set the column widths to be 12 columns on smaller screens, fixing the text/color block overlaps to occurred.

This PR also fixes the issue where the Sample Chart image appeared on smaller screens, injecting itself in the middle of the Chart Colors sample area.

fixes issue https://github.com/patternfly/patternfly-org/issues/983

Old screenshot | New screenshot
--- | ---
Screen Shot 2019-05-08 at 12 12 25 PM | Screen Shot 2019-05-08 at 12 08 26 PM
Screen Shot 2019-05-08 at 12 12 40 PM | Screen Shot 2019-05-08 at 12 08 34 PM
Screen Shot 2019-05-08 at 12 12 54 PM | Screen Shot 2019-05-08 at 12 08 48 PM

### DIFF
--- a/packages/patternfly-4/content/design-guidelines/styles/colors.mdx
+++ b/packages/patternfly-4/content/design-guidelines/styles/colors.mdx
@@ -13,36 +13,36 @@ import { Link } from 'gatsby';
   <GridItem xs={12}>
     <h3>Primary default</h3>
   </GridItem>
-  <GridItem span={12} sm={6} className="grid-modifier-first">
+  <GridItem span={12} lg={7} className="grid-modifier-first">
     <Grid>
-      <GridItem span={6} xl={3}>
+      <GridItem span={4} xl={3}>
         <div className="colors-theme--color-primary-100"></div>
       </GridItem>
-      <GridItem span={6}>
+      <GridItem span={8}>
         <h3>#0066CC</h3>
         <p>Global CSS variable</p>
         <code>--pf-global--primary-color--100</code>
       </GridItem>
     </Grid>
   </GridItem>
-  <GridItem span={12} md={6} rowSpan={2} className="grid-modifier-four">
+  <GridItem span={12} lg={7} className="grid-modifier-third">
+    <Grid>
+      <GridItem span={4} xl={3}>
+        <div className="colors-theme--color-primary-200"></div>
+      </GridItem>
+      <GridItem span={8}>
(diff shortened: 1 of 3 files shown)


christiemolloy:
I think that @gdoyle1 has some designs for how these colors respond, I believe they shouldn't wrap at all, but Gina could you attach those specs you did?

Screen Shot 2019-05-13 at 1 48 50 PM

gdoyle1:
@christiemolloy @mindreeper2420 This is what I had designed a while back https://redhat.invisionapp.com/share/9GQCHXH46QV (before Liz did the redesign) - I believe she has a copy of the mobile mocks of the colors page in her designs too but I can't find the link.

christiemolloy:
Nice job, I still think these two breakpoints are weird and awkward, Im happy to take a look at it if neeed be

Screen Shot 2019-05-13 at 2 01 59 PM
Screen Shot 2019-05-13 at 2 02 11 PM

mindreeper2420:
@christiemolloy @gdoyle1 Looking at the Colors page from Liz, I don't see a mobile mock - just the desktop one. I can update the page to not wrap the color blocks, if desired, but that would make them smaller and perhaps harder to see.

Would it be okay for this to go in the fix the current issue and then create a new issue to update the colors page to not wrap, but to resize the blocks?

mindreeper2420:
@christiemolloy what viewport size is that broken layout occurring at? I have a feeling the vertical nav is causing some issues as well, since it is still expanded, giving us less space to lay out the content.

gdoyle1:
@mindreeper2420 Yes that sounds good to me! The chart colors are shown in more detail on it's designated page so I don't think it's a big deal if they are small on this page

christiemolloy:
@mindreeper2420 yeah Its definitely happening when the left nav is expanded, forcing the page section width to be really small. I think we could leave it for now because we dont handle this case in Patternfly currently, or use some js to determine the width of that page section.

christiemolloy:
@mindreeper2420 I pushed some changes to your branch that should make the grid work better now particularly for when the nav is expanded out. Let me know what you think!

mindreeper2420:
@christiemolloy looks great! If you're good with this, then I'm good to get it into Master, so that others don't run into these issues.

Issue #1080, opened 2019-05-07
Demos are pulling in website styles


christiemolloy:
<img width="1164" alt="Screen Shot 2019-05-07 at 3 59 08 PM" src="https://user-images.githubusercontent.com/20118816/57329114-1a238a00-70e1-11e9-89d6-154b65211f8c.png">

rachael-phillips:
@mindreeper2420 I unassigned you for now until you have some time!

christiemolloy:
The demos aren't pulling in website styles anymore from as far as I can tell, but the previews are. But I think @jschuler fixed that yesterday?

christiemolloy:
Closing this issue as it is resolved

PR #1083, opened 2019-05-08
fix(demos): fix typography on demo pages

mindreeper2420:
Applies the Overpass font family from PatternFly to the demo pages.

This will be able to be removed once PatternFly utilizes Red Hat typography.

fixes issue https://github.com/patternfly/patternfly-org/issues/1080

### DIFF
--- a/packages/patternfly-4/_repos/react-core/src/demos/PageLayout/examples/PageLayoutDefaultNav.js
+++ b/packages/patternfly-4/_repos/react-core/src/demos/PageLayout/examples/PageLayoutDefaultNav.js
@@ -192,6 +192,7 @@ class PageLayoutDefaultNav extends React.Component {
           isManagedSidebar
           skipToContent={PageSkipToContent}
           breadcrumb={PageBreadcrumb}
+          className="pf4-site-demo-typography"
         >
           <PageSection variant={PageSectionVariants.light}>
             <TextContent>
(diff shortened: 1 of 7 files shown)


packages/patternfly-4/_repos/react-core/src/demos/PageLayout/examples/PageLayoutDefaultNav.js
```diff
@@ -192,6 +192,7 @@ class PageLayoutDefaultNav extends React.Component {
           isManagedSidebar
           skipToContent={PageSkipToContent}
           breadcrumb={PageBreadcrumb}
+          className="pf4-site-demo-typography"
```
jschuler: Anything under _repos will get overwritten when newer patternfly-react and core versions are pulled in. So this would only be a temporary solution. Perhaps we can add a class identifier upstream?

mindreeper2420: @jschuler this is really only necessary until PatternFly moves over to using the Red Hat typography (which will then match the rest of PatternFly.org). I don't know the timeline for that (I couldn't find an issue in the patternfly/patternfly-next repo), but it probably wouldn't make sense to put an identifier upstream.

christiemolloy: @mindreeper2420 there is a PR up in pf next currently that adds an option to use the new font family: https://github.com/patternfly/patternfly-next/pull/1813 .. I dont know how long until it will be merged but I can check with Michael Coker

christiemolloy:
@mindreeper2420 theres a decent amount of issues with these demos where they're pulling in website styling, i.e. the blue border on the nav, and the padding around the main section areas.

Screen Shot 2019-05-13 at 2 19 31 PM

mindreeper2420:
@christiemolloy I started to notice those other issues as well. I'm thinking that it'd be better to both wait for the PatternFly PR to be merged, and then update various items with PatternFly site classes (either a whole-site wrapper class, or individual ones on elements). Either that, or we could write site-specific components.

mindreeper2420:
Closing, as the work for updating PatternFly to use Red Hat fonts has begun, nullifying the need for this.

Issue #755, opened 2019-03-19
Styling for Search


rachael-phillips:
Visual design will advise on styling for search based on the newly implemented search feature on the website workspace: https://v2.patternfly.org/

And based on the Algolia documentation that powers the search: https://community.algolia.com/docsearch/styling.html

gdoyle1:
Initial InVision wireframes for search experience: https://redhat.invisionapp.com/share/W3Q3PI2RNFV

rachael-phillips:
We would have to move in the font work first to the website. This will need a design review. Wait to do a review of the Red Hat font then start adding Red Hat font on Search.

rachael-phillips:
For refining our search experience I think the interaction that Medium uses will work beautifully: https://medium.com/

And here is reference to how it could look like in our site:

I can send over the svg.
<img width="1047" alt="Screen Shot 2019-04-05 at 11 30 19 PM" src="https://user-images.githubusercontent.com/40275806/55664403-fed01100-57fa-11e9-96b4-53e5ab3ae748.png">

**1** - The static state of the search before the user interacts
**2** - Once the user has clicked on the search icon the icon moves left and help text appears as seen on medium.com

rachael-phillips:
We want to modify the look of the Algolia tag on search. How can we make the purple icon fit in more with the styling of the website?

mindreeper2420:
@rachael-phillips From what I have seen, we cannot alter the Algolia logo. That is provided in the code as an image, directly from their codebase.

rachael-phillips:
I'm closing this issue because we have completed this work

Issue #904, opened 2019-04-26
Search box text "Search" difficult to see


Rohoover:
Maybe it's my old eyes but specifically in Safari the search field text is difficult to see. It's lighter in Safari (mac)
Safari on left, Chrome on right.
<img width="555" alt="Screen Shot 2019-04-26 at 2 25 18 PM" src="https://user-images.githubusercontent.com/18266453/56828586-9be8fd00-682f-11e9-8817-2a905a50dac1.png">

rachael-phillips:
Agreed. We should build in more contrast.

mindreeper2420:
related to https://github.com/patternfly/patternfly-org/issues/755

PR #1099, opened 2019-05-14
fix(search): update styling for Search

mindreeper2420:
Update the Search styling to closer match designs.
- Improve the usability on smaller viewports by expanding the display width of the results dropdown.
- Improve padding around the results to increase clickable area (from around 25px to around 40px in height).
- Improve contrast of the Input Placeholder
- Shrink footprint of the Search input (width) so that it fits better in the navigation and does not sit right next to the PatternFly logo.

fixes Issues https://github.com/patternfly/patternfly-org/issues/755 and https://github.com/patternfly/patternfly-org/issues/904

----

**Desktop**
Screen Shot 2019-05-14 at 11 15 26 AM

**Mobile**
Screen Shot 2019-05-14 at 11 21 30 AM

----

PFSite_Search

### DIFF
--- a/packages/patternfly-4/src/styles/_navigation.scss
+++ b/packages/patternfly-4/src/styles/_navigation.scss
@@ -1,5 +1,10 @@
 .pf-c-page__header {
   background-color: #151515;
+
+  @media (max-width: 768px) {
+    grid-template-columns: 1fr auto;
+  }
+
   .pf-c-page__header-brand {
     background-color: #151515;
   }
@@ -13,6 +18,7 @@
     display: none;
     @media (max-width: 768px) {
       display: block;
+      position: relative !important;
     }
   }
 }
--- a/packages/patternfly-4/src/styles/_search.scss
+++ b/packages/patternfly-4/src/styles/_search.scss
@@ -20,7 +20,7 @@
 
 /* Category (eg. Downloads) */
 .algolia-autocomplete .algolia-docsearch-suggestion--subcategory-column {
-  padding: 5px 0 !important;
+  padding: 10px 0 !important;
   &-text{
     margin-right: var(--pf-global--spacer--sm) !important;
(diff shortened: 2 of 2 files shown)


christiemolloy:
Nice work! I think the focus state looks a little off, were you going to try outline-offset here? Even just giving the search some left padding/margin would be great.

Screen Shot 2019-05-15 at 11 59 03 AM

mindreeper2420:
@christiemolloy good call on the `outline-offset` - I've updated the PR with `outline-offset: 2px;`, which gives it a good amount of spacing between the icon and the focus ring, as well as allowing for the bottom border to be visible.

Screen Shot 2019-05-15 at 1 04 17 PM

christiemolloy:
@mindreeper2420 focus looks great. Just found the designs via this link: https://github.com/patternfly/patternfly-org/issues/755#issuecomment-480470677 , if they've been updated, you can ignore my  comment, but looks like it should just be an icon, then when the user interacts text appears, and text should be changed from Search to Search Patternfly

mindreeper2420:
@christiemolloy I'll look into updating the placeholder text to change when the input becomes active. The icon changing to an input (slide input in from right, hidden until active), might require some React component work. I tried to do that with just CSS, but the Algolia search results dropdown didn't cooperate.

mindreeper2420:
@christiemolloy @jschuler Can you review to see if we can get this into prod? Thanks!

mindreeper2420:
Also, I went through and looked at replacing the text on active input, but was not able to using just CSS. I think it'd be best to wait to make those changes until we can get some more advanced logic into the interaction.

christiemolloy:
@jschuler want to merge this when you approve

mindreeper2420:
@christiemolloy I made one final update to lengthen the search input on smaller screens after reviewing with Rachael. I promise that is the last change for this PR :-)

PR #1119, opened 2019-05-16
add google analytics support

jschuler:
There will be another PR to track outbound links


redallen:
Staging URL https://d1kr9ssmiimkrj.cloudfront.net/

PR #1126, opened 2019-05-17
content update to lists/tables - add sortable table

### LINKED ISSUE #1104: Content refinement: Add sortable data list/table usage guidelines
There has been some confusion around how sortable column headers work after reading this documentation: https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/lists-and-tables#data-tables. 

Some things to consider while creating the documentation:

If a column header is currently being sorted by, the text will turn blue and the arrow will point in the direction of the sort. Hovering over the arrow changes to a darker grey (bc the icon is actionable)

Sort behavior and guidance:

1. The default sort order for a table should support the primary use case for the application.
For example, if a table contains these two attributes:        System Name   |    Last Sync
you may want to show the most recently synced system at the top of the table (ie, it is the default sort column), because a primary use case for this table is verifying that you have successfully connected or troubleshot the system’s connection to Cloud Services.

For example, if a table contains these three attributes:     System Name   |    Last Sync   |  Severity
you may want to  show the system with the highest Severity because that is the system the user should tend to first.

2. All columns in a table do not require sort functionality. That is, you can disable the header sort function on some columns and enable it on others

fyi @KendraMar @mcarrano

gdoyle1:
Content refinements to lists/tables documentation:
1. Standard data table example
2. Sortable data table example
3. Updated "toolbar" link bc was broken

closes #1104


packages/patternfly-4/content/design-guidelines/usage-and-behavior/lists-and-tables/design.md
```diff
@@ -91,6 +91,17 @@ Do not use for less structured or variably structured data that cannot be easily
...
+### Sortable data table
+
+![sortable data table](img/sortable-data-table.png)
+
+1. **Sorted column**: when a column is being sorted by, the column header will turn blue and the sort icon will represent the direction of the sort. Subsequent clicks on the sortable column header will toggle the direction of the sort.
+2. **Hovered sort**: when a column is sortable, the sort icon will appear to the right of the column header. Upon hover, the  icon will change to a darker grey indicating that the icon is actionable.
+3. **Sortable column**: when a column is sortable, the sort icon will appear to the right of the column header in a light grey color. Sorting will not become active until the user selects the column header. This triggers the arrow to point upwards and the content to be sorted in ascending order.
+
+#### When to use
+Use a sortable data table over a standard data table. When column headers allow for sorting, it eases the ability to scan and read through the content. This is desired for data tables rather than adding sorting functionality to the toolbar.
```
mcarrano: Use a sortable data table over a standard data table when?

packages/patternfly-4/content/design-guidelines/usage-and-behavior/lists-and-tables/design.md
```diff
@@ -67,20 +67,20 @@ The expandable list adds an expansion panel to every row to reveal more details
 Use an expandable data list when you have more information than will comfortably fit inside a row, or you want to provide a way for advanced users to access information that is not applicable to all users.
 
 ## Data tables
+In general, PatternFly supports several variations of the data table component. Here are where they differentiate the most.
```
mcarrano: This feels a bit awkward. Maybe say, "PatternFly supports several variations of the data table component. These include:"  and then list them. Since you added Sortable Table as a new variant, shouldn't that be added here?

packages/patternfly-4/content/design-guidelines/usage-and-behavior/lists-and-tables/design.md
gdoyle1: @mcarrano I was basically trying to say that we should always use a sortable table rather than a standard table since it is preferred to have sorting within the table rather than only in the toolbar. Would you agree?

packages/patternfly-4/content/design-guidelines/usage-and-behavior/lists-and-tables/design.md
gdoyle1: Okay yeah I'll change that - I wasn't sure if that was the intent with this introductory sentence because I don't think it is written that way with data lists. I can list types of data lists as well though so they are consistent

gdoyle1:
@mcarrano just fixed this content based on our conversation

KendraMar:
Thanks, guys! 

Just curious if the guidance part belongs in PatternFly, too? See below:

Sort behavior and guidance:

The default sort order for a table should support the primary use case for the application.
For example, if a table contains these two attributes: System Name | Last Sync
you may want to show the most recently synced system at the top of the table (ie, it is the default sort column), because a primary use case for this table is verifying that you have successfully connected or troubleshot the system’s connection to Cloud Services.
For example, if a table contains these three attributes: System Name | Last Sync | Severity
you may want to show the system with the highest Severity because that is the system the user should tend to first.

All columns in a table do not require sort functionality. That is, you can disable the header sort function on some columns and enable it on others

gdoyle1:
@KendraMar Thanks for bringing this up! We'll add it to the usage section so users have guidance on when/how to use sortable tables

PR #1146, opened 2019-05-24
comment out broken link

### LINKED ISSUE #1130: 404 on Avatar design react doc page
https://www.patternfly.org/v4/documentation/react/components/avatar/

Link to 'Masthead' 404s

christiemolloy:
Close #1130 

Commented out the Link to the Masthead as that page doesnt exist right now. Wondering if it should link to the "page" or "nav" components as they are probably closest to the masthead? cc: @stacymcauliffe


stacymcauliffe:
LGTM - I dont think we want to link to nav. The masthead is called out as an element of the navigation system, but we dont talk about or show an example of the avatar there so it might just wind up confusing people.

PR #1220, opened 2019-06-06
Use shadow dom

jschuler:
Use shadow-dom for the examples so they do not accidentally receive css overrides from the website styling


packages/patternfly-4/scripts/getPatternFlyCss.js
```diff
@@ -0,0 +1,48 @@
+const fs = require('fs');
+const request = require('request');
+
+const packageJson = require('../package.json');
+// get the specific version of patternfly as used by the react examples
+const pfVersion = packageJson.dependencies['@patternfly/src-patternfly-next'].split('#v')[1];
```
redallen: I'm about to break this with submodules. I don't mind fixing it, just tell me what exactly it's doing.

packages/patternfly-4/src/components/ShadowDomPreview.js
```diff
@@ -0,0 +1,52 @@
...
+  className: '',
+  isReact: false,
+  isFull: false
+};
+
+const ShadowDomPreview = ({ children, className, isReact, isFull }) => {
+  // patternfly-icons assumes classes are nested under html tag, doesn't work in shadow dom so strip them out
+  const strippedHtmlIconsStyles = patternFlyIcons.replace(/html\s/g, '');
+  return (
+    <ShadowDOM>
```
redallen: Much simpler than I anticipated!

packages/patternfly-4/scripts/getPatternFlyCss.js
jschuler: actually i think i'll take another look at this. basically i need to load one set of patternfly css for the core examples, and one set for the react examples, since they may not be in sync

redallen: We fixed this making the version of @patternfly/patternfly in _repos/react-core exactly the version in _repos/core.

Issue #1188, opened 2019-06-03
Change Colors on the Website


christiemolloy:
cc: @gdoyle1 

The primary color has the right color but not the right description:

<img width="540" alt="Screen Shot 2019-06-03 at 10 27 31 AM" src="https://user-images.githubusercontent.com/20118816/58809603-3fc18780-85ea-11e9-91e6-5560bac9d9bf.png">

The secondary color for alert it also incorrect:
<img width="524" alt="Screen Shot 2019-06-03 at 10 28 13 AM" src="https://user-images.githubusercontent.com/20118816/58809664-52d45780-85ea-11e9-8b35-87fe9e2231dd.png">

rachael-phillips:
Are we ready to close this issue based on @lareplog most recent PR?

rachael-phillips:
Closing because this is complete

PR #1223, opened 2019-06-07
fix(color): update color 

lareplog:
Fixed the issues with color, second primary default color now has correct hex code and the danger icon color is now fixed (now 300 (color that matches the color in context: danger alert example))

**Fixed Hex Code For Primary Default:**
Screen Shot 2019-06-07 at 2 03 20 PM
**Fixed Danger Icon Color:**
Screen Shot 2019-06-07 at 2 03 40 PM


lareplog:
fixes #1188

christiemolloy:
Add "fixes #1188" to the original comment, you can edit it using the kebab on the right :)

PR #1245, opened 2019-06-18
fix(nav): remove margin

### LINKED ISSUE #1244: Investigate the space between the tabs
In the recent CSS Army chat @matthewcarleton bought up a really good point about the spacing in the navigation being too large. Currently it uses patternfly's spacing plus extra padding in the tabs. Could we look at redesigning these to remove the padding, this will allow us to remove the scroll buttons on a larger viewport and allow for a better user experience.

Screen Shot 2019-06-12 at 2 55 32 PM

christiemolloy:
Closes #1244 

By removing the margin between the nav items, the space between them becomes 32 which is what is default in Patternfly. This is because 16px of padding was added to the left and right of the nav-link.

Screen Shot 2019-06-18 at 2 06 40 PM
Screen Shot 2019-06-18 at 2 06 47 PM


christiemolloy:
@rachael-phillips and @mindreeper2420 could you both take a look at this PR :)

rachael-phillips:
@christiemolloy is there any way in this PR to also solve the strange problem of the black line we see at the top of the browser or will that need to be a different effort?

christiemolloy:
@rachael-phillips I removed the min-height on the header that was causing that problem

Screen Shot 2019-06-25 at 9 13 58 AM

Issue #1166, opened 2019-05-29
Toolbar examples hard to see / interpret


ajacobs21e:
https://www.patternfly.org/v4/documentation/core/components/toolbar-full

These examples are overlapping each other. It looks like I can click on things like the sort dropdown but nothing actually happens. Is this meant to be a working code example or just static images?

rachael-phillips:
Thanks for filing @ajacobs21e we've tried resolving this. @jschuler how can we solve this?

ajacobs21e:
i'd recommend either a) making the example clickable so the user can open / close each element as needed or b) spacing these out so the menus don't overlap the example below

rachael-phillips:
Agreed

jschuler:
Each full page example is supposed to open with just that example and not overlapping with the others. This is done in the HTML examples by appending `?component=something` to the URL, for example: https://www.patternfly.org/v4/documentation/core/components/toolbar-full?component=Toolbar%20simple. This behavior is buggy right now and needs to be fixed.

In terms of interactivity, HTML examples are static and not interactive. So even if the example is displayed by itself as it should be, it won't be interactive. The React examples are interactive.

rachael-phillips:
@jschuler does this have to be fixed in React upstream?

jschuler:
no we'll fix this in the website, i think this issue is relevant #926

Issue #1272, opened 2019-07-01
Page component demos are not full height


mcoker:
https://www.patternfly.org/v4/documentation/react/demos/pagelayout/

https://www.patternfly.org/v4/documentation/core/demos/page

If you go to any of the Page demos, they aren't the full height of the browser. Every extra element between `<body>` and `<div class="pf-c-page">` should have `height: 100%;` applied. Is this something we can fix?

jschuler:
They're full height now

Issue #1270, opened 2019-07-01
no background on PF4 login example page


andreasn:
The [preview](https://www.patternfly.org/v4/documentation/react/components/loginpage/) has a background image, but the [example page](https://www.patternfly.org/v4/documentation/react/components/loginpage/simpleloginpage/) doesn't.

mcoker:
Looks like this is due to the negative `z-index` on the background image component and the extra `<body>` element added to the page from the example being served in the shadow DOM. We apply a white background color to `body` in PF4, so the background image is rendering behind the extra `<body>` element. If you apply `background-color: transparent;`, to that `<body>` element, the background image works as expected.

Not sure the best approach to fix this. I wonder if the negative `z-index` on that component is necessary. It will be a breaking change if we want to try and remove that though.

@jschuler Is there any way for the shadow DOM to not render the `<body>` element? And if so, what about the `<html>` and `<head>`?

jschuler:
Thanks for investigating this @mcoker ! I was wondering what was happening here. I need the html tag so that font icons properly work. I think the best for now would be to add the transparent background color to the body tag. Does that seem ok?

mcoker:
@jschuler are you able to test a branch of mine that has a fix? I think I can fix the FA issue in core. https://github.com/mcoker/patternfly-next/tree/issue-pf-org-1270

jschuler:
@mcoker i just tested and your changes to fix FA issue works!

mcoker:
@jschuler thanks! submitted a PR in core with this fix. https://github.com/patternfly/patternfly-next/pull/2018

jschuler:
Thanks for reporting @andreasn , this is fixed now

PR #1273, opened 2019-07-02
fix core fullscreen urls

jschuler:
Fixes: #1166, #1270, #1272


jschuler:
I'll reopen this when it's ready

PR #1281, opened 2019-07-11
update pf4 site deps to match submodules, and add pf3 deps to its noh…

jschuler:
…oist list


packages/patternfly-3/package.json
```diff
@@ -71,7 +71,23 @@
...
+      "bootstrap-combobox",
+      "bootstrap-datepicker",
+      "bootstrap-select",
+      "clipboard",
+      "datatables",
+      "font-awesome",
+      "google-code-prettify",
+      "jquery-match-height",
+      "jquery.scrollto",
+      "patternfly",
```
redallen: `patternfly` and `fontawesome` are here twice, but that's ok.

jschuler: oh thanks! i copied the lines from dependencies and overlooked that some were already in here. Maybe the next PR can address this?

redallen: Yep!

PR #1289, opened 2019-07-15
update PF3 repos on build

jschuler:
updates just the PF3 submodules since we want manual control over the PF4 submodules

### DIFF
--- a/packages/patternfly-4/package.json
+++ b/packages/patternfly-4/package.json
@@ -5,12 +5,13 @@
   "author": "Red Hat",
   "private": true,
   "scripts": {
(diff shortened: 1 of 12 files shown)


packages/patternfly-4/package.json
```diff
@@ -5,12 +5,13 @@
   "author": "Red Hat",
   "private": true,
   "scripts": {
-    "build": "yarn compile-examples-css && node --max_old_space_size=4096 ./node_modules/.bin/gatsby build --prefix-paths",
```
redallen: Why shouldn't this be a part of the build step?

jschuler: it will update an existing file that is already checked in. How does the build handle new/updated uncommited changes?

PR #1377, opened 2019-07-29
removed 4 from body copy

stacymcauliffe:
removed references to PF4 from get started content and home screen


stacymcauliffe:
@rachael-phillips - removed instances of 4 from the contribute section. I think updating that doc further should probably be a separate issue.

Issue #1367, opened 2019-07-24
Contact us page


rachael-phillips:
Here is the tentative design for the contact us page:

![image](https://user-images.githubusercontent.com/40275806/62154807-845d5c80-b2d5-11e9-8f09-aef47d0adcd1.png)

@mcarrano , @stacymcauliffe , and @LHinson do you have any other info you would like to add to this page?

mcarrano:
Actually, yes @rachael-phillips .  When I was talking with @mmenestr yesterday, we were trying to decide on a good placement for information about key contacts on the project. Right now, it's being placed at the bottom of the Get Started page for Designers (see draft here https://docs.google.com/document/d/1Heig3CdlQLpNCFcCGs9ySq3vo-1dxnEj59f4_0qYKKk/edit) But this does not seem like the best place for it. Would make sense here if we decide to include named contacts.

rachael-phillips:
Hi @mcarrano I have reservations about listing named contacts because our main channels of communication should suffice for answering questions.

mcarrano:
We should discuss further @rachael-phillips . I could go either way with this, but it came out of a convo between @LHinson and I about how to make PatternFly feel less impersonal. Maybe there are other ways to do this.

rachael-phillips:
I talked about this today with @mcarrano and @stacymcauliffe. We decided that the most visible and understandable way to make the Contact Us visible on the website is to bring it up to the top nav. 

![image](https://user-images.githubusercontent.com/40275806/62162279-e756ef80-b2e5-11e9-9e4c-46f88318741f.png)

We will also need to update the footer with a Slack link.

LHinson:
I like this solution @rachael-phillips

rachael-phillips:
We will call this page "Get in touch" instead of "Contact us"

PR #1401, opened 2019-08-07
Get in touch page

jschuler:
Closes: #1367 

- Adds the new `Get in touch` page
- Added linting
- Cleaned up a few linting errors and removed some unused files

Screen Shot 2019-08-07 at 3 57 49 PM


rachael-phillips:
Can we have the light grey expand to fit the full window - it would be responsive, so the user only sees the footer when they scroll down

PR #1432, opened 2019-08-28
Always compile example CSS

### LINKED ISSUE #1430: Shadow dom css example script is not automatically updated
yarn update in the patternfly-4 package.json should be automatically run whenever we update the submodules (possibly by the build server).  Right now, it's a manual step which means it's going to be missed almost every time.

redallen:
Always recompile example CSS. Closes #1430 .


jschuler:
No good reason I can think of, thanks @redallen

PR #1437, opened 2019-08-30
Update issue templates

rachael-phillips:
Adding this template to our org repo.

### DIFF
--- a/.github/ISSUE_TEMPLATE/patternfly-issue-template.md
+++ b/.github/ISSUE_TEMPLATE/patternfly-issue-template.md
@@ -1,0 +1,17 @@
+---
+name: PatternFly issue template
+about: Please populate the PatternFly issue template so we can learn more about your
+  needs.
+title: ''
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/patternfly-issue-template.md
```diff
@@ -0,0 +1,17 @@
+---
+name: PatternFly issue template
+about: Please populate the PatternFly issue template so we can learn more about your
+  needs.
+title: ''
+labels: ''
+assignees: ''
+
+---
```
dlabrecq: For patternfly-react, we don't have such a table. Don't believe we want assignees, labels, etc. here.

redallen:
This template is better suited for Next/React rather than infrastructure and design. It's also named incorrectly and doesn't need header content. If we want to make an issue template we can talk about what issues people are currently posting in this repo.

PR #1441, opened 2019-09-04
Experimental Features (Styling)

christiemolloy:
Adding components and updated styling for the experimental features, however have run out of time before tomorrow to add functionality

@jschuler or @redallen are you able to quickly help me with the functionality here: 
1) Only turn alerts on when the feature is deprecated
2) Only have the experimental label for the experimental components.

Screen Shot 2019-09-04 at 1 40 06 PM

@rachael-phillips can you check design ^


redallen:
@christiemolloy We need some metadata for each example on what experimental "state" it's in. For React, I want to add it in https://github.com/patternfly/patternfly-react/pull/2809 .

rachael-phillips:
State 1 - Default
1. Blue info inl-line alert should read "This is an experimental feature in the early stages of testing. It's not intended for production use."

State 2 - Only if we have decided we will deprecate the component or need warning otherwise
2. Yellow inline alert with the deprecation text on the wireframe

christiemolloy:
Closing this in favor of https://github.com/patternfly/patternfly-org/pull/1448

PR #1464, opened 2019-09-13
switch the website root to v4

jschuler:


dgutride:
one question - why can't we change the root index.html page to basically be a full copy of the index.html page in /v4/ so we don't have to do the redirect magic - just a bit worried about seo and how this impacts evderything

jschuler:
@dgutride we can do that, but seems that gatsby still does a redirect to /v4/ internally somehow so the URL bar changes immediately to /v4/

PR #1616, opened 2020-01-14
fix(docs): added max-width to text & images

### LINKED ISSUE #1133: Add max-width to markdown content
Add a max-width to prevent content from pages authored in markdown from stretching across the entire width of a page. 

Pages impacted:
All content in getting started section
Design guidelines > Usage and behavior
Design guidelines > Content
All descriptions at the top of the component docs pages

evwilkin:
resolves #1133

### DIFF
--- a/packages/patternfly-3/Gemfile.lock
+++ b/packages/patternfly-3/Gemfile.lock
@@ -47,6 +47,3 @@ DEPENDENCIES
   jekyll-sitemap
   json
   rack
-
-BUNDLED WITH
-   1.16.2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,12 +177,6 @@ jobs:
         keys:
         - *v3_ruby_deps_cache_key
     - run:
-        name: Configure Bundler
-        command: |
-          echo 'export BUNDLER_VERSION=$(cat Gemfile.lock | tail -1 | tr -d " ")' >> $BASH_ENV
-          source $BASH_ENV
-          gem install bundler
-    - run:
         name: Install v3 deps
         command: |
           cd packages/patternfly-3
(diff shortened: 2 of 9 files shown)


packages/patternfly-3/Gemfile.lock
```diff
@@ -47,6 +47,3 @@ DEPENDENCIES
   jekyll-sitemap
   json
   rack
-
-BUNDLED WITH
```
redallen: Removing the BUNDLED WITH worked! Nice!!!

evwilkin:
> Core preview: https://patternfly-org-pr-1616-next.surge.sh
> React preview: https://patternfly-org-pr-1616-react.surge.sh
> Org preview: https://patternfly-org-pr-1616-org.surge.sh

@rachael-phillips @gdoyle1 looks like this got merged already, but visual review of the changes would still be good!

gdoyle1:
@evwilkin I checked it out yesterday and it looked good!

rachael-phillips:
@gdoyle1 @mcarrano @evwilkin  I really like the width changes - a great improvement! The only thing that looks a bit strange to me is that the tables are half the length of our examples on the component pages. As a user, I'm not sure I would make the connection as to why the styling is different. Here's a view from where the example differs from the table: 

Screen Shot 2020-01-15 at 9 15 51 AM

What do you think about the idea of all of our tables still being full width? Curious to hear your thoughts!

gdoyle1:
> What do you think about the idea of all of our tables still being full width? Curious to hear your thoughts!

@rachael-phillips @evwilkin @mcarrano I agree that on the components pages, we should expand our table width to match the component width! I don't think we should do that EVERYWHERE though - for example, on the typography page, the width it is at now looks good to me.
Screen Shot 2020-01-15 at 9 46 36 AM

evwilkin:
I agree - picking up this change in #1622

rachael-phillips:
Thanks @gdoyle1 and @evwilkin !

PR #1620, opened 2020-01-15
empty state fixes

### LINKED ISSUE #1612: Fix some issues in empty state documentation
Issues found [here](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/empty-state/);

**General edits:**
- “No data to show” link and “addition or creation” link at the top of page do not work
- comma placement off in sentence "Below is an example of a large empty state inside a full-page outlining, all the elements inside an empty state."
- Title link on bullet 2 does not work. Maybe change from “Title” being capitalized to “title “ non capitalized
- Body link on bullet 3 does not work
- Primary button link on bullet 4 does not work
- Secondary buttons (optional) on bullet 5. link does not work
- add a space between table and "You should always try to use an icon in your empty state, unless there is an empty state where an icon does not match any of the situations mentioned above, or if an icon does not fit into the layout."
- Add return for bullet 1. Filtering or search situation for situations where stand-alone link button should be used…
- Add dash between in-table, first-use, in-page, full-age, calls-to-action everywhere on page
- "User not entitled or does not have access” link doesn’t work

**Image size reductions needed:**
- All do..’s and don’t,,,s under title and body section
- All pics under primary and secondary buttons

mmenestr:
Fixed some formatting and link issues in empty state documentation.
[Fixes #1612 ]


mmenestr:
@mcarrano that's weird that the icons look so big here, when they don't look that big if you look at them here https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/empty-state#icons.

and sure! I can change the column order :)

evwilkin:
@mcarrano it looks like the icon size is related to #1133 where we resized the max-width for images in the docs, and these icons load as images - I'll see if I can resolve it in my new PR #1616

mcarrano:
Thanks @evwilkin See the updates you made in #1616

mmenestr:
@mcarrano been thinking about the column reordering, and i feel like as a designer, it would be more useful for me to have the column order as is. My thinking is that as a designer, the main use I have for this table, is identifying what icon matches to what type of scenario. Let me know your thoughts on that! 

Also, for the icons, can I assume that's all set based on what @evwilkin did?

mcarrano:
That's fine @mmenestr . I don't feel strongly about it. So can we assume this is ready to merge then @evwilkin ?

evwilkin:
@mcarrano @mmenestr the icons issue should be resolved with this PR that was just merged:
https://github.com/patternfly/patternfly-org/pull/1623

PR #1647, opened 2020-01-27
feature(docs): Property names are no longer wrapping. this looks allo…

ddonahue007:
Before:
Screen Shot 2020-01-27 at 2 56 18 PM

After:
Screen Shot 2020-01-27 at 2 54 05 PM

### DIFF
--- a/packages/patternfly-4/gatsby-theme-patternfly-org/components/propsTable/propsTable.js
+++ b/packages/patternfly-4/gatsby-theme-patternfly-org/components/propsTable/propsTable.js
@@ -34,7 +34,9 @@ export const PropsTable = props => {
       caption={props.caption}
       rows={props.rows.map(row => ({
         cells: [
-          row.name,
+          <div className='pf-m-fit-content'>
+            {row.name}
+          </div>,
           renderType(row),
           <div>
             {row.required ? 'Yes' : 'No'}


packages/patternfly-4/gatsby-theme-patternfly-org/components/propsTable/propsTable.js
```diff
@@ -34,7 +34,9 @@ export const PropsTable = props => {
       caption={props.caption}
       rows={props.rows.map(row => ({
         cells: [
-          row.name,
+          <div className='pf-m-fit-content'>
```
redallen: Funny enough, I think we had done this before at some point. Don't remember why we reverted it.

redallen: Also, I'm surprised that single quotes in `className='pf-m-fit-content'` works! I'm used to HTML requiring double quotes. I think this is a "feature" of JSX, but until we lint this code I don't mind :)

ddonahue007: Yeah I originally . had double quotes, but then saw the rest of the file was using single so I followed the pattern...

PR #1684, opened 2020-02-10
downgrade circleci

redallen:


redallen:
The Gatsby cloud check can be ignored. I'm playing around with using it for Gatsby builds since they take so long on 2 core / 7.5GB of RAM instances.

PR #1698, opened 2020-02-21
chore(docs): replaced homepage principles section with new design

### LINKED ISSUE #864: Slightly fuzzy image on home page (resolution)
Using: Firefox on Mac, Retina Screen

The image is a little low res looking on a Retina screen
Xnip2019-04-26_14-10-58

evwilkin:
Resolves #864 - this PR replaces the Dashboard image on patternfly.org homepage with new icon design.


evwilkin:
Thanks @gdoyle1 I've updated this PR to address your first 3 points, and will stay tuned on number 4!

gdoyle1:
@evwilkin I have an update on #4 and two additional requests...
Update on #4) Can we match the width of this section to the one above it? So we'll have the icons left-aligned with where the text in the section above it begins. The spacing can still remain the same between the icon and text but the text width would extend to match the right-end of the laptop image above. (This _may_ need to be reduced depending on how large it looks).
2) Can we center the "PatternFly principles" title
3) Can we change the button styles so they match the ones in the section above (link with arrow) but they will be **white** instead.

Updated marvel file: https://marvelapp.com/5hed282

evwilkin:
Thanks @gdoyle1 - I’ve pushed the changes you request, the Org preview link should now be updated. LMK what you think!

gdoyle1:
@evwilkin This looks way better! Thank you for making those changes. I found one strange break point thing (around 780px) where "PatternFly principles" wraps to two lines. Any way we can make it stay on one line? Here's the screenshot:
Screen Shot 2020-02-25 at 10 03 05 AM

evwilkin:
@gdoyle1 great catch, I've updated the preview so that the title doesn't wrap.

PR #1699, opened 2020-02-22
fix(docs): set min-height on sidebar

### LINKED ISSUE #1602: Fix page height for HTML "Training" section
Training page under "HTML" dropdown does not have min height, so footer is currently pulled up to the middle of the page.

evwilkin:
Fixes #1602 - This PR sets min-height on sidebar so page content always fills screen vertically, fixes issue of footer floating up the page on HTML training section.


mcoker:
That's one way to fix it, but `500px` looks like a rough estimation of the leftover space from the sections of the page that aren't the nav and main section. Unless that `min-height` is perfect, you'll get an unnecessary scrollbar with content overflowing the page when it isn't necessary when the value is too large (currently), or end up with the same issue the PR resolves if the value is too low.

Looks like the root cause is from a few issues:

* The parent element to `.pf-c-page` needs `height: 100%` to consume the available height of the viewport.
* We're re-using `.pf-c-page__main-section` outside of `.pf-c-page` and that's causing issues.
* `.pf-c-page` is designed to be the main wrapper for the page content and it has a height property that is causing problems.

If we can add a class (`.ws-page-wrapper`) to the `<div>` that wraps the page content and banner/footer elements, we could fix up the styling with these styles

```
/* serves as the main page container */
.ws-page-wrapper {
  height: 100%;
  display: flex;
  flex-direction: column;
}

/* disables the default flex-grow: 1 for the last .pf-c-page__main-section in a block */
.pf-c-page__main-section.ws-org-pfsite-l-footer-dark {
  flex-grow: 0;
}

/* keeps the banner from shrinking */
#ws-pf3-banner {
  flex-shrink: 0;
}

/* tells .pf-c-page to consume the available space left over from the banners/footers */
.ws-page-wrapper > .pf-c-page {
  flex: 1 1;
  height: auto;
}
```

evwilkin:
Thanks for the feedback @mcoker !  I was able to add some specific classes to the elements we're targeting and updated the rules you created to match these new classes.

packages/patternfly-4/gatsby-theme-patternfly-org/components/footer/footer.css
```diff
@@ -99,6 +99,8 @@
 .ws-org-pfsite-l-footer-dark {
   padding: 1rem !important;
   background: #151515 !important;
+  /* disables the default flex-grow: 1 for the last .pf-c-page__main-section in a block */
+  flex-grow: 0;
```
mcoker: We need to increase the specificity of the selector that applies this style, since the original style's specificity is 20 (`.pf-c-page__main-section:last-of-type`). In [my other comment](https://github.com/patternfly/patternfly-org/pull/1699#issuecomment-590454683) I recommended `.pf-c-page__main-section.ws-org-pfsite-l-footer-dark`, but you could also use `.ws-org-pfsite-l-footer-dark:last-child` or `.ws-side-nav-layout .ws-org-pfsite-l-footer-dark`. And I noticed you use `!important` in these styles, that would also work here, though I generally try to avoid `!important` but that's just me :)

This is the way it looks now

Screen Shot 2020-02-27 at 2 55 23 PM

After the change, it should look like this (the main section fills all of the available space)

Screen Shot 2020-02-27 at 2 55 08 PM

evwilkin: Updated this - I had to combine your original selector `.pf-c-page__main-section.ws-org-pfsite-l-footer-dark` and add the parent selector `.ws-side-nav-layout` to override the existing flex styling.

mcoker:
@evwilkin sorry for the delay! This fell off my radar. 

This looks great, other than on this page https://patternfly-org-pr-1699-react.surge.sh/, the first section in  `<main>` is a `<div>` with a bunch of custom styles (display: flex; height: 80vh;). Looks like you can just remove that. The `height` is causing the page to overflow vertically, and removing the `<div>` fixes it.

On https://patternfly-org-pr-1699-next.surge.sh/ there is an inline style in the main section content that sets `height:70vh`. That should be removed so the page height fits the window properly. Also just an aside, you can add `class="pf-c-content"` to the div that wraps the text on that page, then remove the `.ws-p` classes from the `<p>` elements, and our content component's spacing will space the content there. You could also wrap the first sentence on that page in a `<p>` tag.

PR #1772, opened 2020-04-07
Update colors page

### LINKED ISSUE #1517: Colors page redesign
There are color variables that are in the Sketch document (e.g. pf-orange-300) that are not available in PF4. 

PF3 had these documented on the [color palette page](https://www.patternfly.org/v3/styles/color-palette/)
PF4 has only select colors (the most common & tied to functions) on the [color palette page](https://www.patternfly.org/v4/design-guidelines/styles/colors)

The [PF4 global variable list](https://www.patternfly.org/v4/documentation/react/overview/css-variables/) also does not list the color variables.

They appear to be in PF4 based on [this](https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/sass-utilities/colors.scss) ? But are not in any clear documentation.

The request is to have the color variables documented in the Global CSS variables list.

redallen:
Closes #1517

Preview: https://patternfly-org-pr-1772-v4.surge.sh/v4/design-guidelines/styles/colors

Couple of differences from design:
## Color palettes
- Cursor moved to fit in the circle because the drop shadows on hovers appear as circles rather than squares.
- Color families don't open when linking from a palette. It does scroll you to the right color in the family, though.
## Status and state colors
- ~Status and state colors section is a little cramped compared to the design, which I don't think accounted for the space a side nav takes. Do we want to keep the 3 column layout or make the font smaller @gdoyle1 ?~
## Color families
- ~The shadows family has no shadows currently. It's included on one page in the design but not another... do we want it in or not @gdoyle1 ? I'm fixing it to look correct now just in case...~


redallen:
1. :heavy_check_mark: 
2. Difficult since our `react-token`s values are wrong. I'll try to use @jschuler 's new `patternfly_variables.js`
3. :heavy_check_mark: 
4. :heavy_check_mark: 
5. :heavy_check_mark: 
6. :heavy_check_mark: 
7. :heavy_check_mark:

PR #1845, opened 2020-05-13
Fixes "kebab" spelling in all occurrences

jenny-s51:
Closes [#3914 in patternfly-react](https://github.com/patternfly/patternfly-react/issues/3914).


mcarrano:
Thanks for finding and making these updates @jenny-s51 !  If you are doing this, the text that is referenced by @srambach in the issue is in this file: https://github.com/patternfly/patternfly-org/blob/master/packages/v4/src/content/design-snippets/overflow-menu.md  Can you also update that as part of this PR?

jenny-s51:
@mcarrano Certainly! I've included that in this PR (see here)[https://github.com/patternfly/patternfly-org/pull/1845/files#diff-e1b6c7df4911f2bfde698225ac07a292R5]

PR #1849, opened 2020-05-18
Add Notification drawer design guidelines

### LINKED ISSUE #1224: Notification Drawer - Design Doc
Introduce design documentation for the notification drawer.

Related issue: #35

mcarrano:
Adds design guidelines for the Notification Drawer

Closes #1224


mcarrano:
@mceledonia @maryshak1996 @mmenestr This PR adds design documentation for the Notification Drawer which has now been added to the PatternFly release. This is based on the designs done last year that were implemented by OpenShift. The badge updates will be added in a future release so are not included here.

Please review for both content and form. Thanks!

mcarrano:
@mmenestr Thanks for the detailed review. I've gone in and fixed your issues. Please review this again and approve if good so this can be merged.

Regarding this question...

> Does the notification drawer work independently from toast alerts? I had originally imaged that anything that popped up as a toast alert would then be stored into the notification center. Kind of how facebook does it - if you are on Facebook when you get a notification, it pops up, and then gets stored in their notification "center". If the notification happens while you are not on the page, then it just gets stored in the notification center right away. Wondering if we want to mention anything about that possible interaction or if it's up to each app whether they still want to use toast alerts and then store them, or if they want to X out toast alerts and only use the notification center. 

Generally what you are saying makes sense, but I've written this to be as general as possible because I've found that applications have a range of policies about how they manage notifications so the drawer needs to be flexible. It would be nice to have more consistency, but this to difficult to drive from a PatternFly perspective.

mcarrano:
@redallen This should be ready to be merged. Please take a look and merge it if everything looks good to you. Thanks.

PR #1853, opened 2020-05-22
Fix/1850 typography size

### LINKED ISSUE #1850: Font size issue on typography page
https://www.patternfly.org/v4/design-guidelines/styles/typography

Looks like there is a problem assigning the `font-size` var, it's being set as `font-size: var(undefined);`

Screen Shot 2020-05-19 at 10 38 01 AM

### LINKED ISSUE #1877: Font missing from release notes page
See screenshot.
The font in the release links is serif, everything else is sans-serif. I'm thinking this wasn't intentional?

image

### LINKED ISSUE #1880: Font missing from Get in touch page
Screen Shot 2020-06-11 at 9 56 32 AM

See screenshot. Header is in a serif font.

evwilkin:
Closes #1850 
Closes #1877 
Closes #1880 

This PR fixes the font size being passed to `PfStyledText` component by changing the prop used from fontSize to variableName.
This PR also removes usage of the RedHatDisplay font throughout the site as follows:
- Red Hat footer text/links
- Release notes card headers
- Typography page
  - Removed "opt in" alert instructing user to wrap application with `.pf-m-redhat-font`
  - Updated *Our font family* copy to remove "Red Hat Display"
  - Removed "Font family" lines from *Usage guidelines* style tables
- Get in touch page header

Note: new issue https://github.com/patternfly/patternfly-org/issues/1883 created to review & updated Red Hat font page.


mcoker:
Thanks @evwilkin! If this is part of the v4 release, we no longer ship "RedHatDisplay", so the font isn't rendering properly in this PR.

We only use a single `font-family` now ("RedHatText"), so I'm not even sure if we need to call out "font family" in these examples? 

But in the short term, if we can just update the inline styles to "RedHatText", and I'm assuming we'll want to update the documentation under each example to reflect that, too.

-----

Longer term, we have vars for all of the inline styles that it seems like we should use instead of static values? Specifically `font-family`, `font-weight` and `line-height`. Also looks like this whole section should be reviewed post v4 release if it isn't already in the works.

evwilkin:
@mcoker thanks for the feedback - I've removed the RedHatDisplay font (both on this page and throughout the site) and mostly everything automatically reverted to RedHatText.

I did remove the "font family" rows from the [Usage guidelines](https://www.patternfly.org/v4/design-guidelines/styles/typography#usage-guidelines) section & I've updated the description with the rest of the changes made, but LMK if you think that Usage guidelines section still needs to be tweaked to add additional variable names or any other info.

PR #1878, opened 2020-06-09
removed 600px max-width on homepage headline

### LINKED ISSUE #1875: fix text break on homepage
I might be alone here but the max-width that is causing this orphan text in the header could be fixed so it's all on one line.
Screen Shot 2020-06-09 at 4 11 03 PM

evwilkin:
Closes #1875

### DIFF
--- a/packages/v4/src/pages/index.js
+++ b/packages/v4/src/pages/index.js
@@ -31,17 +31,17 @@ const IndexPage = ({ location }) => {
   <SideNavLayout location={location} hideSideNav showGdprBanner={false} showBanner={true}>
     <PageSection
       variant={PageSectionVariants.light}
-      className="pf4-c-background-image pf-u-display-flex pf-u-justify-content-center pf-u-align-items-center ws-homepage-main-section"
+      className="pf4-c-background-image ws-homepage-main-section"
     >
       <Grid>
         <GridItem sm={12} md={8} mdOffset={2} lg={6} lgOffset={3} className="pf-u-py-2xl pf-u-text-align-center">
           <TextContent>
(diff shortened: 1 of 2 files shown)


evwilkin:
@matthewcarleton @rachael-phillips thanks for the additional feedback - this has been updated so that text does not wrap onto 2 lines on larger screens, and got some help avoiding a single word ending up on its own line when text does need to wrap.

packages/v4/src/pages/index.js
```diff
@@ -31,17 +31,17 @@ const IndexPage = ({ location }) => {
...
         <GridItem sm={12} md={8} mdOffset={2} lg={6} lgOffset={3} className="pf-u-py-2xl pf-u-text-align-center">
           <TextContent>
             <img src={orb} alt="PatternFly logo" className="fadeInDown animated fadeInOne" />
             <Title headingLevel="h1" size="4xl" className="pf-m-white pf4-site-c-hero fadeIn animated fadeInTwo">
-              Build scalable experiences in the open.
+              Build scalable experiences in the&nbsp;open.
             </Title>
             <Title size="xl" headingLevel="h2" className="pf-m-white pf-u-mb-md pf-u-mb-3xl-on-md fadeInUp animated fadeInThree">
-              PatternFly is an open source design system built to drive consistency and unify teams.
+              PatternFly is an open source design system built to drive consistency and&nbsp;unify&nbsp;teams.
```
matthewcarleton, line 44: curious why you wouldn't use CSS to fix this? It kind of feels like styling inline and doesn't seem as bullet proof.

PR #1881, opened 2020-06-11
Add alert to Training because Katacoda is down

### LINKED ISSUE #1874: Add Alert to Training Page to Communicate status of Katacoda
The PatternFly training on Katacoda is currently broken. We should communicate this to our PatternFly community by adding an Alert component at the top of the Training page. It would be great if we could turn this on/off everytime we know that Katacoda isn't working.

christiemolloy:
closes #1874 

Screen Shot 2020-06-11 at 10 10 01 AM

@redallen there used to be navigation in the top right corner of the tutorial page, I think the code is still there but im not sure why its not displaying any more 

Screen Shot 2020-06-11 at 10 13 03 AM


redallen:
@christiemolloy Recent breaking changes might have broken that in the layout. Open an issue with what you'd expect/what was there before and we can fix it!

PR #1903, opened 2020-06-22
don't show nav toggle when hiding side bar

### LINKED ISSUE #1900: Page header - bug near the search box
There is a small black box below the search box that shouldn't be there. It only happens at small breakpoint range. I think the problem is with this:

https://github.com/patternfly/patternfly-org/blob/84c0d0885cf3bb78f1b88bf5619b111445e4b03d/packages/gatsby-theme-patternfly-org/layouts/sideNavLayout/sideNavLayout.css#L20-L41

Screen Shot 2020-06-22 at 10 23 09 AM

redallen:
Closes issue reported by @mcoker


evwilkin:
Looks good, in combination with #1904

Issue #1899, opened 2020-06-22
Typography guidelines image updates


mmenestr:
Update these two images on[ this page](https://www.patternfly.org/v4/design-guidelines/styles/typography) to use Red Hat Text:
Image remains the same otherwise, it's just a font change!

![image](https://user-images.githubusercontent.com/51165119/85294087-36f95900-b46c-11ea-9f7e-6f451ab5215b.png)

PR #1906, opened 2020-06-23
Updated Typography Images

abchen-redhat:
Updated typography guidelines image
Resolves [#1899](https://github.com/patternfly/patternfly-org/issues/1899)

@gdoyle1


evwilkin:
@mcarrano yes this page has been updated in Org to use the RedHatDisplay font, but the latest changes haven't been pulled into Org which include that font file.  You should be good to proceed with this as is, and we'll see that font rendering fixed after testing wraps when Org is bumped.

Issue #1887, opened 2020-06-16
Update Modal design guidelines to include additional content guidance


mcarrano:
Update Modal guidelines here: https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/modal

@abigaeljamie please add details

abigaeljamie:
Thanks, @mcarrano ! Here's what I'd like to add to the "Content" section: 

__________________________________
Confirmation dialogs should be clear, concise, and direct. Give the user the important information first—additional detail like “Are you sure?” clutters your message and can be distracting.  

Depending on your design template, follow this format for confirmation dialogs: 
**Headline**: A headline is usually phrased as a question. Include important keywords (like “permanent”) in the headline.
**Body text**: Body text gives information about the action’s consequence. 
**Buttons**: Buttons allow a user to answer the headline question. These can simply be “yes” or “no,” but adding some context increases clarity—especially when skimming.     

Before: 
**Are you sure you want to delete this record?** 
If you delete this record, then you will permanently lose all of its information. Do you wish to proceed?  

Cancel | Delete 


After: 
**Permanently delete record?** 
Record name, location, and notes will be gone forever.

No, keep | Yes, delete 

_________________________

The "Before" and "After" can be side-by-side in a chart, like we'll have in the PF content style guide.

Issue #1787, opened 2020-04-14
Update Modal design guidelines to reflect revised spacing


mcarrano:
Update Modal dialog design guidelines located [here](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/modal) to:
1. Include information on the "compact" modal spacing vs. the "large" modal spacing
<img width="592" alt="Screen Shot 2020-06-18 at 10 30 02 AM" src="https://user-images.githubusercontent.com/30870144/85033212-b2969580-b14e-11ea-8bfb-e398be460f98.png">
See the [PatternFly template file](https://www.sketch.com/s/729c2eee-e8b6-4fcd-8a79-f6faa8c30f89) for these artboards
2. Update all modal images to use the "compact" modal spacing

PR #1913, opened 2020-06-25
Modal Dialog Page Updates

abchen-redhat:
Updates made:

Content
• Added compact and large spacing documentation
• Added Get confirmation modal documentation

Images
• Updated all images to use the compact spacing
• Updated all images to use PF4 library (fonts, etc)

Icon use in modal dialogs
• Removed error icon row
• Updated all icons to colored versions

This solves the issues [#1787](https://github.com/patternfly/patternfly-org/issues/1787) and [#1887](https://github.com/patternfly/patternfly-org/issues/1887)

@gdoyle1


packages/v4/src/content/design-guidelines/usage-and-behavior/modal/modal.md
```diff
@@ -17,6 +17,17 @@ Use modal dialogs load and display important information without navigating away
...
+### Spacing
+#### Compact
+<img src="./img/compact.png" alt="Compact" width="756"/> 
+
+The compact modal dialog uses a spacing of 24px. Use this for smaller modals that don’t contain much content.
+
+#### Large
+<img src="./img/large.png" alt="Large" width="756"/>
+
+The compact modal dialog uses a spacing of 32px. Use this for larger modals that contain a lot of content (code previews, forms, tables...).
```
mcarrano, line 29: I think you mean the Large modal here

packages/v4/src/content/design-guidelines/usage-and-behavior/modal/modal.md
```diff
@@ -32,25 +43,33 @@ Modal dialogs interrupt user task flow and should be avoided for any information
...
-![Modal](./img/confirm-non-destructive-action.png)
+<img src="./img/confirm-non-destructive-action.png" alt="Confirm a non-destructive action" width="1100"/>
+
 Provide buttons on confirmation dialogs that make it simple for users to understand and confirm their choices. When choosing button labels, use descriptive verbs that clearly indicate the consequence of a selection, like Delete, Save, or Create. Avoid vague labels like OK or Done.
 
 Ambiguous title and button labels make it more difficult for the user to understand their options and make a decision. Descriptive title and button labels make it easy for the user to scan the content and make a decision quickly.
 
 ##### Confirm a destructive action
-![Modal](./img/confirm-destructive-action.png)
+<img src="./img/confirm-destructive-action.png" alt="Confirm a destructive action" width="1100"/>
```
mcarrano, line 63: Hmmm... While I agree about the messaging, I actually like the button labeling of the first example better -- just Delete and Cancel and clear and to the point.

packages/v4/src/content/design-guidelines/usage-and-behavior/modal/modal.md
```diff
@@ -32,25 +43,33 @@ Modal dialogs interrupt user task flow and should be avoided for any information
...
+
 If an action is difficult or impossible to undo, a warning icon can be incorporated into the modal to add emphasis. If an action will seriously impact the system and cannot be undone, use a destructive style button. Otherwise, use a primary button.
 
 Sometimes it is necessary to require an additional confirmation with a destructive action. An example using a form field can be seen below. In order for the destructive action to become actionable, the user needs to type in a word or phrase to make the button active.
-![Modal](./img/2-step-confirm-destructive-action.png)
+<img src="./img/2-step-confirm-destructive-action.png" alt="2 step confirm a destructive action" width="1100"/>
```
mcarrano, line 68: Maybe this example can be improved. This is a poor confirmation dialog because it never tells me the name of the thing I am going to delete. Sometime I've seen this done where the user needs to somehow type or select the name before Delete is enabled.

packages/v4/src/content/design-guidelines/usage-and-behavior/modal/modal.md
```diff
@@ -32,25 +43,33 @@ Modal dialogs interrupt user task flow and should be avoided for any information
 
 ### Types of modals
 #### Get confirmation
```
mmenestr, line 45: Would change the header here to match the link above: "Validate user decisions"

packages/v4/src/content/design-guidelines/usage-and-behavior/modal/modal.md
mmenestr: Yeah, I think the buttons can remain the same on both (although I think both are fine)! I think a point that can be made here is that the "question" you are posing in the modal should be easily answered via the buttons. So maybe on the left example, you could change the title to "Are you sure you want to proceed", and have the buttons be "Delete" and "Cancel" to show the possible confusion. (Ie: a natural answer to "Are you sure you want to proceed" would be "Yes, proceed" or just "Proceed" and would not be "Delete") Does that make sense?

packages/v4/src/content/design-guidelines/usage-and-behavior/modal/modal.md
mmenestr: @mcarrano I think the name in this case is "Database X"! But yes, maybe changing it to something else would make that more clear.

packages/v4/src/content/design-guidelines/usage-and-behavior/modal/modal.md
mcarrano: The way I look at this, the affirmative action should always be labeled to reflect the result on clicking the button. In this case, Delete. While I agree with everything said about the messaging, the fact is that users often don't read. They will glance at this and either immediately click Delete or Cancel because they got here by mistake. Using labels like Yes or Proceed forces them to do more cognitive work by reading to understand what they are saying yes to.

abchen-redhat:
@mcarrano @mmenestr Made updates according to your suggestions! Let me know what you think

packages/v4/src/content/design-guidelines/usage-and-behavior/modal/modal.md
mmenestr: This is nitpicky, but could you expand the left modal so that the header fits on one line?

mmenestr:
@mcarrano @abigaeljamie just have the one comment about the picture that has a modal with a two lined header. Other than that LGTM!

packages/v4/src/content/design-guidelines/usage-and-behavior/modal/modal.md
abigaeljamie: The "Leave page without saving?" has a repetitive first line. The UX writing best practice is to have a heading (which we do), then the body text should get right to the point. Any extra questions distracts the user. So the photo should have: 

Leave page without saving? 
Any unused information will be lost.

abchen-redhat:
@mcarrano @mmenestr @abigaeljamie All updated!

mcarrano:
Thanks @abchen-redhat @mmenestr . I will merge.

Issue #1665, opened 2020-02-03
Typography page update to clarify monospaced font type


LHinson:
Update the typography page [1] to specify that Liberation mono is the available monospaced font in PF. 

[1] https://www.patternfly.org/v4/design-guidelines/styles/typography

mmenestr:
More guidance:
Add another "table" like the one below for the font mentioned above^?

Call this "Code text"? 
Font family: Liberation Mono
etc

![image](https://user-images.githubusercontent.com/51165119/85301815-4c738080-b476-11ea-8bcb-dd4210c9e106.png)

PR #1915, opened 2020-06-26
Typography Copy

japhilli-RH:
Changing copy in the typography section of PF for brevity and clarity 
Closes [issue 1665](https://github.com/patternfly/patternfly-org/issues/1665)

### DIFF
--- a/packages/v4/src/content/design-guidelines/styles/typography/typography.md
+++ b/packages/v4/src/content/design-guidelines/styles/typography/typography.md
@@ -18,12 +18,12 @@ We use Red Hat Display and Red Hat Text.
 <Button style={{borderRadius: '0px', fontWeight: '600', paddingTop: '12px', paddingBottom: '12px', paddingLeft: '24px', paddingRight: '24px'}} variant="primary" component="a" href="https://github.com/RedHatOfficial/RedHatFont" target="_blank">DOWNLOAD</Button>
 
 ## Usage guidelines
-Use our typographic styles to communicate visual hierarchy. A consistent and logical hierarchy provides a clear pattern for users, making it easier to quickly scan and understand information on a page.
+Use typography to create visual hierarchy. A consistent and logical hierarchy makes it easier for users to quickly scan and understand information on a page.
 
 <TitleLevel
   asGrid
   title="First level title*"
-  note="*Not to be used in content block (Landing pages, login, etc.)"
+  note="*Not to be used in content block (such as landing pages and login screens)"
   styleProps={styleProps.first} />
 <TitleLevel asGrid title="Second level title" styleProps={styleProps.second} />
 <TitleLevel asGrid title="Third level title" styleProps={styleProps.third} />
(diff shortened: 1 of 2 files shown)


packages/v4/src/content/design-guidelines/styles/typography/typography.md
```diff
@@ -18,7 +18,7 @@ We use Red Hat Display and Red Hat Text.
 <Button style={{borderRadius: '0px', fontWeight: '600', paddingTop: '12px', paddingBottom: '12px', paddingLeft: '24px', paddingRight: '24px'}} variant="primary" component="a" href="https://github.com/RedHatOfficial/RedHatFont" target="_blank">DOWNLOAD</Button>
 
 ## Usage guidelines
-Use our typographic styles to communicate visual hierarchy. A consistent and logical hierarchy provides a clear pattern for users, making it easier to quickly scan and understand information on a page.
+Use typography to create visual hierarchy. A consistent and logical hierarchy makes it easier for users to quickly scan and understand information on a page.
 
 <TitleLevel
   asGrid
```
abigaeljamie, line 24: Line 26: What does "login" mean on line 26? Are we referring to a login screen or field? Also, let's remove the "etc." and instead go with something along the lines of: *Not to be used in content block (such as landing pages and login screens)  

Line 34: Remove the "etc." so that we have: *Some components use RedHatText at 700 font weight, which is RedHatText Medium (such as alerts and navigation) >> instead of "such as," we can also do "example," or "for example,"

Line 40: To keep these notes consistent, let's remove the period and capitalization of the first word in the parentheses. So we would have something like: *Not to be used in content block (only used with data visualizations when 14px is not small enough)

packages/v4/src/content/design-guidelines/styles/typography/typography.md
```diff
@@ -41,29 +41,33 @@ Use our typographic styles to communicate visual hierarchy. A consistent and log
   styleProps = {styleProps.tiny} />
 
 ## Line height
-<div style={{marginBottom: '32px'}}>When laying out text-based content, you need to factor in line-height when measuring spacing. Line height is a relative number used in CSS that represents a ratio to the text size. For example, body text uses a line height of 1.5. Body text size is 16px. 16*1.5 = 24px line height. So, when creating designs in design software, you must include the 24px line height as a part of the text.</div>
+<div style={{marginBottom: '32px'}}>Accommodating for line-height is important when measuring spacing in text-based content.
```
abigaeljamie, line 44: We have a hyphen for "line-height" here, but not anywhere else. For consistency, let's go with "line height" 

"Accommodating for line height" sounds a little confusing. I recommend restructuring this so that we give the user an actionable step: "When measuring spacing in text-based content, plan for line height."

packages/v4/src/content/design-guidelines/styles/typography/typography.md
```diff
@@ -41,29 +41,33 @@ Use our typographic styles to communicate visual hierarchy. A consistent and log
...
+</div>
 
 <Grid>
   <GridItem span={12}>
     <LineHeightTitle>
       <CheckCircleIcon color="#52A549" />
       <span style={{color: '#151515', marginLeft: '8px'}}>Correct</span>
     </LineHeightTitle>
-    <div>Note how the line height space is included when laying out with spacer elements.</div>
+    <div>Note how line height space is considered when laying text with spacer elements.</div>
```
abigaeljamie, line 55: Let's make the "do" and "don't" copy consistent. The "don't" copy is more actionable, so let's restructure this "do" copy to align: 

"Include line height space when laying text with spacer elements."

packages/v4/src/content/design-guidelines/styles/typography/typography.md
japhilli-RH: Thanks for catching these!

abigaeljamie:
Getting close, @japhilli-RH ! Looks like PF4 rejected it. You can see what needs to be fixed by clicking **Details** beside the ci/circleci at the bottom.

packages/v4/src/content/design-guidelines/styles/typography/typography.md
abigaeljamie: Np! I reviewed your changes, and they look great, but the pesky period is still in there after "enough" in "*Not to be used in content block (only used with data visualizations when 14px is not small enough.)" Once you remove that, we should be good to go!

japhilli-RH: Oops! Just fixed.

abigaeljamie:
Hi, @gdoyle1 and @mmenestr ! Are you able to approve and merge this?

gdoyle1:
@japhilli-RH 
Font family: Liberation Mono `--pf-global--FontFamily--monospace`
Font weight: 400 (regular)
Line height: 1.5
Font size: 16px
Font size CSS variable: `--pf-global--FontSize--md`

I know the Font family example is different than the rest on the page but I think it's important to call out the variable for that font style!

cc: @mmenestr

Issue #1804, opened 2020-04-20
Card usage guidelines


mcarrano:
Introduce card usage guidelines based on analysis documented in https://github.com/patternfly/patternfly-design/issues/798

Lucia worked on some in-depth documentation on when to use cards within a content view and when not to: https://marvelapp.com/22g12786

She also created this Marvel project with examples of what those Card Views will look like now: https://marvelapp.com/5996ff7

TLDR: Use a white background for card views when the content view can change (i.e. master detail, where the card can open an expandable panel on the right _also_ where the card view can toggle into a table or list view). Use a gray background for card views when the sole purpose of that content view is the cards. Reference Design issue [#834](https://github.com/patternfly/patternfly-design/issues/834) for more details (this one is specifically about pagination but gives the different examples)

Related issue: #1786

abigaeljamie:
@mcarrano , are we adding the guidelines to this existing doc? https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/card-view

mcarrano:
That's a good question. I think it might make sense to have a more general topic labeled "Cards" and incorporate card view into that. But I'm not wedded to that idea. @mmenestr @gdoyle1 what do you think?

mmenestr:
I agree, I would make a new page called "Cards" and incorporate card view into that. I would also make sure to differentiate the difference between "Cards" and "Card view" since that is a bit confusing!

abigaeljamie:
Sounds good to me, thanks!

abigaeljamie:
Sent draft to @mcarrano and @mmenestr for technical review.

mcarrano:
@abigaeljamie There is a change we made to Card View to account for a bottom footer. See https://github.com/patternfly/patternfly-design/issues/834 I wasn't sure if this was already accounted for in these updates, but if not, we should include it. Thanks!

abigaeljamie:
Thanks, @mcarrano ! I can add that in and get the image from Gina. There's a bit of convo there about color and stickiness. How much detail do we want in the doc? Should we mention the stickiness at all, or should we advise readers to add a footer with pagination and then show the image Gina created as an example?

mcarrano:
I'd avoid talking about stickiness at this point since we are still trying to figure out details of how that will work.

Issue #1786, opened 2020-04-14
Update Card View with revised spacing for cards


mcarrano:
Update Card View design guidelines to reflect new card spacing. We may want to think about renaming the "Card View" section to something like "Cards"... or keeping 2 sections so that one focuses on types of Cards vs. Card view.
**Spacing updates:**
We now recommend using compact cards by default. Compact cards use 16px spacing with 8px spacer between the header and body, and 16px between the body and footer.
We also provide another card type, large cards, that use 24px spacing with 16px spacer between the header and body, and 24px between the body and footer.
(You can find images in the [PatternFly Template File](https://www.sketch.com/s/729c2eee-e8b6-4fcd-8a79-f6faa8c30f89))
<img width="758" alt="Screen Shot 2020-06-22 at 10 15 55 AM" src="https://user-images.githubusercontent.com/30870144/85297900-5e9ef000-b471-11ea-9105-443f5c7adcd7.png">
Related issue: #1804

abigaeljamie:
Are these the guidelines to be updates, @mcarrano ? https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/card-view

mcarrano:
Yes

abigaeljamie:
Sent draft to @mcarrano and @mmenestr for technical review.

PR #1924, opened 2020-07-01
Card page revamp

abigaeljamie:
Closes [issue 1786](https://github.com/patternfly/patternfly-org/issues/1786)
Closes [issue 1804](https://github.com/patternfly/patternfly-org/issues/1804)


packages/v4/src/content/design-guidelines/usage-and-behavior/card-view/cards.md
```diff
@@ -0,0 +1,138 @@
...
+Use a card:
+- To visually group content or distinguish information from the rest of a page’s content. Examples include dashboards and catalog views. 
+- To organize content in layouts with multiple columns or grids.  
+- To compare separate groups of information.  
+- To make media such as images and videos easier to scan.  
+
+<img src="./img/dashboard.png"  alt="card view with charts"  width="990"/> 
+
+Do NOT use a card:
+- Behind a component, like a data table or a list, when it's the only section on the page. (But when there are both cards and tables on the page, you can separate the page using cards.)
```
mcarrano, line 41: Can we strike the first 2 bullets here and the screen shot below? Due to some implementation issues, we are rethinking changes to card usage for full screen views.

abigaeljamie: Thanks, @mcarrano ! I removed those bullet items and the screenshot.

packages/v4/src/content/design-guidelines/usage-and-behavior/card-view/cards.md
```diff
@@ -0,0 +1,135 @@
...
+- A card can contain a header with an image, actions, and a body.
+- A card can contain only a header and a body. 
+- A card can contain a header with an image, a body, and footer. 
+
+## Card types 
+PatternFly supports two types of cards: 
+1. **Compact (recommended)**: Compact cards use 16px spacing with an 8px spacer between the header and body and 16px between the body and footer.
+2. **Large**: Large cards use 24px spacing with a 16px spacer between the header and body and 24px between the body and footer.  
+
+<img src="./img/card-types.png"  alt="compact and standard card types"  width="779"/>
```
mmenestr, line 29: Found a small inconsistency here. In the text it's referred to as "Large" but in the image I referred to it as "Standard" card. @mcarrano which one is the correct one. If it's large let me know so I can update the image! :)

mcarrano: Good question. I think we settled on "Large." Can you confirm @gdoyle1 ? But I think it sounds wrong to say "Large card title because the size of the card title is not different. Can we update to just call this a "Card title" in either case?

mmenestr: It seems like in pf org documentation it's called "Basic". Also, I can definitely update the card titles to just say "Card title"

abigaeljamie:
Will do, @mmenestr ! Can you please email me those images so that I have all of them handy?

packages/v4/src/content/design-guidelines/usage-and-behavior/card-view/cards.md
gdoyle1: Sorry for the delayed response @mcarrano @mmenestr @abigaeljamie - Margot and I settled on:
1. **Compact (recommended)**: Compact cards use 16px spacing with an 8px spacer between the header and body and 16px between the body and footer.
2. **Standard**: Standard cards use 24px spacing with a 16px spacer between the header and body and 24px between the body and footer.  

We spoke with Matt Nolting and found that in code, you will get the "standard" spacing by default and only will receive the "compact" spacing if you apply the` pf-m-compact`. modifier. So I think that makes the most sense! I will update in sketch library too.

abigaeljamie:
Thanks for the feedback! I replaced the copy so that we have the Compact and Standard card types, and I replaced all the images from @mmenestr . The only thing I need is one more image (it appears broken in the preview because I deleted it). @gdoyle1 , do you mind sending it to me again? You sent it to me a few weeks back in Slack, but Slack erased it. It's the image that shows the sticky bottom toolbar.

PR #1925, opened 2020-07-06
Chore/1793 redesign combine docs

### LINKED ISSUE #1793: [Redesign] Combine core/react documentation
Per proposed design in #1785, we need to combine Core and React component documentation pages.

image

Our Core and React components do not map 1-1. I think the best way would be add some frontmatter as a common key upstream in Core/React. Design would have to choose what they want that common key to be (ie. About modal or About modal box @mcarrano @rachael-phillips  ?)

evwilkin:
Closes #1793 

POC showing that Core & React docs can be sourced together and displayed on a single page. Also updates Org sidenav to remove context switcher and instead show all components under the "Components" section regardless of Core vs React component.

Note that this does not incorporate work towards building out the actual design for the redesign, simply dumps both React & Core docs together onto one page.

### DIFF
--- a/packages/gatsby-theme-patternfly-org/components/example/example.js
+++ b/packages/gatsby-theme-patternfly-org/components/example/example.js
@@ -24,8 +24,8 @@ const getSupportedLanguages = className => {
 
 // This component uses hooks in order to call useMDXScope()
 export const Example = props => {
-  const html = props.html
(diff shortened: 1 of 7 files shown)


evwilkin:
Closing in favor of the work towards #1951 which incorporates redesign into the rebuild.

PR #1927, opened 2020-07-06
Toolbar doc updates

### LINKED ISSUE #1531: Revisit Toolbar design guidelines
Content in Toolbar design guidelines here: https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/toolbar  is not consistent with implementation in terms of spacing, items, and groups. Update the documentation to better sync with the actual implementation.

mcarrano:
This PR updated the Toolbar design guidelines to reflect the latest layout guidance for using the Toolbar component in a design.

Closes #1531


mcarrano:
@redallen Any idea why this build fails? It was working but then I added some extra line br3eaks to fix image/text formatting issues.

mcarrano:
@gdoyle1 @mattnolting This PR includes all proposed updates to the Toolbar design guidelines as we discussed. Please review to let me know if 1) this is technically accurate and 2) if the guidance provides designers what they need to know to use the Toolbar component effectively. Thanks.

Preview link: https://patternfly-org-pr-1927-v4.surge.sh/v4/design-guidelines/usage-and-behavior/toolbar

packages/v4/src/content/design-guidelines/usage-and-behavior/toolbar/toolbar.md
```diff
@@ -9,147 +9,126 @@ Toolbars can be included to allow a user to manage and manipulate a data set. Da
...
+ <img src="./img/filter-group.png" alt="filter group" width="355" />
 
-Items within a group have 16px spacing between them. Items that are made up of multiple parts (for example the filter dropdown and input group) are treated as one item in regards to the spacing.
+  A filter group combines multiple select filters into a single group. See [Filters](/design-guidelines/usage-and-behavior/filters) for more information about using filters.
 
+#### Search filter
+<img src="./img/search-filter.png" alt="search filter" width="355" />
 
-### Multiple Groups
+A search filter is intended to be used with a filter group as shown below when you want to include a single search field together with one or more select filters. It include an 8px spacer to the right of the item to enforce association with other filters. See the [Filters](design-guidelines/usage-and-behavior/filters) documentation for more information about using filters.
```
gdoyle1, line 31: Found that include should be changed to includes - "It includes an 8px spacer..."

packages/v4/src/content/design-guidelines/usage-and-behavior/toolbar/toolbar.md
```diff
@@ -9,147 +9,126 @@ Toolbars can be included to allow a user to manage and manipulate a data set. Da
...
-**Custom toolbar**
-![custom](./img/custom-toolbar.png)
-Custom toolbars can vary by use case. In the above example, the toolbar is primarily comprised of filters.
+The following is an example of a complex toolbar optimized for mobile.
 
-**Responsive toolbar**
+<img src="./img/toolbar-filters-mobile.png"  alt="toolbar with filters and actions hidden on mobile"  width="1308"/>
 
-The toolbar adapts to smaller viewport sizes by collapsing or hiding elements that take significant space, including filters and actions. On mobile devices, when used, pagination will be available from the footer, only.
+Here, the search filter and filter group containing three drop-down filters are placed in a toggle group that collapsed when the screen shrinks to mobile size. The two action button are part of an overflow menu that collapses to a single kabob menu. The bulk selector and sort icon button are implemented as toolbar items that remain visible at all breakpoints.
```
gdoyle1, line 122: Action buttons* instead of "action button"

packages/v4/src/content/design-guidelines/usage-and-behavior/toolbar/toolbar.md
```diff
@@ -9,147 +9,126 @@ Toolbars can be included to allow a user to manage and manipulate a data set. Da
 
 ## Elements
 
-The toolbar is a container for holding a variety of elements (toolbar items and groups) that are used to manipulate a set of data.
+The toolbar is a container for holding a variety of elements (toolbar items and groups) that are used to manipulate a set of data. All items should be centered vertically within the toolbar. There is a 24px of padding before and after the toolbar content.
```
mattnolting, line 12:
```suggestion
The toolbar is a container for holding a variety of elements (toolbar items and groups) that are used to manipulate a set of data. All items should be centered vertically within the toolbar. There is 16px of horizontal padding before and after the toolbar content.
```

mattnolting: I've identified the padding values as a potential enhancement. Currently, the padding left/right is `16px`, but I believe we should update this padding to be consistent with `pf-c-page__main-section`, `__card`, etc
https://github.com/patternfly/patternfly/issues/3240

packages/v4/src/content/design-guidelines/usage-and-behavior/toolbar/toolbar.md
```diff
@@ -9,147 +9,126 @@ Toolbars can be included to allow a user to manage and manipulate a data set. Da
...
+3. **Overflow menu:** In this example an overflow menu is used to hold global actions. But a simple button group could also be used here if you have two or less actions.
+4. **Item count**: When a filter is applied, the item count will be updated to represent the filtered subset of items. For example, “10 of 17 items.” The item count should be the right-most element.
 
+**Toolbar with pagination**
+<img src="./img/basic-toolbar-pagination.png"  alt="basic toolbar with pagination"  width="1210"/>
 
-**4. Action button group:** Any global actions that apply to selected data items are included left-aligned following filter and icon buttons. These may be placed within an overflow menu component to collapse the group as the viewport shrinks. No more than two items should be exposed as buttons. Button labels should be limited in length to preserve horizontal space.
+When pagination is used, the pagination component replaces the item count. It should be the right-most element. Here, the toolbar is shown with both full and compact pagination options.
```
mattnolting, line 87: I believe we decided to prescribe `compact pagination` always be used in header applications. Like, when a toolbar appears above a dataset.

packages/v4/src/content/design-guidelines/usage-and-behavior/toolbar/toolbar.md
mcarrano: Yes, I do remember a discussion about adjusting this. So I guess that means the toolbar examples will also need adjusting to reduce the spacing on either end. @gdoyle1 that will then also apply to the Sketch Library which has 24px spacers.

@mattnolting can you confirm this before we make those updates?

packages/v4/src/content/design-guidelines/usage-and-behavior/toolbar/toolbar.md
mcarrano: I think this guidance makes sense, although I would not want to say "always" as you could opt to use the full pagination as it applies to your use case. @gdoyle1 what do we currently default to in the Sketch template? Another area where we should try to be consistent.

packages/v4/src/content/design-guidelines/usage-and-behavior/toolbar/toolbar.md
mcarrano: Based on our conversation yesterday, I just deleted this statement.

packages/v4/src/content/design-guidelines/usage-and-behavior/toolbar/toolbar.md
mcarrano: OK. I added a statement about this.

mcarrano:
@mattnolting @gdoyle1 I have made updates in response to your comments. Based on our meeting yesterday, I decided not to say anything about the left/right padding for now and to leave the mocks as is with the 24px spacer. My preference would be to merge this, as is, and come back later and update the mockups if we need to.

@mmenestr if you can also give this a quick read and merge when ready, that would be great. Thanks!

mmenestr:
Just read through and it looks great - but these are clear guidelines and I like the way it's laid out! My only comment is that some of the images still use the overpass font - I would probably edit those to all have red hat font components. The images that need editing are:
1. Image under "Elements" section (pagination + buttons still in overpass) 
2. Images under toggle group (everything still in overpass)
3. Image under vertical dividers (pagination still in overpass) + top image doesn't have pagination but bottom one does
4. Images under "Basic toolbar" (pagination still overpass)
5. Image under Toolbar with an icon group (pagination still overpass)
6. Image under custom toolbar (pagination still overpass)
7. Image under "toolbar on mobile" (pagination still overpass)!

I also do have a question about the spacing between buttons because in wizards and modals, the spacing is 16px, but here it's 8px, so I just wanted to double check if that's correct and meant to be that way!

mcarrano:
@mmenestr Regarding the button spacing, yes, it is intentional that the toolbar uses tighter spacing. 

As for the font, I suppose that many of these mocks were created before we switched to the RH font. @gdoyle1 is there a way to globally update that in the Sketch files? Have we been updating other pages with older mockups? If this needs to be a manual update then I question whether it's worth the effort. Thoughts?

mmenestr:
@mcarrano it's odd because they do use Red Hat font for the most part! There's no way to globally update (that I know of) but we've definitely been updating the other pages with older mocks to use the new red hat fonts. I'm also happy to make those edits if you need, because I personally think it's worth the effort, just for the sake of consistency!

mcarrano:
@mmenestr OK. I believe that I've updated the images correctly as you noted above. Please take another look. Here is a link to the final Sketch artwork if you or @gdoyle1 want to download.

mmenestr:
@mcarrano all looks great, except for the secondary button in the first image which is still using overpass! Otherwise i'd say it's ready to merge. I will keep this here until you're back from PTO.

gdoyle1:
Will merge the Toolbar stuff in the Design Doc Assets Sketch file once that last change is made! @mcarrano @mmenestr 

And yes regarding the global updates from overpass to red hat - there is no easy automatic way of doing so unfortunately

mcarrano:
@mmenestr @gdoyle1 I updated the first image and hopefully this is now set to merge. Please give it another look.

mmenestr:
@mcarrano just did a check on the links and the "overflow" and "toggle menu" links don't work! I think it might be because of the capital letters in the link - should all be sentence case? Try that!

mmenestr:
@mcarrano  There is still a problem with the first overflow menu link on the page, because it links to overflow-menuS, and not overflow-menu (menu shouldn't be plural)! Other than that it should be good to go, all the other links work now :)

Issue #1935, opened 2019-12-16
Alert Group has axe violations


jessiehuff:
Currently the alert group has axe violations: id attribute value must be unique (issue with overview heading)

(There is currently a link violation as well, but it should be resolved by https://github.com/patternfly/patternfly-org/pull/1588)

redallen:
Closing in favor of #2158.

Issue #1934, opened 2019-12-16
Data list has axe violations


jessiehuff:
- id attribute value must be unique (seems to be issue with headings)

redallen:
Closing in favor of #2158.

Issue #1933, opened 2019-12-16
Login has axe violations


jessiehuff:
- id attribute value must be unique (I believe with the basic preview)
- Heading levels should only increase by one

redallen:
Closing in favor of #2158.

Issue #1932, opened 2019-12-16
Select has axe violations


jessiehuff:
- id attribute value must be unique (issue with headings)

redallen:
Closing in favor of #2158.

PR #1936, opened 2020-07-09
Create unique id for headings

### LINKED ISSUE #1929: Toolbar has axe violations
- id attribute value must be unique (basic example heading)

jessiehuff:
Fixes #1929, #1930, #1931, #1932, #1933, #1934, #1935


evwilkin:
@jessiehuff the build failed, it doesn't recognize `_` - doesn't look like lodash is imported for the `_.uniqueId` function

redallen:
This breaks all our table of contents. Please revert and properly fix.

jessiehuff:
>Please revert and properly fix by changing the headers to be unique downstream.

I don't think this change can be made downstream, but I'll open a new PR with your suggestion to exclude the specific h1 and h2 ids that are used here.

PR #1954, opened 2020-07-16
support example h3s

redallen:
Stop-gap for patternfly.org refactor

### DIFF
--- a/packages/gatsby-theme-patternfly-org/components/example/example.js
+++ b/packages/gatsby-theme-patternfly-org/components/example/example.js
@@ -1,6 +1,7 @@
 import React, { useLayoutEffect } from 'react';
 import { LiveProvider, LiveEditor, LivePreview, LiveError } from 'react-live';
 import { Badge } from '@patternfly/react-core';
+import * as PatternflyReactCore from '@patternfly/react-core';
 import { useMDXScope } from 'gatsby-plugin-mdx/context';
 import { ExampleToolbar } from './exampleToolbar';
 import { AutoLinkHeader } from '../autoLinkHeader/autoLinkHeader';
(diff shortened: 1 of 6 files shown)


packages/gatsby-theme-patternfly-org/components/example/example.js
```diff
@@ -56,7 +57,10 @@ export const Example = props => {
   }
   const exampleName = title.replace(/-/g, ' ').replace(/  /g, '-');
   const fullscreenLink = `${location.pathname}/${slugger(title.toLowerCase())}`;
-  const scope = useMDXScope();
+  const scope = {
+    ...PatternflyReactCore,
```
redallen, line 61: This is so React devs don't have to `import { Stuff } from '@patternfly/react-core';` at the top of each MDX file. The whole thing is included in the bundle anyways so we might as well inject it in each example.

packages/gatsby-theme-patternfly-org/gatsby-node.js
```diff
@@ -62,7 +62,8 @@ exports.onCreateNode = ({ node, actions, getNode }) => {
     const source = parent.sourceInstanceName;
     const componentName = path.basename(node.fileAbsolutePath, '.md');
 
-    let { section = 'root', title, propComponents = [''], wrapperTag = 'main' } = node.frontmatter;
+    let { section = 'root', id, title, propComponents = [''], wrapperTag = 'main' } = node.frontmatter;
```
redallen, line 65: We're moving away from 
``````
```
title: Component name
```
``````

in favor of
``````
```
id: Shared component name across core, react, and org
```
``````

File: packages/gatsby-theme-patternfly-org/helpers/codeTitleTransformer.js [hunk 3]
```diff
@@ -0,0 +1,43 @@
+const visit = require('unist-util-visit');
+
+function exampleCaptions() {
```
redallen, line 3: Basic `h3` support as a rehype plugin. This should be a remark plugin but `gatsby-theme-mdx` nukes changes we make to the MDAST, so it has to be a rehype plugin.

packages/gatsby-theme-patternfly-org/helpers/extractExamples.js
```diff
@@ -35,6 +35,27 @@ module.exports = {
         if (match) {
           id = getId(match[1]);
         }
+        else {
+          // Starting from node, look up for h3s
+          let startLooking = false;
```
redallen, line 40: This part uses the MDAST and is for core examples.

PR #1981, opened 2020-07-22
feat(webpack): add prerendering

### LINKED ISSUE #1979: Enable prerendering on Org

redallen:
Closes #1979

**Changes:**
- Separate out webpack config into server/client builds. Merge results using [webpack-merge](https://www.npmjs.com/package/webpack-merge)
  - Server build (inspired by [preact-cli](https://github.com/preactjs/preact-cli/blob/e9bc3c2df1cdd50ca68ea8061a728eeca83a7a03/packages/cli/lib/lib/webpack/prerender.js)):
    - Outputs `.cache/ssr-build/ssr-bundle.js`
    - Runs in CJS environment (can ```require(`${process.cwd()}/.cache/ssr-build/ssr-bundle`)``` in `prerender.js`)
    - Uses [`<ServerLocation>`](https://reach.tech/router/api/ServerLocation) for route-specific rendering
  - Client build:
    - Outputs `public` folder
    - Prerenders HTML pages using server bundle (each route in `routes.js` gets mapped to an `HTMLWebpackPlugin` in `getHtmlWebpackPlugins.js`)
      - Inspired by [render-html-plugin](https://github.com/preactjs/preact-cli/blob/e9bc3c2df1cdd50ca68ea8061a728eeca83a7a03/packages/cli/lib/lib/webpack/render-html-plugin.js)
      - Follows `template.ejs` template for HTML
      - Use [`ReactDOM.hydrate`](https://reactjs.org/docs/react-dom.html#hydrate) in production
- Add Google Analytics
    - Copied current GA scripts from [gatsby-plugin-google-analytics](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-plugin-google-analytics/src)
- Add `pathPrefix: '/v4'`
  - Dev environment does not use a path prefix
  - Absolute `<Link>`s have path prefix added
  - Need a `<ConfigContext>` to pass this down to `<Link>` component (`@reach/router` has no way of sharing its `BaseContext` from `<Router basepath={LayoutOptions.pathPrefix}>`) 
- Add Algogia scripts
  - Doc search works!
- Add back `height: 100%` to 3 outermost `#root`, `#ws-router`, and `.ws-page-layout` divs
- Update TODO with all these changes

**TODO:**
- Create CLI webpack wrapper app with encapsulated config
    - Stop doing ```require(`${process.cwd()}/patternfly-docs.config`)```
- Use [favicons-webpack-plugin](https://github.com/jantimon/favicons-webpack-plugin#advanced-usage) for favicons


packages/theme-patternfly-org/helpers/configContext.js
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export default React.createContext({});
```
jschuler, line 3: can you init it with context object keys set to null? so it's easier to tell what this context contains

packages/theme-patternfly-org/scripts/getHtmlWebpackPlugins.js
```diff
@@ -0,0 +1,19 @@
...
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+const prerender = require('./prerender');
+const { googleAnalyticsID } = require(`${process.cwd()}/patternfly-docs.config`);
+
+module.exports = (routes, isProd) =>
+  Object.entries(routes).map(([url, { title }]) => 
+    new HtmlWebpackPlugin({
+      template: path.resolve(__dirname, '../template.ejs'),
+      filename: `${url}/index.html`.replace(/^\/+/, ''),
+      title: `PatternFly 4${title ? `- ${title}` : ''}`,
```
jschuler, line 11: is this the title for each page? Just a suggestion, but I think a bullet instead of a hyphen would look good here as well, since the title can also contain hyphens. If you hover over a github tab you'll see what I mean :)

packages/theme-patternfly-org/template.ejs
```diff
@@ -0,0 +1,37 @@
...
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width,initial-scale=1,shrink-to-fit=no">
+    <%= htmlWebpackPlugin.tags.headTags %>
+  </head>
+  <body>
+    <div id="root">
+      <%= prerendering %>
+    </div>
+    <script
```
jschuler, line 12: Usually it is recommended that js should appear before the body closing tag, and css in the head tag

package.json
```diff
@@ -10,11 +10,11 @@
     "build": "yarn build:v4",
     "build:v3": "yarn workspace patternfly-org-3 build && yarn copy:v3",
     "build:v4": "yarn workspace patternfly-org-4 build && yarn copy:v4",
-    "copy:v3": "mkdir -p build/patternfly-org && cp -r packages/v3/_site build/patternfly-org/v3",
-    "copy:v4": "mkdir -p build/patternfly-org && cp -r packages/v4/public/ build/patternfly-org/v4 && cp -r build/patternfly-org/v4/assets build/patternfly-org/assets",
+    "copy:v3": "rm -rf build/patternfly-org/v3 && mkdir -p build/patternfly-org && cp -r packages/v3/_site build/patternfly-org/v3",
```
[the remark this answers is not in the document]
redallen: The clean target already does `rm -rf build`. The structure of the folder is:
```
build
└── patternfly-org
    ├── assets
    ├── v3
    └── v4
```
and I don't want to necessarily remove `assets` or `v3` to do a clean `v4` build.

packages/theme-patternfly-org/helpers/configContext.js
redallen: Yes, good idea!

packages/theme-patternfly-org/scripts/getHtmlWebpackPlugins.js
redallen: I just went with what our site currently uses. I'll use `•` instead :)

packages/v4/src/pages/home.js
```diff
@@ -26,8 +26,8 @@ import './home.css';
 const aboutPatternFly = `Hello, we're PatternFly.`;
 const patternFlyPrinciples = 'Flyer fundamentals';
 
-export const HomePage = ({ location, layoutOptions }) => (
-  <SideNavLayout location={location} hasGdprBanner {...layoutOptions}>
+export default ({ location, layoutOptions, children }) => (
```
evwilkin, line 29: It looks like `children` is destructured but never used?

packages/theme-patternfly-org/template.ejs
redallen: Right, but we don't want to block loading the entire site for Algogia CSS. Usually it takes the user longer than the 50Kb of Algogia CSS loads before they click the search box.

packages/v4/src/pages/home.js
redallen: Good call! I played around with it at one point.

packages/theme-patternfly-org/scripts/getHtmlWebpackPlugins.js
jschuler: cool, there should also be a space after the 4

redallen: Fixed!

PR #1998, opened 2020-07-30
feat(redesign): add content toc

redallen:
- Add sticky TOC for content on right hand side
- Use tabs instead of `<ul>` for switching between sources
- Add 404 page


evwilkin:
@redallen your note mentions this PR switches to tabs in place of `ul` for switching sources, but I'm still seeing the `ul` used.  Also wanted to confirm for this PR that "CSS Variables" shouldn't be listed in the sticky TOC yet, and that the Props table is still a todo?
Screen Shot 2020-07-31 at 10 18 45 AM

redallen:
@evwilkin I have a lot of unpushed changes and opened this PR for @mcoker to help out with the sticky TOC!

Yes, props won't be in this PR. I hope to include it in the next one.

packages/theme-patternfly-org/components/tableOfContents/tableOfContents.css
```diff
@@ -0,0 +1,15 @@
+.ws-toc {
+  top: var(--pf-global--spacer--md);
+  order: 1;
+  position: sticky;
+  position: -webkit-sticky; /* Safari */
```
jschuler, line 5: maybe should add postcss autoprefixer to webpack instead

jschuler:
Also, when i click on a tab (React | Design guidelines | Core) the active doesn't change and it still has anchor hover over style that shows underline

packages/theme-patternfly-org/templates/mdx.js
```diff
@@ -20,42 +22,75 @@ const InlineAlert = ({
...
-                  <li key={source}>
+            <Tabs activeKey={activeSource} className="ws-source-tabs">
+              {sourceKeys.map(source => (
+                <Tab
+                  key={source}
+                  eventKey={source}
+                  title={(
                     <Link to={source}>
-                      {source}
+                      {capitalize(source.replace(/-/g, ' '))}
```
evwilkin, line 78: Currently users can only click directly on the link within the Tab, rather than anywhere on the Tab (including the padding), to select that Tab.

packages/theme-patternfly-org/components/tableOfContents/tableOfContents.css
redallen: Done

packages/theme-patternfly-org/templates/mdx.js
redallen: Fixed!

PR #2026, opened 2020-08-10
Feat/code splitting

redallen:
Instead of all documentation residing in one big bundle, this uses `splitChunks: { chunks: 'all'  }` to split by route in `routes.js`.

**Changes:**
- Split main content of pages into separate bundles
  - Use [react-ssr-prepass](https://www.npmjs.com/package/react-ssr-prepass) to prerender async components
    - Fix react-charts hanging prerendering by replacing problematic `d3-timer` module on server
  - Use asyncComponentFactory to preload links into a cache onMouseOver. Wait up to 500ms on current page before navigating to show a "Loading" component
  - Cut initial bundle size in 1/5:
image

- Require a few less modules in react-tokens by using ESM. Separate out duplicate CSS
- Consistent page layout for single/multi-source pages
  - ~~Hide TOC at 1300px for viewing on 15" Macs~~
  - Move TOCs to main bundle for no layout changes while loading
- Make key unique for TOC
- Remove react-refresh (it's full of caveats like [not working on class components](https://github.com/pmmmwh/react-refresh-webpack-plugin#caveats))
- Clean up pageData pipeline


packages/theme-patternfly-org/components/footer/footer.js
```diff
@@ -16,227 +16,223 @@ import redhatLogo from "./RHLogo.png";
...
+                      to={"/get-started/about"}
+                      aria-label="Get started with PatternFly 4"
+                    >
+                      Get started
+                    </Link>
+                  </li>
+                  <li className="pf-c-nav__item">
+                    <Link
+                      className="pf-c-nav__link"
+                      to={"/documentation/react/components/aboutmodal"}
```
jschuler, line 50: This link won't work. Any plan to setup forwarding for old links to new ones?

jschuler: Some other footer links are not correct as well, maybe go through all of them?

packages/theme-patternfly-org/components/tableOfContents/tableOfContents.js
```diff
@@ -52,17 +52,17 @@ const ReleaseNotesTOC = () => (
...
   if (Array.isArray(item)) {
     return (
-      <ul key={item.join('-')}>
+      <ul key={index}>
         {item.map(renderItem)}
       </ul>
     );
   }
 
   return (
```
jschuler, line 64: When i click on a ToC link it doesn't scroll the item into view. Same when you try and load the page to an item directly, i.e. `https://patternfly-org-pr-2026-v4.surge.sh/v4/components/drawer#breakpoint`

packages/theme-patternfly-org/helpers/asyncComponentFactory.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+    render() {
+      if (cache[route]) {
+        return React.createElement(cache[route]);
+      }
+      AsyncComponent.preload().then(() => this.setState({ isLoaded: true }));
+
+      // Simple loading state
+      return React.createElement('div',
+        { style: { height: '100vh' } },
```
jschuler, line 26: We have a Spinner component as well

packages/theme-patternfly-org/components/footer/footer.js
redallen: Yep! After fullscreen pages I'll run the site through a dead link checker and we'll resolve all broken links (hopefully) in 1 PR.

packages/theme-patternfly-org/scripts/getHtmlWebpackPlugins.js
```diff
@@ -1,23 +1,26 @@
...
-      },
-      scriptLoading: 'defer',
-      inject: false
-    }),
-  ).concat(
+async function getHtmlWebpackPlugin(url, isProd, title) {
+  return new HtmlWebpackPlugin({
+    template: path.resolve(__dirname, '../templates/html.ejs'),
+    filename: `${url}/index.html`.replace(/^\/+/, ''),
+    title: `PatternFly 4${title ? ` • ${title}` : ''}`,
```
jschuler, line 10: In the preview i am not seeing the title correctly for the browser tab

packages/theme-patternfly-org/components/tableOfContents/tableOfContents.js
redallen: Good catch! I'm not sure why it's not scrolling... will investigate.

packages/theme-patternfly-org/helpers/asyncComponentFactory.js
redallen: I can't import from patternfly in this file because it is `require()`d from a NodeJS context.

It'd take a little bit of refactoring but could be possible in the future. As it is you'll rarely see this thanks to preloading!

packages/theme-patternfly-org/scripts/getHtmlWebpackPlugins.js
redallen: Yep, it seems the HtmlWebpackPlugin isn't respecting this option. I'll fix it!

packages/theme-patternfly-org/components/tableOfContents/tableOfContents.js
redallen: The reason it's not scrolling is that surge sometimes postpends `/` to URLs... https://patternfly-org-pr-2026-v4.surge.sh/v4/components/about-modal/#custom-background-image does NOT work, but https://patternfly-org-pr-2026-v4.surge.sh/v4/components/about-modal#custom-background-image does.

jschuler:
Saw some artifacts on this page as it loads, in this case the toast alerts are briefly seen, can reproduce by setting network speed in dev tools to 3G.
`https://patternfly-org-pr-2026-v4.surge.sh/v4/components/alert-group/html/`

redallen:
@jschuler That problem exists on patternfly.org today and there isn't a great way to solve it until we properly make take screenshots of fullscreen pages instead of using hacky CSS transforms.

packages/theme-patternfly-org/components/tableOfContents/tableOfContents.js
jschuler: no matter what I do, a slash is added before the hashtag, even when clicking on the second link here. Or when the page is already open and I delete the slash and hit Enter it is also re-inserted

evwilkin:
May be work for a follow-up PR but I noticed the TOC is cutoff in some situations:
Screen Shot 2020-08-18 at 10 50 02 AM

The width of content also changes between React and HTML tabs for some components, which again causes TOC to be pushed off screen:
Screen Shot 2020-08-18 at 10 54 44 AM

Screen Shot 2020-08-18 at 10 54 35 AM

evwilkin:
Clicking on the current page in the sidenav causes the page to reload, which is especially noticeable (though not really a bug) for large visuals like Background image & Branch components.

However, for several Demos such as Page it causes the example to size up then back down quickly, and for others such as Table it causes strange flashes as seen below:
Screen Shot 2020-08-18 at 11 15 16 AM

packages/theme-patternfly-org/pages/404.js
```diff
@@ -4,16 +4,14 @@ import { PageSection, Title } from '@patternfly/react-core';
 import { SideNavLayout } from '../layouts/sideNavLayout/sideNavLayout';
```
evwilkin, line 4: It looks like the `SideNavLayout` import can be removed here

evwilkin:
Navigating to the 404 page and back seems to always through a lengthy "Loading" prompt, even when you've previously visited both the 404 page and the previous page - is there any way to clean this up or only require loading on the first 404 page visit?

redallen:
@evwilkin Could you demonstrate what you mean? The 404 page is included in the main bundle of the site and should never show the "Loading..." text. I tried to reproduce on the preview site but was unable.

packages/theme-patternfly-org/components/tableOfContents/tableOfContents.js
redallen: I _think_ this is mostly fixed. I'd love steps to reproduce if you can.

jschuler:
Hm, clicking on a ToC link still does not scroll the page for me, example (in Chrome)
open `https://patternfly-org-pr-2026-v4.surge.sh/v4/components/alert`
Click on the `Alert timeout` example in the ToC, I can see the browser URL changes to `https://patternfly-org-pr-2026-v4.surge.sh/v4/components/alert#alert-timeout` but nothing happens

Also, the browser tab title is set correctly on first load, but when changing pages it keeps the first one

evwilkin:
@redallen the 404 navigation delay was on my local build, but I do see it being a clean transition on the preview site as you mentioned - thanks!

The spam clicking issue looks to be the same - click "Table" under the demos section in the side nav.  The fix does look to be working locally for me though, but not on the preview link.

jschuler:
LGTM

PR #2049, opened 2020-08-25
Add demo tabs

redallen:
Assign demos to components per the strawpoll.

~~Search input/Search Input have duplicate entries until I bump react-docs after https://github.com/patternfly/patternfly-react/pull/4725 is released.~~ Bumped.


mmenestr:
I think people might have an issue with the fact that design guidelines is in the middle - I think it should be either at the beginning or the end so that the code stuff sticks together! I also wonder if it should be in a tab, or if it should just be appended to the HTML and React tabs that already exist...

redallen:
I'll change the order so design guidelines is at the end.

Appending to an existing tab would break the rule of one MD file per page. To get around that we'd have to concatenate the demo MD files into the component MD files upstream in Core/React which in some cases (like Table) would make the file 4k+ lines long.

PR #2055, opened 2020-08-25
feat(toc): redesign

redallen:
Implement https://marvelapp.com/prototype/b628dag/screen/72220616 and add scroll spy.

Thanks to @mcoker for [HTML/CSS](https://codepen.io/mcoker/pen/YzqZQqK).

@rachael-phillips Check out pages like https://patternfly-org-pr-2055-v4.surge.sh/v4/layouts/flex which have a huge TOC. What is the max-height you'd want to allow before just showing first-level items?

### DIFF
--- a/packages/theme-patternfly-org/components/autoLinkHeader/autoLinkHeader.js
+++ b/packages/theme-patternfly-org/components/autoLinkHeader/autoLinkHeader.js
@@ -20,7 +20,7 @@ export const AutoLinkHeader = ({
   headingLevel,
   children,
   metaText,
-  ...props
+  className
 }) => {
   const slug = slugger(children);
 
@@ -29,7 +29,8 @@ export const AutoLinkHeader = ({
       id={slug}
       size={sizes[size]}
       headingLevel={headingLevel || size}
-      {...props}>
+      className={`ws-heading ${className}`}
+    >
       <a href={`#${slug}`} className="ws-heading-anchor" tabIndex="-1" aria-hidden>
         <LinkIcon className="ws-heading-anchor-icon" style={{ verticalAlign: 'middle' }} />
       </a>
--- a/packages/theme-patternfly-org/components/footer/footer.css
+++ b/packages/theme-patternfly-org/components/footer/footer.css
@@ -4,6 +4,8 @@
   background-color: #1a1a1a !important;
   padding: 48px;
   padding-top: 32px;
+  /* Hide long overflowing tocs */
+  z-index: 1;
 }
 .ws-org-pfsite-l-footer .ws-org-pfsite-footer-menu-list-title {
   margin: 0;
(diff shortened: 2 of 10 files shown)


mmenestr:
Looks really good!! Just a couple of things:

1. Need to add 32px padding to the right of the table of contents to avoid this issue of the text not wrapping and ending at the edge of the browser.
image

2. Is there any way the table of contents can scroll with you and have the selected item in view at all times. When I went to the flex view and was scrolling through the page, I was expecting that the highlighted part of the table of contents would always be in view, and it was at first, but once you got to an item below the "fold" the selected item would go out of view.

3. Table of contents placement changes from page to page and is way further to the left in some cases. That causes it to overlap with content when you scroll down ([see this page](https://patternfly-org-pr-2055-v4.surge.sh/v4/charts/area-chart)). I suspect it's because the text isn't made to wrap, so it just takes on the width it needs for each item to fit on one line? But I would make the table of contents fixed, to the immediate right of the content area which I thought was fixed at 990px? But maybe that changed
image

4. Issue with table of contents overlapping the footer if you scroll all the way down ([see here](https://patternfly-org-pr-2055-v4.surge.sh/v4/charts/about)):
image

redallen:
1. Fixed.
2. I'm not sure what you mean. A screenshot of what is highlighted and what you're expecting would be helpful. The algorithm works by comparing header positions to the top of the page (`scrollTop` prop). It'd be tricky, but I suppose I can target the middle of the view instead.
3. This is a problem with the props table being too wide. Despite being given 832px, this table is deciding to be 997px wide. I've added some classes to fix this.
4. Fixed with by changing z-indexes of the footer sections.

mmenestr:
1. Can you add 32px, instead of just 16px? In some cases it still looks too visually close 
image

2. This might be a non-issue now that you've truncated the really long table of contents. But essentially, in the flex layout before, If I scrolled all the way down in the page, the table of contents remained fixed to the tip, so once an item really low in the bottom of contents got selected I couldn't see it unless I scrolled the table of contents manually myself. (again can't recreate so that's why there's not picture). The other issue with that is that if I scrolled all the way to the bottom of the table of contents and page, but then scrolled up to the top of the page, but not the table of contents, it would stay scrolled to the bottom so when I reached the top of the page, I would see the bottom of the table of contents (like the "Table of Contents" header wasn't even visible anymore). That I was able to get a screenshot of:

Screen Shot 2020-08-26 at 12 31 35 PM

3. Although it seems better, the table of contents still moves around from page to page... see difference between the "alert group" and the "Application launcher"  table of contents placement!

4. This seems to have been fixed yes! Thank you :) 

(NEW) 5. Found a bug in Release notes where it can't identify which of the "other" or which of the "component" sub header it's on, so it highlights all of them.
image

redallen:
1. Done.
2. It'd still be an issue for short viewports. I've fixed it by adding some JS that checks bounding boxes and assumes the page header height.
3. Per our slack discussion, I've left-aligned this 80px to the right of the main content.
5. Unique header ids are a complicated problem that's outside the scope of this PR. I'd be happy to address it in another PR later. The current site suffers the same problems.

PR #2062, opened 2020-08-28
Feat/design guidelines placeholder

### LINKED ISSUE #2061: Add placeholder text for design guide tabs with nothing in them
Add the following placeholder text when design guidelines have no content in them:

"We’re working on some guidelines for this component. If you have any questions in the meantime, feel free to [get in touch with us](https://www.patternfly.org/v4/get-in-touch/)."

also would be great if this could be automated so that whenever a new component is created, and a new design guideline tab is added, this placeholder automatically gets placed in the empty file. That way there is something thing until we can get to creating guidelines for it!

redallen:
Closes #2061.

Doesn't autogenerate files, but does ensure the tab is always there and has some placeholder content if there is no file.

Check out these pages:
- Bullet chart
- Donut utilization chart
- Legend chart
- Scatter chart
- Themed charts
- Threshold chart
- Tooltip chart
- Sparkline chart
- Label group
- Menu
- Search input
- Splitter

### DIFF
--- a/packages/theme-patternfly-org/scripts/md/parseMD.js
+++ b/packages/theme-patternfly-org/scripts/md/parseMD.js
@@ -145,7 +145,7 @@ function toReactComponent(mdFilePath, source) {
     // Add custom PatternFly doc design things
     .use(require('./anchor-header'))
     .use(require('./styled-tags'))
-    // Transform HAST object to JSX string, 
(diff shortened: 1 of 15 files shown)


redallen:
Thanks for the report @mmenestr , will fix this afternoon!

PR #2073, opened 2020-09-04
remove Components and demos used

redallen:
Missed these in #2051 since their headings were different


mcarrano (approved):
Looks good to me.

PR #2044, opened 2020-08-20
move UX writing section, fix tables

redallen:
Closes #2034

PR #2074, opened 2020-09-04
Fix/fullscreen thumbnails

### LINKED ISSUE #1660: Render fullscreen image thumbnails as <img> tags
To make maintenance easier, currently we render fullscreen previews using CSS transforms and some janky Javascript.

```css
.ws-preview-fullscreen {
  position: relative;
  pointer-events: none;
  width: 1280px;
  height: 800px;
  transform-origin: top left;
  /* transform: scale added in JS based on parent element width */
}

.ws-preview-container {
  position: relative;
  display: block;
}

.ws-preview__thumbnail-link {
  position: relative;
  display: inline-block;
  line-height: 0;
  overflow: hidden;
}

.ws-preview__thumbnail-link::before,
.ws-preview__thumbnail-link::after {
  position: absolute;
  z-index: 500;
  transition: .2s;
  opacity: 0;
}

.ws-preview__thumbnail-link:hover::before,
.ws-preview__thumbnail-link:hover::after {
  opacity: 1;
}

.ws-preview__thumbnail-link::before {
  top: 0;
  right: 0;
  bottom: 0;
  left: 0;
  content: '';
  background-color: rgba(0,0,0,.5);
}

.ws-preview__thumbnail-link::after {
  top: 50%;
  left: 50%;
  transform: translate(-50%, -50%);
  font-family: "Font Awesome 5 Free";
  content: "\f35d";
  font-size: 52px;
  color: rgba(255,255,255,.4);
}
```

```js
  if (isFullscreen) {
    useLayoutEffect(() => {
      const handleResize = () => {
        const resizeWidth = Math.min(
          document.getElementsByClassName('ws-example')[0].clientWidth,
          800
        );

        const scale = resizeWidth / 1280;

        setPreviewStyle({ transform: `scale(${scale})` });
        setPreviewContainerStyle({ height:`${scale * 800}px`, width:`${scale * 1280}px` });
      }

      if (!previewStyle.transform) {
        handleResize();
      }
      window.addEventListener('resize', handleResize);

      return () => window.removeEventListener('resize', handleResize);
    });
  }
```

~~We need a script that goes through a local `public` folder of a site using gatsby-theme-patternfly-org, takes screenshots, saves them back to the public folder with a cache-busting SHA like `public/img/${SHA}-fs-${componentName}-${exampleName}.png` and runs a regex to replace a placeholder `/img/${componentName}-${exampleName}.png` in `public/documentation/**.html`. We also need to make sure our JS doesn't rerender over it, this is the tricky part.~~

On the build side taking screenshots can be a post-build step. Do a build, serve it, then in another script `require('routes.js');` and visit all fullscreen pages taking 800x600px screenshots using [`puppeteer-cluster`](https://www.npmjs.com/package/puppeteer-cluster). Save them as `[route]/[example-title]-[jpg-data-hash].jpg` to the output folder webpack is configured with, which is `public`. Have script replace occurrences of `[route]/[example-title].jpg` with `[route]/[example-title]-[jpg-data-hash].jpg`. Also save them to a local folder which is copied with `copy-webpack-plugin` to `public` for use in dev mode. If the size is manageable enough, do not add them to `.gitignore` and have the `release` job in CI also commit and update the images.

On the client side loading the screenshots can be as simple as `` which will be replaced in the post-build step with the cache-busting SHA. An appropriate fallback can be shown for dev mode.

redallen:
Add fullscreen image thumbnails under the `generated` folder alongside generated JSX for component documentation.

The generated routes index has been updated to include both examples and fullscreen-only examples (previously examples were _guessed_ from the TOC).

Screenshots are generated using `puppeteer-cluster`. Once serving the docs, run `yarn screenshots <serving url>`. Serving from a build is the fastest way to generate these (ie `yarn build` -> `yarn serve` -> `yarn screenshots http://localhost:5000/v4`).

Images are imported by webpack via checking if an image matching `./${source}/${slugger(title).replace(/-/g, '')}.jpg` exists. A fallback is imported in `example.js` in case a screenshot hasn't been taken yet.

A few other bug fixes made their way in here like adding React demos and fixing example captions for the React table demo.

## Maintenance
CI could run this task and commit screenshot changes as a part of lerna (on my local machine `yarn screenshots http://localhost:5000/v4` takes 16s). I consider that outside the scope of this PR but good future work. For now when bumping react-docs we will also need to locally build and bump the screenshots.

Closes #1660


jschuler:
Looks good! Can the images have a little higher resolution? These look a bit blurry 
https://patternfly-org-pr-2074-v4.surge.sh/v4/demos/primary-detail
(also noticed this demo page does not scroll the ToC btw.)

jschuler:
oh, and are there different image sizes for mobile?

redallen:
The images are 800x450, which is 1920x1080 downscaled to 800px width. The max width images can be is 800px per https://github.com/patternfly/patternfly-org/pull/2044#issuecomment-678322467. There are no different image sizes on mobile.

If we want to optimize for mobile, doing things like removing lodash from `@patternfly/react-table` (70kb) will go farther than making a `srcset` for images that are currently ~20kb each. Also, adding functionality like search and the version switcher are easier and better starting points to improving the mobile experience IMO.

mmenestr:
@redallen agreed, looks good but the images look super low res
image

redallen:
@jschuler @mmenestr I've switched to using PNGs which have less graininess than JPGs at the cost of a few more KBs.

evwilkin (approved):
LGTM

jschuler (approved):
LGTM

PR #2154, opened 2020-09-22
Error message guidelines revamp

### LINKED ISSUE #2096: UX writing style guide edits: Error messages examples & pattern
[UX writing style guide: Error messages](https://www.patternfly.org/v4/design-guidelines/content/error-messages) 

1. Suggest a pattern for folks to use to make sure error messages are clear and useful: Description + Reason + Resolution   
2. Add Schweda's error message example:
- Before:  
++++++
**Not connected to the host**
Unable to log into <SystemIPAddress> automatically.
You need to provide your authentication credentials every time you connect to this machine.
You may want to set up your SSH keys for automatic login.
++++++

- After:
**Login failed**
The SSH key for auto-login is either not available, is unauthorized, or is password protected.
To manually log in to the host, click **Log In**.  

3. Replace the first error message before/after example. 
4. Add example for leading with the benefit.

abigaeljamie:
Closes #2096

### DIFF
--- a/packages/v4/src/content/design-guidelines/content/error-messages.md
+++ b/packages/v4/src/content/design-guidelines/content/error-messages.md
@@ -19,46 +19,67 @@ _To manually log in to the host, click **Log in.**_
 
 Keep these best practices in mind when crafting error messages:
 
-**Don’t blame users**. A user should never feel like the error is their fault. Avoid language like “You did something wrong.” Depending on your message, you may need to use the passive voice instead of the active voice so that you don't assign blame to the user.  
+**Don’t blame users**. A user should never feel like the error is their fault. Avoid language like “You did something wrong.” Depending on your message, you may need to use the passive voice instead of the active voice so that you don't assign blame to the user.   
+
+<div class="ws-content-table">
 
 | **Before**                      | **After**                  |
 |:-------------------------------:|:--------------------------:|
-| You did not provide your authentication credentials. | Authentication credentials weren't provided.|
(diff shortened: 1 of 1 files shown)


mmenestr:
@evwilkin or @redallen, can't recall which one of you did this, but is the following string `<div class="ws-content-table">` so that the table formatting is the same for each table?  And should we therefore make sure not to delete those? Just want to make that known if that's the case!

redallen:
That string just adds:
```css
.ws-content-table > table {
  table-layout: fixed;
  width: 100%;
}
```
You don't need it for any of the tables. It just might make some look more like you want.

mmenestr:
@redallen I know we don't _need_ it, but if we want all the tables to be the same width, and the columns to be even should we add it? Is that what that css does? I just remember one of you did something to make that happen, and in the preview, the tables aren't all the same width (ie: some of the "before" columns are wider than others, and we want them to be the same width for all tables)

evwilkin:
@mmenestr @redallen yes we added those wrapping divs to apply the table styling that ensures the column widths are the same down the entire page, so you don't get the staggered "Before/After" column headers for each section.  The one catch is to make sure there's a blank line before & after the `<div>`, otherwise you'll pretty easily see the table styling not applied.

abigaeljamie:
Thanks for the help, @evwilkin , @redallen , and @mmenestr ! I added the div tags, and the tables now look even 👍

abigaeljamie:
H, @mmenestr ! Are you able to review and merge this?

mmenestr (approved):

PR #2161, opened 2020-09-23
Add 2020.12 release highlights

mcarrano:
This PR is adding a folder to hold release highlights markdown docs in the Org repo. 

@evwilkin @redallen The primary purpose of this is to have a place for this content to live that is open and outside of the RH firewall. If you think this content should live elsewhere, I'm open to ideas. Please review and merge if you are OK with this.


redallen:
Hey @mcarrano , I like this content and the "What we’re working on..." section feels similar to a roadmap to me. The rest I believe belongs in our Core/React release notes (but we've historically made those just a changelog -- I believe that should change).

I'm all for adding a roadmap to the site, or if we don't want to update it every release, just to the GitHub repo. How often would we update this file?

mcarrano:
@redallen The idea is to create this after every release. The goal was to provide a condensed and easy to consume alternative to reading the full set of release notes or the roadmap project board that I maintain. I started to share this out with stakeholders as a Google doc after the 2020.11 release and have gotten good feedback. People seemed to appreciate getting something in this condensed format. The only issue that was raised is that the Google doc is not viewable outside of Red Hat which led me to create this in Markdown and push to the repo. This is not intended to be exposed on the website.

So yes, I recognize some redundancy with the release notes but that was intentional. If we wanted to change how we present release notes, I'd consider that, too. But the main point is that the current change log style release notes are not helpful to most people. What are your thoughts @evwilkin ?

evwilkin:
@mcarrano @redallen I agree that we've already received very positive feedback from teams on this condensed release summary doc format, so it's worth moving forward with.  I hadn't considered incorporating it into the release notes, I'm open to discussing this more - do you think adding this before our existing notes format and then naming the existing format "Full list of release notes" or something to that effect would be worth looking at?

redallen:
Yes, I think a format that includes a few paragraphs of highlights and then a list of commits would work awesome! See [React 17's notes](https://reactjs.org/blog/2020/08/10/react-v17-rc.html) for example.

evwilkin:
@redallen @mcarrano let's keep the conversation open around optimizing our release notes/roadmap/release summary, which we can base on feedback from other teams & consumers utilizing this information as it's refined.  For now I'm going to approve this as we still need to store these documents publicly for now as Matt continues creating them for the 2020.13 release.

evwilkin (approved):
LGTM - thanks!

PR #2170, opened 2020-09-30
Fix/table breakpoints

### LINKED ISSUE #874: Documentation tables are too wide and no horizontal scroll to allow viewing them
Documentation tables are too wide to view at certain viewports and there's no horizontal scroll to even get to them.

Example, but on a lot or all HTML pages
https://www.patternfly.org/v4/documentation/core/components/alert#alert

image

redallen:
Closes #874.

### DIFF
--- a/packages/theme-patternfly-org/scripts/md/mdx-hast-to-jsx.js
+++ b/packages/theme-patternfly-org/scripts/md/mdx-hast-to-jsx.js
@@ -37,7 +37,7 @@ function toJSX(node, parentNode = {}, options = {}) {
   }
 }
 
-function compile(options = { }) {
+function compiler(options = { }) {
   this.Compiler = function(tree) {
     return toJSX(tree, {}, options);
   }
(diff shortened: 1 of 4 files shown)


evwilkin:
@redallen I think this looks good for the UX Writing tables, but will defer to @mcarrano or @rachael-phillips for the design feedback.

That said, I think this work goes towards (but doesn't close) #874 as there are still tables on the site too wide for mobile design, such as the props tables in the component docs:

Screen Shot 2020-10-09 at 3 27 15 PM

mcarrano (changes requested):
I am seeing the same problem that @evwilkin reports on the Propos table for the Alert component. Interestingly we found that other components do not have the same issue. For example, if you reduce the screen size for About modal, the text inside of the table will wrap. So not clear what's different.

mcarrano:
Thanks for pointing that out @redallen . I've always wondered if those examples were more appropriate as demos, anyway. @tlabaj see the thread above. Do you think that the Dynamic Live Region alert examples need to be on this page or should we move those to a demo page?

evwilkin (approved):
LGTM other than specific examples called out, which can be split out into separate issue.

Issue #2165, opened 2020-09-24
Primary-detail demo preview displaying "?" instead of image


mcarrano:
The Primary-detail demo shows "?"

![Screen Shot 2020-09-24 at 10 44 05 AM](https://user-images.githubusercontent.com/6619860/94160798-018d4e80-fe53-11ea-8eb5-eadc97d868ca.png)

Seems like the same issue as was fixed here: https://github.com/patternfly/patternfly-org/issues/2149

PR #2171, opened 2020-09-30
fixed missing images for html demo primary detail and html demo modal top aligned

janwright73:
Hopefully fixes https://github.com/patternfly/patternfly-org/issues/2165.


redallen (approved):
Thanks @janwright73 !

PR #2189, opened 2020-10-08
Adding labels documentation

bmignano:
Adding design guidelines for labels.


mmenestr:
A few comments:
1.  can you add a space break between the "unfilled" and "filled" bullet points so that the images appear underneath and not next to the word. (just press enter one more time between the two)
2. Can you change this text to "**2. Add an icon** to the left of the text for added clarification" to "**Label with icon:** can be used to further clarify the label" just to keep the sentence structure of the bullets the same!
3. Can you make the "when to use...." headings, H3s so they appear in the table of contents (so just ###)
4. In retrospect, the image for clickable labels might be confusing bc it's both a label with an icon, and a dismissible label. Could you change it to a basic label (ie: no icon, no dismissible) that has the click hand on top of it (that you can get from the symbol lib: Pointer > 2 Click)
5. Last thing, most of the images look like they were exported in PNG x1, instead of PNG x2, can you just double check that? they look a little 'blurry' to me! That happens to me sometimes when I export all the art boards at the same time - and it says x2 but actually if you click on the individual art boards, some will say x1

Thank you!!! I know that's a lot

bmignano:
@mmenestr addressed your comments, please take another look!

mmenestr:
Ok last few things
1. For the label features, can you make them singular in the numbered list - Dismissible label and clickable label
2. Can you un-bold the "when to use..." headers
3. The last image still looks like it might be in PNG x1 - but lmk if that's not the case

After that you should be all set!!

bmignano:
Made those updates! The last image says it's 2x but it also was a screenshot I pulled into Sketch. Let me try to get the original image and I will update :)

bmignano:
@mmenestr should be good now!

mmenestr:
Ok looks great, thanks for making all those changes! Merging

mmenestr (approved):

PR #2190, opened 2020-10-09
fix(toc): fix long TOC

redallen:
Removes TOC limit of 25 items, allows duplicate items, and optimizes bundle sizes a bit by using ESM instead of CJS imports. Some quality of life scrollspy improvements have been added as well like recomputing header locations onScroll, not rerendering twice `onClick`, and pushing the hash to the URL `onClick`.

Also fixes content flex bug for smaller viewports.


packages/theme-patternfly-org/templates/mdx.css
```diff
@@ -356,6 +356,7 @@
 }
 
 .ws-mdx-content {
+  flex-grow: 1;
```
mcoker, line 359: Now that you have this, you shouldn't need to set this width:

https://github.com/patternfly/patternfly-org/blob/77ae9a929eeddb6e2abb088eec25dc3e3f7ff65d/packages/theme-patternfly-org/templates/mdx.css#L364-L368

That element already has a `max-width`, and it's a block element, so it will grow to be 100% of it's parent up to 832px

https://github.com/patternfly/patternfly-org/blob/77ae9a929eeddb6e2abb088eec25dc3e3f7ff65d/packages/theme-patternfly-org/templates/mdx.css#L369-L371

jschuler:
everytime i visit a page, leave it, and come back, I get more Props and Css Variables groupings, i.e. on 
https://patternfly-org-pr-2190-v4.surge.sh/v4/layouts/flex
Screen Shot 2020-10-09 at 11 34 06 AM

redallen:
Good catch @jschuler , will fix!

jschuler (approved):

PR #2212, opened 2020-10-19
Feat/404 redesign

### LINKED ISSUE #1607: New 404 page design
Improve the user experience when a user is on a page that doesn't exist in another version and they get a 404 when navigating between them. We don't have a current pattern for this in PF so I would suggest looking at 404 patterns within other RH sites and other design system sites for inspiration!

This is what the current 404 page looks like on patternfly.org (you can get to it if you try to navigate to patternfly.org/v5 for example)
Screen Shot 2020-06-22 at 9 43 12 AM

cc @evwilkin @dgutride

### LINKED ISSUE #2211: Search bar icon wraps
image (2)

redallen:
Closes #1607
Closes #2211


fionayhlin (commented):
Looks great!

Issue #2027, opened 2020-08-10
Create design guidelines for tree view


mmenestr:
Create design guidelines for the tree view pattern.

doruskova:
Doc in progress [here](https://docs.google.com/document/d/1cZkyfz4AodGRg9a7B9Bssa8S1ZogCF0_x9As7CC52FU/edit?usp=sharing)

mmenestr:
Merged

PR #2216, opened 2020-10-21
Adding a tree view guideline

doruskova:
I've added a guideline for a tree view. I removed the animation because the indentation is fixed, but it's using a blue color everywhere. Let me know if there is anything to fix or change. 

Note: Sorry that it took a while when I finished it. 

Solves #2027


packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,92 @@ id: Tree view
 section: components
 ---
 
-We’re working on some guidelines for this component.
-If you have any questions in the meantime, feel free to [get in touch with us.](/get-in-touch)
+A **tree view** is a structure that displays data in a hierarchical view. It can be used in a [dropdown](/components/dropdown), [drawer](/components/drawer), [primary-detail](/demos/primary-detail), [modal](/components/modal), or [wizard](/components/wizard).
+
+A default tree view provides a tree structure with items and an arrow to expand or collapse child items. Every tree view can be extended with these [optional elements](#tree-view-capabilities).
```
mmenestr, line 8: I would change this a little to say "with items and arrows to expand..."
Basically make arrow plural instead of singular, since there will be more than one arrow in a tree view structure!

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,92 @@ id: Tree view
...
+
+## Tree view elements
+
+A default tree view consists of 5 elements:
+
+<img src="./img/tree-view-elements.png" width="332" />
+
+1. **Expand/collapse button:** A button that allows you to expand or collapse child nodes
```
mmenestr, line 16: I actually don't know if I'd use the word "button" here, because the whole area of the item is clickable to expand, not just the arrow. Also isn't it expanding the parent, and not the children? So maybe change it to something like:
"Expand/colla[se: Expand or collapse parents node to reveal child nodes"

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,92 @@ id: Tree view
...
+## Tree view elements
+
+A default tree view consists of 5 elements:
+
+<img src="./img/tree-view-elements.png" width="332" />
+
+1. **Expand/collapse button:** A button that allows you to expand or collapse child nodes 
+2. **Parent node:** A node that contains other items (for example, a folder that contains multiple documents)
+3. **Child node:** A node within the parent node (for example, a document within a folder)
+4. **Leaf node:** The last item in a tree, which doesn't have any child nodes
```
mmenestr, line 19: Does this have to be the last item in a tree? Isn't this just any items that didn't start from inside a folder?

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,92 @@ id: Tree view
...
+5. **Truncation indicator:** A control that automatically shortens long names
```
mmenestr, line 20: I wouldn't call this a "control" I would just say "Longer names that do not fit in the node width will be indicated with ellipses" or something like that

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,92 @@ id: Tree view
...
+<img src="./img/Default-tree-view.png" width="1004" />
+
+Use a tree view when:   
+* Selecting is the main use-case. 
+* You apply filters that need to be structured in a clear hierarchy. 
+* The data is structured into levels.
+* You need to display hierarchies that have more than 2 levels.
+
+Don’t use a tree view when: 
+* The items are not structured hierarchically. Use the [data list](/components/data-list) instead.
```
mmenestr, line 37: or a menu list too? or simple list? Or would you say that's not the same use case as tree view. I just feel like data lists aren't primarily used as selection tools

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
abigaeljamie: To keep it consistent with how the other options are written, we could try, "Truncation indicator: Ellipses indicating truncated text for longer names that don't fit in the node width"

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,92 @@ id: Tree view
...
+* You apply filters that need to be structured in a clear hierarchy. 
+* The data is structured into levels.
+* You need to display hierarchies that have more than 2 levels.
+
+Don’t use a tree view when: 
+* The items are not structured hierarchically. Use the [data list](/components/data-list) instead. 
+* You select 1  item from a small set. Instead, use [select lists](/), [radio buttons](/components/radio), [checkboxes](/components/checkbox), or a [dropdown](/components/dropdown).
+* You need to see an overview of a large amount of data.
+* You need to compare two large sets of data. Use [charts](/charts/about) for a better view. 
+* The hierarchy is too deep and includes hundreds of items.
```
mmenestr, line 41: What would someone use instead then? I feel like I would use a tree view for that but idk hahah

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
mmenestr: Also i would say "optional capabilities" instead of "elements" to match the header for that section

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
abigaeljamie: Makes sense! 👍 I just recommend that we keep each line consistent in terms of part of speech. The other options start with a definition ("A node that..," or "The last item..."), so we'd want to keep the same thing going here instead of giving an action. Perhaps "An option to expand or collapse parent nodes to reveal child nodes" could work?

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,92 @@ id: Tree view
...
+
+### Adding checkboxes to a tree view
+
+Add checkboxes to a tree view if you’re selecting multiple items or applying multiple filters. Selecting should be a primary function of your use-case. If your dataset is small and you select just 1 item, use [select lists](/components/select), [radio buttons](/components/radio), [checkboxes](/components/checkbox), or a [dropdown](/components/dropdown) instead. 
+
+<img src="./img/with-checkboxes.png" width="932" />
+
+### Adding badges to a tree view
+
+Add badges to a tree view if you want to show the number of items inside. Don’t use it for displaying different information. The last item in a tree will not have a badge because it has 0 children.
```
mmenestr, line 66: Why the last item only? I feel like I would just say "Non-folder items should not/can not have badges as they do not contain anything"

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,92 @@ id: Tree view
...
+
+### Adding badges to a tree view
+
+Add badges to a tree view if you want to show the number of items inside. Don’t use it for displaying different information. The last item in a tree will not have a badge because it has 0 children. 
+
+<img src="./img/with-badges.png" width="332" />
+
+### Adding icons to a tree view
+
+Add icons to a tree view if you’re visually representing the type of items in a tree. For example, the folder system. Apply icons on all items in a tree to keep consistency. We don’t recommend to use icons and checkboxes at the same time due to large indentation and loss of space.
```
mmenestr, line 72: Would just tweak the wording on this to say:

"Add icons to a tree view **if you want to** visually represent the type of items in the tree view. For example, **a**folder system. **If you choose to add icons to a tree view,** apply the icons on all items in the tree, to **maintain** consistency". We don't recommend **using** icons and checkboxes at the same time due to **the** large indentation and loss of space **it would result in.**"

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,92 @@ id: Tree view
...
+
+### Adding icons to a tree view
+
+Add icons to a tree view if you’re visually representing the type of items in a tree. For example, the folder system. Apply icons on all items in a tree to keep consistency. We don’t recommend to use icons and checkboxes at the same time due to large indentation and loss of space. 
+
+<img src="./img/with-icons.png" width="332" />
+
+### Adding action items to a tree view
+
+Add action items to a tree view if you’re triggering some type of action above the items. These actions appear on hover to avoid overcrowding the tree view. We recommend using 1 type of action across the whole tree and the maximum of actions for 1 item is 1.
```
mmenestr, line 78: Would reword "We recommend using 1 type of action across the whole tree and **to limit** the maximum of actions for 1 item **to** 1."

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
doruskova: Yeah, it can be confusing. What about this? **The leaf node is a node with no children.**

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
doruskova: It sounds good to me.

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
abigaeljamie: Just a quick note that we'd want to make the description a fragment (not a full sentence) so that it's consistent with the other descriptions in this list. So could we do "Leaf node: A node without children" ?

doruskova: YES! :) Works for me

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
doruskova: You're right. I'll change it. Do you mean this menu list? https://www.patternfly.org/v4/components/menu
I found a simple list.

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
doruskova: Haha, I meant when the dataset is really small (5-10 items) and doesn't have a hierarchy, then you don't need a tree view. You can use just a checkbox or radio button etc. 

I think it's almost the same as the first point.

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
doruskova: Yeah, what about: Items without children can't have badges as they don't contain anything. ??? :D

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
mmenestr: Yes although maybe a menu (which actually is just used in drop downs or select lists) doesn't make sense for the tree view use case-  so nevermind, simple list should be enough, thank you!!

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
mmenestr: I agree with that! That comment was actually about the last bullet though, "The hierarchy is too deep and includes hundreds of items"

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
mmenestr: Yep that works!!

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
doruskova: Oh :D Sometimes you have hundreds or thousands of items and the hierarchy can be really deep. I know that Allie had a use case that included so many items and then we agreed that would be better to use a drill-down menu instead of a tree view. I think it was merged yesterday. So, maybe we can mention it. WDYT?

doruskova:
@mmenestr I've mentioned a drilldown menu in the "Don't use a tree view when" section because it's available in PF components :) And I merged the previous points about the selection from a small set. Did you notice anything else what can I fix/change? Thanks!

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,91 @@ id: Tree view
...
+
+<img src="./img/with-checkboxes.png" width="932" />
+
+### Adding badges to a tree view
+
+Add badges to a tree view if you want to show the number of items inside. Don’t use it for displaying different information. Items without children can't have badges as they don't contain anything.
+
+<img src="./img/with-badges.png" width="332" />
+
+### Adding icons to a tree view
```
mmenestr, line 69: Just a small comment - to be it seems weird to add a folder icon in a child node, because it's not a folder haha. Is it **required*** that if you use icons, you HAVE to use them on everything?

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,91 @@ id: Tree view
...
+Add action items to a tree view if you’re triggering some type of action above the items. These actions appear on hover to avoid overcrowding the tree view. We recommend using 1 type of action across the whole tree and to limit the maximum of actions for 1 item to 1.
+
+<img src="./img/with-action-items.png" width="632" />
+
+## Tree view in a page
+
+Tree views can be included in pages within these components: [dropdown](/components/dropdown), [drawer](/components/drawer), [primary-detail](/demos/primary-detail), [modal](/components/modal), or [wizard](/components/wizard). 
+
+### Tree view in a wizard
+A tree view in a wizard allows users to apply multiple filters or select multiple items that are specific for the current step.
```
mmenestr, line 86: @abigaeljamie should it be "specific **for** the current step" or  "specific **to** current step"?

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,91 @@ id: Tree view
...
+
+<img src="./img/tree-view-wizard.png" width="1500" />
+
+### Tree view in a primary-detail or drawer 
+A tree view in a [primary-detail](/demos/primary-detail) or [drawer](/components/drawer) allows users to switch content based on user's selection from a tree.
```
mmenestr, line 91: I would change this a little to say "allows users to switch content based on **their** selection from a tree"

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
mmenestr: Hey sorry I didn't see your reply here until now^! Yes, I think mentioning it would be helpful, so like "the hierarchy is too deep and includes hundreds of items. Instead, use a drill-down menu." Thank you!! Also left a few other tiny comments, after this it should be good to go :)

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
abigaeljamie: Good catch, @mmenestr ! "Specific to" is the one we should go with 👍

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
doruskova: I agree, but it can be an icon placeholder. I changed the icons to be more relevant now. 
And I think that it should be defined: 
- use it everywhere or you can leave the leaf nodes/items without icons. But, don't combine these leaf items with icons and without. 
1st option: 
with-icons

2nd option:
Default with icons tree view

doruskova: I've changed the image and the sentence: _If you choose to add icons to a tree view, apply the icons on all items in a tree or leave the items without children empty, to maintain consistency._

Don't know if the word "empty" is right.

mmenestr: You could use "iconless" instead of empty! Do you know from a technical stand point if it's possible to have icons on some nodes and not others within the same tree?

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
```diff
@@ -3,5 +3,91 @@ id: Tree view
...
+Add action items to a tree view if you’re triggering some type of action above the items. These actions appear on hover to avoid overcrowding the tree view. We recommend using 1 type of action across the whole tree and to limit the maximum of actions for 1 item to 1.
+
+<img src="./img/with-action-items.png" width="632" />
+
+## Tree view in a page
+
+Tree views can be included in pages within these components: [dropdown](/components/dropdown), [drawer](/components/drawer), [primary-detail](/demos/primary-detail), [modal](/components/modal), or [wizard](/components/wizard). 
+
+### Tree view in a wizard
+A tree view in a wizard allows users to apply multiple filters or select multiple items that are specific to current step.
```
mmenestr, line 86: Now you left out a word haha should be "to **their** current step"!

packages/v4/src/content/design-guidelines/components/tree-view/tree-view.md
doruskova: I'll check it with Michael Coker to be 100% sure.

doruskova: Checked with Michael > it should be fine :)

mmenestr: Great!

mmenestr:
@doruskova I just re-read it and it looks good to go now! Did you want anyone else to review it, or were there any other changes you want to make? Otherwise I will merge

doruskova:
@mmenestr I think it's ok now, you can merge it :)

mmenestr (approved):

PR #2263, opened 2020-11-11
Adding label groups documentation

### LINKED ISSUE #2117: Create label group design guidelines
Write about label groups + link back to general label guidelines.
This doesn't have to be long, and can be similar to the chip group guidelines.

bmignano:
Adding design guidelines for label groups.

Closes #2117


mmenestr:
Just a few comments:
1. Could you bold "label group" in the first sentence and capitalize the L. Just a weird consistency thing we do on all design guidelines haha
2. In the first example image, with the label group in a table, the labels aren't left aligned with the column header like i'd expect. Could you fix that?
3. I think it would be helpful to add a mini header before each example picture like "Label groups in a table" "Label groups in a popover, inside a table" "Label groups in a description list"

Otherwise looks great, thanks for doing this so fast!!

mmenestr (approved):

PR #2287, opened 2020-11-20
Adding switch vs. checkmark guidance

### LINKED ISSUE #2201: Patternfly component guidelines - Switch on/off
As  a designer, I would like more guidance and clarity about the use cases for the switch component. 
- What is the difference between a switch with a checkmark icon and one without? When should the former or latter be used?

https://www.patternfly.org/v4/components/switch

afialkoff:
Closes #2201


packages/v4/src/content/design-guidelines/components/switch/switch.md
```diff
@@ -6,24 +6,26 @@ A **switch** toggles the state of a setting (between on and off). Switches and [
...
+* Depending on your use case, add labels to clarify between toggled settings, such as *On* and *Off*.
 
 
 ### When to use switches versus checkboxes
 
-The main difference between a switch and a checkbox is that a switch changes an option and saves it simultaneously, while checkboxes require a separate action such as pressing a “Submit” or “Save” button to save the selection. 
+A switch changes an option and saves it simultaneously, while checkboxes require a separate action to save the selection, such as pressing a “Submit” or “Save” button.
 
-**Here are some guidelines for when to use a switch versus a checkbox:**
+Use switches with labels by default unless space is limited, then use the checkmark version instead.
```
mmenestr, line 24: Can you move this, and make it a bullet under "General guidelines" instead? It doesn't make sense here because it's not talking about "checkboxes" it's still just talking about the switcher component itself!

mmenestr: And maybe also to clarify, just say "Use switches with labels by default unless space is limited, then use a switch without a label" and link "Switch without a label" to this: https://patternfly-org-pr-2287-v4.surge.sh/v4/components/switch/react#without-label - thank you!!!

mmenestr:
Hey @redallen do you know what issue with build_v4 is here?

redallen:
Just rebase or merge the latest master to fix it. The CI nodes are running out of memory.

packages/v4/src/content/design-guidelines/components/switch/switch.md
```diff
@@ -6,31 +6,32 @@ A **switch** toggles the state of a setting (between on and off). Switches and [
...
+It is assumed that when a user toggles a switch, the change will save automatically without the need for additional action (like clicking a "Save" button).
 
 <img src="./img/switch.png" alt="Examples of switch buttons" width="244"/>
 
 **General guidelines**
 
-*  Add a label in front of your switch to specify what is being switched on or off (for example, wi-fi or bluetooth).
-* Don’t use a switch if the options you’re presenting to the user are anything other than “on” or “off.” Instead, use radio buttons. 
-* Depending on your use case, you may want to add labels to your switches for added clarity, like “on” and “off” to indicate to the user what setting they’ve switched to.
+* Use switches with labels by default unless space is limited, then use a [switch without a label](/components/switch/react#without-label).
```
mmenestr, line 15: I wonder if there's a way we could distinguish between the label that comes with the switch (ie: to say on or off), and the label that signifies what the switch is (ie: the form label) because right now, bullet 1 and 2 are talking about different labels, but they're called the same thing which might be confusing

mmenestr:
Only have the one comment, otherwise looks go to go!

packages/v4/src/content/design-guidelines/components/switch/switch.md
afialkoff: Agreed, the original doc doesn't make these labels seem like different things... Maybe we can rework and condense to distinguish them more clearly:

- Label each switch option to clarify between toggled settings unless space is limited, then use a switch without a label.
- Title your switch with a label that specifies the setting a user turns on or off by toggling it (for example, wi-fi or bluetooth).
- Don’t use a switch if the options you're presenting to a user are anything other than “on” or “off.” Instead, use radio buttons.

mmenestr: @kevinhatchoua this is PR for the issue you filed about clarity around using the switch with label, vs using a switch with a checkmark inside it. Would love to know what you think about this revision^

kevinhatchoua: @mmenestr & @afialkoff Thank you guys for the work on this! I think the text is helpful and speaks to when we should and shouldn't use a switch component with a label. As well as when it's best to defer to radio buttons or checkboxes.

However I was wondering if we can make it clear when to use the switch with the checkmark inside of it? On the same note is there a reason why we don't have a variant with a x inside of it to indicate off?

mmenestr: Maybe we should just use the naming conventions in the switch examples.
There's a "Basic" switch which is the switch with a label to indicate what the switch state is, and then there is the "Switch without label" which is the one with a checkmark inside of it.

So maybe we change the sentence to "Use a basic switch by default unless space is limited, then use a switch without a label. The difference between the basic switch, and the switch without a label is that the basic switch uses an added label to the right of the switch to indicate the state, whereas the switch without a label uses a checkmark inside the switch itself"

Obviously feel free to re-word/shorten that, but you get the idea haha

afialkoff:
Hi @kevinhatchoua and @mmenestr, the latest push addresses your edits and is ready for your review. 🙂

kevinhatchoua:
@afialkoff LGTM!

mmenestr (approved):
Looks great, thanks Alana!

PR #2291, opened 2020-11-24
fix scrollbars, anchor links, page section padding

### LINKED ISSUE #2185: Spacing, anchor links, double scrollbar
Double scrollbar appearing when you have a long TOC. For example, the table component:

Screen Shot 2020-10-07 at 1 25 16 PM

Clicking on these anchor links ("compact table" and "default table") on the [table design guidelines](https://www.patternfly.org/v4/components/table/design-guidelines#compact-and-default-tables) page updates the URL to the correct URL but doesn't anchor me to the right spot on the page. I'm not sure if this is isolated, I only tried these links. Though the URL is correct - if I load that in a new window, it takes me to the right spot on the page.

Screen Shot 2020-10-07 at 1 27 06 PM

The spacing on component/demo pages seems off at different viewports. If you shrink the screen and the vertical nav is hidden, the page section content isn't left-aligned with the hamburger menu - page sections are spaced by default so that they should.

Screen Shot 2020-10-07 at 1 28 14 PM

Also if you keep shrinking the viewport, the component/demo title shifts and is left-aligned with the hamburger menu, but nothing else shifts, and there is extra space added below the tabs. Is there any reason we wouldn't use the spacing that comes from patternfly for the sections & content?

Screen Shot 2020-10-07 at 1 28 21 PM

evwilkin:
Closes #2185 

This PR:
- Hides TOC scrollbar to avoid double scrollbars
- Fixes anchor links on Table design guidelines
- Removes custom PageSection styling in favor of default styling

### DIFF
--- a/packages/theme-patternfly-org/templates/mdx.css
+++ b/packages/theme-patternfly-org/templates/mdx.css
@@ -1,6 +1,6 @@
 #nav-content {
-  background-color: var(--pf-c-page__main-nav--BackgroundColor);
   padding-bottom: 0;
+  padding: var(--pf-global--spacer--xl) var(--pf-global--spacer--xl) 0 var(--pf-global--spacer--xl);
 }
 
 #ws-page-content-router {
@@ -14,7 +14,6 @@
 }
 
 #main-content {
-  flex: 1;
   background-color: var(--pf-global--BackgroundColor--light-200);
 }
 
--- a/packages/theme-patternfly-org/templates/mdx.css
+++ b/packages/theme-patternfly-org/templates/mdx.css
@@ -289,6 +289,10 @@
   border-width: 0;
 }
 
+.ws-mdx-content {
+  flex-grow: 1;
+}
+
 /* Design asked for this. A nice side effect is that the TOC doesn't "jump" between pages */
 .ws-mdx-content-content {
   max-width: 832px;
--- a/packages/theme-patternfly-org/templates/mdx.css
+++ b/packages/theme-patternfly-org/templates/mdx.css
@@ -1,7 +1,3 @@
-#nav-content {
-  padding-bottom: 0;
-}
-
 #ws-page-content-router {
   display: flex;
   flex-direction: column;
--- a/packages/theme-patternfly-org/templates/mdx.css
+++ b/packages/theme-patternfly-org/templates/mdx.css
@@ -1,5 +1,6 @@
 #nav-content {
-  padding: var(--pf-global--spacer--xl) var(--pf-global--spacer--xl) 0 var(--pf-global--spacer--xl);
+  background-color: var(--pf-c-page__main-nav--BackgroundColor);
+  padding-bottom: 0;
 }
 
 #ws-page-content-router {
@@ -12,15 +13,8 @@
   flex-grow: 1;
 }
 
-@media (max-width: 767px) {
-  #ws-page-content-router > section:first-child {
-    padding: 16px;
-  }
-}
-
 #main-content {
   flex: 1;
-  padding: 0;
   background-color: var(--pf-global--BackgroundColor--light-200);
 }
 
@@ -306,7 +300,6 @@
 
 .ws-mdx-content {
   flex-grow: 1;
-  padding: var(--pf-global--spacer--xl);
 }
 
 /* Design asked for this. A nice side effect is that the TOC doesn't "jump" between pages */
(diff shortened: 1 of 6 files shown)


packages/theme-patternfly-org/templates/mdx.js
```diff
@@ -126,10 +126,7 @@ export const MDXTemplate = ({
 
   return (
     <React.Fragment>
-      <PageSection
-        id={isSinglePage ? 'main-content' : 'nav-content'}
-        type={isSinglePage ? 'default' : 'nav'}
```
redallen, line 131: When we put the "HTML/React/Design" tabs in this section I believe it's technically a nav section. It's fine if we don't want to use it, but in that case I'm not sure why the variant exists at all.

packages/theme-patternfly-org/templates/mdx.css
```diff
@@ -12,15 +13,8 @@
...
-@media (max-width: 767px) {
-  #ws-page-content-router > section:first-child {
-    padding: 16px;
-  }
-}
-
 #main-content {
   flex: 1;
-  padding: 0;
   background-color: var(--pf-global--BackgroundColor--light-200);
```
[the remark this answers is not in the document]
evwilkin: Default likely could be used, but this was used to match design mockups for the site

mcoker (changes requested):
This looks good! Noticed a few things though.

* This "data list" link at the bottom of the table design guidelines links to `[...]/data-lists` - should be `data-list`

Screen Shot 2021-01-26 at 1 48 17 PM


* For these `#nav-content` and `:first-child` styles for the first page section, we have `.pf-m-fill` that will cause a section to `flex-grow: 1`, and `.pf-m-no-fill` that will keep it from growing. You could use `.pf-m-fill` here, though I'm not entirely sure why this section is set to grow... Also `.pf-m-light` will give it a white background.

Screen Shot 2021-01-26 at 2 14 57 PM

* `#main-content` could also use `.pf-m-fill`, and by default the page section will have a grey background color. It's a bit different than the one specified here though. I'm all for using the things built in to patternfly, but if that grey needs to be different then that's fine.

Screen Shot 2021-01-26 at 2 15 25 PM

* This footer section could use `.pf-m-no-fill`, and potentially the default page section padding instead of what's specified here. Then the padding will change as the rest of the page sections' padding changes, which may or may not be what you want. Seems like it would be fine using the default padding though.

Screen Shot 2021-01-26 at 2 16 10 PM

evwilkin:
@mcoker FYI I took in your suggestions above but also found a better way to structure some of the components used that made more sense, so now the default spacing should be used that majority of the time rather than custom spacing overrides.  Note that the alignment falls into 3 categories which helped me check different use cases - Component section pages (w/menu below title), Guidelines section (title only), and Training pages.

packages/theme-patternfly-org/templates/mdx.js
```diff
@@ -69,7 +69,7 @@ const MDXChildTemplate = ({
       {toc.length > 1 && (
         <TableOfContents items={toc} />
       )}
-      <div className={katacodaLayout? "ws-mdx-content-katacoda" : "ws-mdx-content"}>
+      <div className={katacodaLayout? "ws-mdx-content-katacoda" : "ws-mdx-content pf-m-fill"}>
```
mcoker, line 72: > `pf-m-fill`

This works on some component/layout elements (like `.pf-c-page__main-section`, or a child of `.pf-l-split`), but doesn't appear to be doing anything here, so not sure if you need it.

mcoker:
@mcarrano @mceledonia I'm curious what you think of how the tabs at the top of the page are aligned. It looks like the left side of the first tab text is aligned with the page gutter, though that's not how it's designed to work in the code anyways. We don't have a page section to wrap the tabs component to align it properly. Looking at our upcoming tabs demo design https://marvelapp.com/prototype/5fba0cj/screen/73740947, it looks like the left edge of the tab underline should be aligned with the left edge of the page gutter.

Here's what it looks like now:

#### big screen
Screen Shot 2021-02-11 at 10 31 29 AM

#### small screen
Screen Shot 2021-02-11 at 10 31 38 AM

Here's what it looks like if I put it in a normal page section and remove the top and bottom padding (`.pf-c-page__main-section pf-m-light pf-u-pb-0 pf-u-pt-0`), which keeps the left/right gutter aligned with the other page sections, and allows the space between the title and tabs to match the page section padding which is 24px on large screens, 16px on small screens.

#### big screen
Screen Shot 2021-02-11 at 10 39 01 AM


#### small screen
Screen Shot 2021-02-11 at 10 39 10 AM

wdyt?

mcoker (commented):
Looks great! Left a comment about the tabs, would be good to get some design feedback about that. Also maybe in an upcoming sprint we can look at using the new sidebar layout to manage the TOC?

mcarrano:
I agree with @mcoker . The new tab alignment looks odd and I think it's better if the left edge of the highlight lines up with the edge of the content.

mcoker (approved):
👍👍

mcarrano:
Looks much better !

PR #2306, opened 2020-12-03
fix(docs): add sidenav shadow on mobile, close sidenav on NavItem click on mobile

### LINKED ISSUE #2205: Mobile collapsible sidebar behavior/design
It seems like when you're on mobile, and the sidebar overlays the content area when expanded, if the sidebar is open and a user clicks on a link in the sidebar, that the sidebar should automatically close. Otherwise a user has to click outside of the sidebar somewhere to close it. The only time I can see that a user might want it to stay open is if they click on an item by accident and want to leave it open so they can click on the correct link. I imagine 80% of users on mobile open the sidebar to go somewhere, then want the sidebar to close so they can see the page content.

Also there is no separator between the sidebar and main page content on org as the sidebar shadow is disabled. Seems like we should retain the shadow on mobile.

What it looks like now

Screen Shot 2020-10-13 at 1 13 27 PM

What it looks like with the shadow

Screen Shot 2020-10-13 at 1 13 33 PM

evwilkin:
Closes #2205 

This PR:
- Adds dropshadow to sidenav on mobile
- Closes sidenav on mobile when user clicks a page link in the sidenav by passing `onNavToggle` to `Sidenav` component & triggering in `onClick` in `navItem`

### DIFF
--- a/packages/theme-patternfly-org/components/sideNav/sideNav.js
+++ b/packages/theme-patternfly-org/components/sideNav/sideNav.js
@@ -1,28 +1,32 @@
 import React from 'react';
 import { Link } from '../link/link';
-import { Nav, NavList, NavExpandable, capitalize } from '@patternfly/react-core';
+import { Nav, NavList, NavExpandable, PageContextConsumer, capitalize } from '@patternfly/react-core';
 import { css } from '@patternfly/react-styles';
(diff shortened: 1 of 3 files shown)


redallen (approved):
Works well!

Issue #2209, opened 2020-10-15
Apply new feedback to PatternFly's "Numerics" page


abigaeljamie:
LINK: https://www.patternfly.org/v4/ux-writing/numerics

Feedback from Catherine Tomasko: https://docs.google.com/document/d/1Ax7fmUYjY7Jh15qmbxwKFrcEogvlN0yuNrqOIULIkNk/edit#heading=h.8j1vzujkxaah

abigaeljamie:
Catherine went over her feedback at the last UX writing style guide sync with Alana, Ingrid, Margot, Matt, and Abi. She'll be editing this doc based on our convo, then it will be ready to add to PatternFly. I followed up with Catherine today to see if she's done editing it.

abigaeljamie:
On hold until we hear back from Catherine. Followed up with her last week.

abigaeljamie:
Catherine finished edits. This one is ready for updating.

abigaeljamie:
Alana to do PR by 12/11

PR #2316, opened 2020-12-11
Applying feedback to Numerics page

afialkoff:
Addresses #2209: 

Catherine's feedback deletes some of Pat's feedback concerning ISO standards for international currency. @abigaeljamie, are we planning on removing that feedback from the page? I left it under the **Currency** heading just in case, and we can follow up next week.

### DIFF
--- a/packages/v4/src/content/design-guidelines/content/numerics.md
+++ b/packages/v4/src/content/design-guidelines/content/numerics.md
@@ -26,7 +26,7 @@ PatternFly date and time formats follow the American standard. When localizing, 
 
 ## Absolute or relative time
 
-Using **absolute** or **relative** timestamps depends on the context. To represent the exact date and time that an event occurred, use an absolute timestamp.
+Using **absolute** or **relative** timestamps depends on the context. To represent the exact date and time that an event occurred, use an absolute timestamp following examples provided in the previous [table of date and time formats](#date-and-time-formats).
 
 To represent how long ago an event occurred, use a relative timestamp. When reporting relative time, follow these examples:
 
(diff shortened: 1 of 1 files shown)


abigaeljamie:
Hey, @mmenestr ! This one is ready for your review :)

packages/v4/src/content/design-guidelines/content/numerics.md
```diff
@@ -3,55 +3,32 @@ id: Numerics
...
+| Month | Write out the full name of the month. If space is limited, use the month’s 3-letter abbreviation. <ul><li>Jan</li><li>Feb</li><li>Mar</li><li>Apr</li><li>May</li><li>Jun</li><li>Jul</li><li>Aug</li><li>Sep</li><li>Oct</li><li>Nov</li><li>Dec</li></ul> | September 17, 2020 <br /><br />Sep 17, 2020 |
+| Duration | HH:MM:SS or HH:MM | 03:15:30 <br /><br />03:15<br /><br />00:15 |
+</div>
 
-Using **absolute** or **relative** timestamps will depend on the context. If your user is primarily interested in understanding the exact date and time that an event occurred, use an absolute timestamp.
+## Absolute or relative time
 
-If users are interested in duration, the amount of time it takes for an event to occur, or the amount of time between events, use an absolute timestamp formatted as HH:MM:SS.
+Using **absolute** or **relative** timestamps depends on the context. To represent the exact date and time that an event occurred, use an absolute timestamp.
```
mmenestr, line 29: The only comment I have is that you mention absolute and relative time here, but then only have examples of the relative time stamps. I'm assuming the table above this section is what you're referring to as the absolute time stamps, but maybe I'd be helpful to mention that so that people have a clear reference point of what absolute means?

mmenestr (changes requested):
Just left one comment, otherwise looks great!!

packages/v4/src/content/design-guidelines/content/numerics.md
afialkoff: Ooh, thanks for catching this! I'll add some context here so that readers will know absolute timestamps examples are shown in that first table.

mmenestr (approved):

PR #2334, opened 2021-01-04
feat(docgen): add @propType and fix @hide

### LINKED ISSUE #2332: Docs: @propType and @hide annotations do not function
When I add `@deprecated` to a component prop, I can see it output in the example docs. However, I don't see a similar result with the `@propType` and `@hide` functionality added via https://github.com/patternfly/patternfly-org/pull/1671.

I'm expecting to see a different property type, for Chart props below, which is shown by the area chart example.

```
export interface ChartProps extends VictoryChartProps {
  /**
   * The animate prop specifies props for VictoryAnimation to use.
   * The animate prop should also be used to specify enter and exit
   * transition configurations with the `onExit` and `onEnter` namespaces respectively.
   *
   * @example {duration: 500, onExit: () => {}, onEnter: {duration: 500, before: () => ({y: 0})})}
   *
   * @propType boolean || object <------------
   */
  animate?: boolean | AnimatePropTypeInterface;
```

Is there a trick to get this working locally? I can see `@deprecated` when I rebuild the docs, but `@propType` and `@hide` do not appear to do anything.

Screen Shot 2020-12-28 at 12 40 14 PM

redallen:
Closes #2332

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ jobs:
         key: *js_deps_cache_key
   build_v3:
     docker:
-    - image: circleci/ruby:latest-node
+    - image: circleci/ruby:2-node
     steps:
     - checkout
     - restore_cache:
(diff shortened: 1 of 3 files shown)


dlabrecq (approved):
Thanks, Zack!

PR #2343, opened 2021-01-07
Move contribute/design-guidelines instructions to repo

mcarrano:
This PR removes the contribute/design-guidelines page from the website and moves content to a page in the patternfly-org repo here: https://github.com/patternfly/patternfly-org/wiki/Contributing-to-patternfly-org-for-designers and links this page from https://www.patternfly.org/v4/contribute/design

Some reasons for this change are the following:

* It is customary to include detailed contribution procedures inside of the repo that they pertain to rather than including on a website. This is consistent with how developer contribution guidance is presented. This simplifies the website and moves the information closer to the place where contributors will work.

* We already had some of this information living in the patternfly-org/wiki and this brings information up to date and centralizes it in one place. The Wiki can be easily updated as things change without affecting the website.


mcarrano:
@mmenestr This is a change that @evwilkin and I wanted to make to simplify the Contribute section of the website by moving the detailed contribution guidelines for design documentation from the website to the Wiki that lives inside the patternfly-org repo. This was mostly just a process of moving the existing content from the page. I also added some information that was missing about the required front matter in the design.md files. But otherwise the existing content was kept in-tack. Please let me know if you see any formatting issues I may have missed in moving the markdown.

One thing I noticed was that you are explicitly called out as the contact person for design guidelines questions, issues, or help. While you are definitely the current go-to person for this, I wonder if we should be directly exposing peoples names or email addresses in a page like this. I don't think we do it elsewhere. @evwilkin @redallen I'm interested in your thoughts on this. What is or should be our common practice on this? I'm not sure what to suggest as an alternative solution, but I think it would be best to not tie this to a specific individual. Thoughts?

redallen:
Emails in plaintext tend to get crawled and then spammed. I believe @gdoyle1 and I removed her email from a previous page on patternfly.org for this reason. That's pretty much the only risk, though. If the email address owners are willing to receive more unwanted mail it's safe.

redallen (approved):

redallen (approved):
I love this change! I'm all for more wiki content.

evwilkin (approved):
LGTM

evwilkin:
@mcarrano I'll defer to @mmenestr for how to include her details in the page copy, I'm not sure an alternate way to add an alias or something similar to attach to PR reviews that would be more generic and could be reassigned to anyone who would take on that role in the future - all I can think of is perhaps to direct to the Forum or Slack channel to notify us which would require an extra triage step to get it assigned correctly.

mmenestr:
Hmm right now, I'm fine with just having my github username on there. As long as my full name isn't included it seems fine to me! I also haven't had any issues with my email being on there in the past

redallen:
Sounds good @mmenestr . Once you fix the merge conflicts @mcarrano we can take this in.

redallen (approved):

Issue #2329, opened 2020-12-18
Add additional guidance to Labels doc


mmenestr:
Add some guidelines around using to the Label design guidelines following this doc https://docs.google.com/document/d/1K2pBe_hKevycrMQFpc8MuNnlxcPC41rmRJ74cxRnauw/edit?disco=AAAALPeUN9I&ts=5fdbe011&usp_dm=false

Maybe phrase it as suggestions:
- PatternFly recommends reserving the red, yellow and green label colors for indicating status etc....
- PatternFly recommends using outlined labels for clickable labels, and filled for un-clickable labels...

mcarrano:
See related discussion here: https://github.com/patternfly/patternfly-design/issues/971

PR #2347, opened 2021-01-08
Updating label guidelines

bmignano:
Updating label guidelines to include recommendations around color use and filled vs. unfilled labels.

Closes https://github.com/patternfly/patternfly-org/issues/2329


packages/v4/src/content/design-guidelines/components/label/label.md
```diff
@@ -17,9 +17,9 @@ Labels are easily customized to fit different use cases. There are two main type
 
 <img src="./img/filled-labels.png" alt="Filled labels" width="646"/>
 
-Use the filled option to add more visual prominence to a label.
+Use the filled option to add more visual prominence to a label. If your use case requires a mix of clickable and non-clickable labels, consider using unfilled for clickable and filled for non-clickable, or vice versa. Whatever convention you choose, be sure to maintain consistency through your UI.
```
mcarrano, line 20: Do we need to include "or vice versa" here? It makes this seem like a non-recommendation.

packages/v4/src/content/design-guidelines/components/label/label.md
```diff
@@ -46,7 +46,7 @@ Optionally add an icon to your label in cases where additional visual informatio
```
mcarrano, line 46: I wonder whether it would be helpful for this example to show same color labels with different icons used to distinguish them. At least that's what I expected after reading the comment above.

mcarrano (commented):
Looks great @bmignano . There were just two minor comments for you to consider.

packages/v4/src/content/design-guidelines/components/label/label.md
bmignano: Agreed, I've removed that piece.

packages/v4/src/content/design-guidelines/components/label/label.md
bmignano: That's fair. Does it make sense to include both scenarios? I can definitely mock up an example with the same color labels and different icons.

mcarrano: Sure. I think you could have two examples.

bmignano: Added!

mcarrano: @bmignano I don't think that this is what you intended, but here's how this looks when I preview the site:

Screen Shot 2021-01-12 at 4 56 42 PM

Maybe place this sentance, "You can also use them to distinguish labels of the same color if needed." between the two image tags and that will make them stack.

bmignano: Ah, thanks for catching that @mcarrano ! Updated with your suggestion.

mcarrano (approved):
Looks great @bmignano . 

@mmenestr please also take a look and merge if ready. Note that we are in the process of making visual updates to the label designs, so the images will need to be updated in the near future, but the guidance should not need to change.

mmenestr (approved):
Looks great Brie, thanks for making these updates!

Issue #2370, opened 2021-01-26
Home page "shimmy" when typing into the search field on Chrome or MS Edge browsers


mcarrano:
When typing into the search field in the header of the masthead, the display "shimmies" left to right as the vertical scroll bar shows/hides. This effect only appears on Chrome based browsers, not Firefox or Safari.

Found by Matt Reid. Video to be added.

mattreid:
From some quick browser testing, this seems to only affect Chromium based browsers (I didn't try it in Brave, but I suspect we'd see the same shimmy). There seems to be something going on between the placement of the search plugin overlay and whatever CSS/styling is happening when mousing over the icons in that area. If you get the cursor in the right spot, it will continuously trigger this behavior. I'm not sure if OS impacts this at all, I'm on a Mac.

It might be related to where we place the overlay, as when that appears, no matter how wide my window is (the videos were taken ~1600px), it adds a horizontal scrollbar to my browser because the search overlay is placed slightly off the active viewport.

I made two quick videos demonstrating this.
https://user-images.githubusercontent.com/1242413/105886626-9dcd8a80-5fd8-11eb-9950-a2490fc955ed.mov

https://user-images.githubusercontent.com/1242413/105886633-9efeb780-5fd8-11eb-974f-4f3f47c2c621.mov

mcarrano:
Thanks for posting the additional info @mattreid .

@redallen @evwilkin can you take a look at this?

redallen:
I can't reproduce this but I've been thinking of making a custom search for patternfly.org that should work better than Algolia's and not have this problem. Interfacing with both their JS/CSS that ties to the search box as well as their backend that (still) doesn't crawl our content very well has been painful in my experience.

I can pretty quickly put together some JS for it. For example searching "Empty state" could reveal the following 8 pages:
```json
[
  '/components/empty-state/design-guidelines': {
    id: "Empty state",
    title: "Empty state",
    toc: [{"text":"Usage"},{"text":"Elements"},{"text":"Design guidelines"},[{"text":"Icons"},{"text":"Title and body"},{"text":"Primary and secondary buttons"},{"text":"Alignment"},{"text":"Toolbars or filters in empty states"}],{"text":"Use case examples"},[{"text":"First-use or get started"},{"text":"No data to show"},{"text":"Required configuration"},{"text":"User not entitled or does not have access"},{"text":"Back-end failure"},{"text":"Success"},{"text":"Addition or creation"}]],
    section: "components",
[...]

PR #2378, opened 2021-01-29
Fix search results overflow & search button on mobile

evwilkin:
Towards #2370 

This PR:
- reduces min-width of the search results container by 20px which prevents it from overflowing the right edge of the screen on all sizes down until it is hidden at 670px viewport width.
- hides the search icon/close icon toggle at 670px to match search input responsive behavior. Currently these icons are visible below 671px but with the search input hidden they have no visible effect.


evwilkin:
@mattreid I wasn't able to recreate the shimmy seen in #2370 in Chrome or Edge but do see the overflow and am hoping this should help fix it - can you take a look and let me know?

mattreid:
@evwilkin Seems like it fixes it! I still see it on Chrome and Edge in prod pf4 using private windows, but the surge link I can't recreate using the same steps. 👍 FF still behaves normally, so no regression there.

redallen (approved):

PR #2379, opened 2021-01-29
Add tooltip labels & hide Type column

### LINKED ISSUE #2223: Labels for icons
All semantic icons should have a label a tooltip. Working with CCS and UXD to craft one-word labels for common/general icons. Labels will be added to PatternFly's icon page: https://www.patternfly.org/v4/guidelines/icons/ 

Brainstorm doc: https://docs.google.com/document/d/1qUvB0vcF_Eh-tqB0B1EYyrd_oZ06nQg3-2df8g2QVxk/edit

evwilkin:
Closes #2223 

This PR:
- Adds new `Tooltip labels` column
- Adds tooltip labels to a handful of icons
- Hides the `Type` column, but still allows it to be searchable

### DIFF
--- a/packages/v4/src/content/design-guidelines/styles/icons/IconsTable.js
+++ b/packages/v4/src/content/design-guidelines/styles/icons/IconsTable.js
@@ -38,7 +38,8 @@ export class IconsTable extends React.Component {
       'Icon',
       { title: 'Name', transforms: [sortable], props: { className: css(styles.modifiers.fitContent)} },
       'Style',
-      { title: 'Type', transforms: [sortable] },
+      'Type',
+      'Tooltip label', 
       { title: 'React', props: { className: css(styles.modifiers.fitContent)} },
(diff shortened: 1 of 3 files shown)


abigaeljamie (changes requested):
Thank you so much, Evan! 

Is there any way to make some space between the "Tooltip" column and the "React" column? They look a little squishy. Other than that, this looks great :)

evwilkin:
@abigaeljamie the padding between the columns is consistent, though some columns are more spread out as they're set to expand for their content to fit & the other columns are set to size down & wrap lines to fit. 

Rather than overriding the table column styling, how about if we move the Tooltip label column all the way to the right - this way the column title doesn't bump up against the column next to it, and as most of the rows for this column are empty it may make sense to have it be the last one listed:

Screen Shot 2021-02-01 at 4 16 22 PM

abigaeljamie:
Thanks, Evan!

abigaeljamie (approved):

PR #2412, opened 2021-03-17
add pf-m-wrap to MD th elements

redallen:
Makes all table headers wrap per discussion with @andybraren 

cc @mcarrano

### DIFF
--- a/packages/theme-patternfly-org/scripts/md/styled-tags.js
+++ b/packages/theme-patternfly-org/scripts/md/styled-tags.js
@@ -38,6 +38,7 @@ function styledTags() {
               tr.children
                 .filter(child => child.tagName === 'th')
                 .forEach(th => {
+                  th.properties.className = th.properties.className || '';
                   th.properties.className += ' pf-m-wrap';
                   th.properties.role = 'columnheader';
                   th.properties.scope = 'col';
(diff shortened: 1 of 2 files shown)


packages/theme-patternfly-org/scripts/md/styled-tags.js
```diff
@@ -38,6 +38,8 @@ function styledTags() {
               tr.children
                 .filter(child => child.tagName === 'th')
                 .forEach(th => {
+                  th.properties.className = th.properties.className || '';
+                  th.properties.className += ' pf-m-wrap';
```
andybraren, line 42:
```suggestion
                  th.properties.classList.add('pf-m-wrap');
```

My JS game is weak, but would something like this work?

andybraren (approved):
Ah yep, fixed. The preceding space that can occur in the `class` attribute is a little ugly, but if you're cool with merging then I am.

```
<th class=" pf-m-wrap" role="columnheader" scope="col">
```

mcarrano:
This is great @redallen !  So does this mean that tables written in Markdown will now wrap properly? No more need to code these directly in HTML?

cc @mmenestr

andybraren:
@mcarrano They'll no longer be clipped and hopefully wrap better in general, at least. See some discussion & screenshots [here](https://patternfly.slack.com/archives/CB9KCDK4M/p1615997027005500?thread_ts=1615991480.000800&cid=CB9KCDK4M).

packages/theme-patternfly-org/scripts/md/styled-tags.js
redallen: Nope. This script is converting AST nodes from MD into JSX.

The leading space doesn't matter except for aesthetics so I'm okay with it.

PR #2413, opened 2021-03-17
fix(buble): perfect JSXText regex?

redallen:
Makes
```jsx
<div>
  1 <span>2</span> 3
</div>
```
into
```js
return React.createElement("div",{},"1 ",React.createElement("span",{},"2")," 3");
```
just like [the Babel compiler!](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=DwEwlgbgfAUABHAjHYBnADgQwHZQEzAD0GOUcAzDEeNEA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=react&prettier=false&targets=&version=7.13.11&externalPlugins=%40babel%2Fplugin-proposal-class-properties%407.12.1)

Related: #2284

### DIFF
--- a/packages/theme-patternfly-org/helpers/buble.js
+++ b/packages/theme-patternfly-org/helpers/buble.js
@@ -58,7 +58,7 @@ const generator = Object.assign({}, baseGenerator, {
     }
     if (node.value) {
       state.write(',"');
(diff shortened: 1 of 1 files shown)


mcoker (approved):
🥳👍

PR #2419, opened 2021-03-19
Add show and hide password to login page

### LINKED ISSUE #2349: Add show/hide password to Login page documentation
Update Login page design docs to include show/hide password guidance.

Good contacts: Michael C, and Matt Carrano

cc @mmenestr

doruskova:
I've added the show/hide icon for password visibility to the Login page. I mentioned this feature in the Elements section and updated the image for it. @mmenestr Should it be more specific? I didn't want to make it too long.

Adding a sketch file with an updated image
[show and hide password.zip](https://github.com/patternfly/patternfly-org/files/6172509/show.and.hide.password.zip)

closes #2349


mmenestr:
I think this is probably fine. @mcarrano what do you think? Do you think adding it as an element in the call outs is enough, or would you expect to have more detail around it? I would think it's pretty self explanatory

mcarrano:
Yes. I think it's fine to just note here as an optional element like this.

mmenestr (approved):

PR #2423, opened 2021-03-24
Adding a 404 page resource to the UX writing style guide

### LINKED ISSUE #2404: Create a "404 page" page in the UX writing style guide
[UX writing style guide link ](https://www.patternfly.org/v4/ux-writing/about)

Add a new page, "404 pages," and include microcopy guidance.

afialkoff:
Closes #2404


afialkoff:
Hey @mmenestr! You already reviewed this content through Google Docs but now it's ready for a PR review!

I see one main issue in the preview: "404 pages" precedes "About" in the *UX writing* nested navigation, which means this page would appear before the secton that introduces the UX writing style guide... Do you think there might be a workaround for this? We could always rename the page to something like "Writing 404 pages" to sidestep that alphabetical ordering.

mmenestr:
Hey @afialkoff looks good! My only comment would be that since this page is named "404" it is the first page that comes up under the UX writing section which strikes me as odd. I'm also now wondering whether this should just live under the "Error messages" section that already exists, since a 404 is technically an error type... what do you think? You could split the page up into "General errors" and "404 errors"

afialkoff:
> Hey @afialkoff looks good! My only comment would be that since this page is named "404" it is the first page that comes up under the UX writing section which strikes me as odd. I'm also now wondering whether this should just live under the "Error messages" section that already exists, since a 404 is technically an error type... what do you think? You could split the page up into "General errors" and "404 errors"

Agreed, this page shouldn't be listed in the nav before the "About" page that introduces the UX writing style guide. Originally, we wanted to surface 404 pages as a separate page in the style guide since they're often written as their own design/content pattern. I'm definitely open to adding this content as a new "404 errors" section on the "Error messages" page if we think that makes the most sense. 👍

@abigaeljamie, would love to hear your thoughts on this!

abigaeljamie:
Good catch, @mmenestr and @afialkoff ! 

What if we rename the page to "HTTP 404 page" or "Standard 404 page" so that it doesn't come before the "About" page? I feel like folks might see a 404 page as more of a landing page than an error message, so I worry that the 404 content guidance would get overlooked if we added it to the "Error messages" page.

afialkoff:
I love those new title ideas, @abigaeljamie! @mmenestr, do you have a preference between "HTTP 404 page" or "Standard 404 page"?

mmenestr:
I don't have a preference!

abigaeljamie:
Thanks, Margot! 

So the choice is entirely yours, Alana!

afialkoff:
Thanks for your input, everyone! 🙌 I think more people would recognize "HTTP 404 page" from a quick nav scan, so let's go with that. I'll update accordingly!

afialkoff:
@mmenestr, ignore that second commit — my terminal did something wonky and made me push twice! 🤯 The page title and file name change to "HTTP 404 pages" both appear in that first commit, "Updating title to HTTP 404 pages."

packages/v4/src/content/design-guidelines/content/HTTP-404-pages.md
```diff
@@ -0,0 +1,100 @@
...
+## Writing 404 page content
+Consider your 404 page as a flight path instead of a dead end. Use clear, informative microcopy to point users in the right direction.
+
+**Use understandable language.** Skip technical jargon by writing your 404 error message in plain and specific terms.
+
+<div class="ws-content-table">
+
+| **Don't**                      | **Do**                  |
+|:-------------------------------:|:--------------------------:|
+| Error 404: Not found | 404: THat page no longer exists|
```
mmenestr, line 31: Oops, jut noticed a typo, should be "That" not "THat"

packages/v4/src/content/design-guidelines/content/HTTP-404-pages.md
```diff
@@ -0,0 +1,100 @@
...
+
+<br />
+
+**Turn error into opportunity.** Always provide a link back to your site’s home page, and consider including supplemental microcopy to encourage users to explore options beyond just going back to where they came from.
+
+<div class="ws-content-table">
+
+| **Don't**                      | **Do**                  |
+|:-------------------------------:|:--------------------------:|
+| That page doesn't exist. | Another page might have what you need, so try searching PatternFly. |
```
mmenestr, line 70: Also last thing - wondering why there are periods at the end of these examples, and not in any of the others

packages/v4/src/content/design-guidelines/content/HTTP-404-pages.md
afialkoff: Yikes! I'll edit that out — good catch.

packages/v4/src/content/design-guidelines/content/HTTP-404-pages.md
afialkoff: I figured we should probably punctuate the examples in this table for two reasons:

1. That second line is punctuated on our actual 404 page we link to at the bottom of the page.
2. This table provides examples of next steps/body copy vs. the heading examples provided in previous ones. Typically, headings aren't punctuated, but next steps are. Should call out guidance for those punctuation best practices, maybe in the elements list? 

If so, we could then rephrase "supplemental microcopy" to "supplemental next steps" in the lines that introduce this table so that readers know they're reading an example of punctuated next step copy:

"**Turn error into opportunity.** Always provide a link back to your site’s home page, and consider including supplemental next steps to encourage users to explore options beyond just going back to where they came from."

What do you think?

mmenestr: Ok that's what I figured! I think it just wasn't super clear to me which ones were heading suggestions and which ones were body suggestions, because the table right above this one seems like body suggestions, but the fact that they don't have periods make me think they're actually heading suggestions?

afialkoff: Yes, the previous table contains heading suggestions! I think we can clarify this by mentioning whether the examples are headings or lines of body copy when we introduce them. I can tinker around with this to make sure we communicate that difference in a way that makes sense but also isn't too redundant. 👍

afialkoff:
Hey @mmenestr! I added some additional punctuation guidance to the initial "Elements" list and rephrased each recommendation so that readers should know exactly which type of content each table demonstrates. I also updated the example for "Channel your brand voice" so that the last two examples show headings and next steps together. 

Let me know what you think!

packages/v4/src/content/design-guidelines/content/HTTP-404-pages.md
abigaeljamie: @afialkoff , are you able to make the left side "Do" and the right side "Don't" ? That would align with our other guidelines in the style guide (example: https://www.patternfly.org/v4/ux-writing/brand-voice-and-tone).

afialkoff: Absolutely, I'll update that now! 🙂 👍

mmenestr (approved):

PR #2430, opened 2021-03-31
chore(release): 2021.04

nicolethoen:


packages/theme-patternfly-org/versions.json
```diff
@@ -1,9 +1,27 @@
 {
   "Releases": [
+    {
+      "name": "2021.04",
+      "date": "2021-03-31",
```
redallen, line 5: Can you change this date to today?

redallen (approved):

PR #2433, opened 2021-04-05
2021.04 Release highlights

mcarrano:
Adds release highlights for the 2021.04 release.


nicolethoen (dismissed):
I notice that the link description after the dash for the 'what we are working on' sections, you are inconsistent in your use of uppercase at the beginning of the descriptions. I'm not sure if that matters based on your intended use of this doc.

nicolethoen (approved):

PR #2452, opened 2021-04-14
added sentence on clickable badges

### LINKED ISSUE #2107: Add to badges design guidelines
https://www.patternfly.org/v4/components/badge/design-guidelines

This should be a short one, but we would like to add a guideline to say that "Badges shouldn't be used if selectable tags are needed (instead use labels)" 

I think this fits in a future issue we already have filed

mmenestr:
Closes #2107 

@mcarrano just added one sentence authored by @lclay2 on clickable badges


mcarrano (approved):
Looks good to me. I will merge. Thanks @mmenestr @lclay2

PR #2457, opened 2021-04-19
Adding design guidelines for application launcher

### LINKED ISSUE #2360: Add guidelines for "Application Launcher" component
Create guidelines for the application launcher component: https://www.patternfly.org/v4/components/application-launcher/design-guidelines

Ideas for what to include:
- Usage
- What it could contain
- Where it should be placed in the UI

bmignano:
Adding design guidelines for application launcher.

Closes #2360


mcarrano (commented):
Looks good @bmignano . But should we add something about using the Favorities feature?

bmignano:
@mcarrano I looked for PatternFly favorites guidelines but couldn't find anything to link to. Do you think it maybe just calls for a sentence or two explaining that when an item is favorited, the favorites section appears at the top and the item is repeated there as well?

mcarrano:
@bmignano  yes, I think so. It would just serve to make people aware that this feature exists and perhaps when you should use it. For example, this only seems to make sense IMO if you have a long list of items in the menu and you want to pin some to the top.

bmignano:
@mcarrano I've added additional details around the favoriting mechanism, PTAL!

packages/v4/src/content/design-guidelines/components/application-launcher/application-launcher.md
```diff
@@ -2,4 +2,26 @@
...
+Application launchers are usually found within the masthead of a UI. They are commonly used when you want to allow users to launch multiple applications from a common location. Use the external link icon to indicate when a link opens a new window. You can also use icons in the list.
+
+For longer lists, you can optionally add the ability to search the list and favorite items. Favoriting an item using the star icon allows users to pin items to the top of the menu in a separate Favorites section. The item will appear in the list twice: once in the Favorites section and once in its original location.
+
+
+## Examples
+**Application launcher with external links**
+<img src="./img/example1.png" alt="Example of the application launcher component" width="1680"/>
+
+**Application launcer with favorites and search**
```
mmenestr, line 26: There's a typo here with "launcer" haha

mcarrano:
Updated text is fine @bmignano . Should we also update the image under elements to include the Favorite items and mark them as optional? Our standard practice is to typically show all features with separate callouts and then mark those that are optional. Is that what you would also expect @mmenestr ?

mmenestr:
That might be a good idea actually.. like have both zoomed in views side by side, and have the third call out be the favorite icon?

mcarrano (approved):
Looks great. Thanks @bmignano !

mmenestr (approved):

PR #2470, opened 2021-04-30
Updates about beta components and promotion

nicolethoen:
I made an update to the tag added to Beta components for a little clarity. I am open to suggestions, and am happy to revert the change if people would prefer to update our information about Beta components a different way.

I also added a markdown file regarding the promotion candidates for the June release. This is what I would present at the community meeting and future UXD Dev meeting.


packages/theme-patternfly-org/templates/mdx.js
```diff
@@ -53,7 +53,8 @@ const MDXChildTemplate = ({
       )}
       {beta && (
         <InlineAlert title="Beta feature">
-          This Beta component is currently under review, so please join in and give us your feedback on the <a href="https://forum.patternfly.org/">PatternFly forum</a>.
+          This Beta component is currently under review. It is available for use in product with the caveat that user feedback could result in future breaking changes.
```
mcarrano, line 56: While what you say in the second sentence is true, I worry that reminding people of this fact will make it less likely for them to use the beta component. In the past, I had gotten feedback that developers were reluctant to user a beta component for fear of breaking changes, but then we don't get feedback.

mcarrano (commented):
I like the direction @nicolethoen . Just had one comment about the proposed change to the beta tag,

As for the markdown doc, I agree this will be a good tool for use at community meetings to gether feedback. I only wish there was a way to directly collect feedback from the doc itself. What would you think about opening a GitHub issue to collect feedback. Folks could comment there. Not sure if we need one issue per item or if this could be a global issue. We should discuss.

packages/theme-patternfly-org/templates/mdx.js
evwilkin: Agree with @mcarrano here - we want to encourage adoption with an understanding that they may continue to be improved upon.

nicolethoen: hm... i just want to make sure we are also being honest. 
since what I stated is actually my understanding of the definition of a beta component.
Would it be better for the first two sentences to read, `This Beta component is currently under review and is still open for further evolution. It is available for use in product.`

mcarrano: I like that better.

beta-component-promotions/2021.06-promotions/2021.06-promotions.md
```diff
@@ -0,0 +1,42 @@
+# PatternFly Beta Component Promotion Candidates
+## For Release 2021.08
+### [Banner](https://www.patternfly.org/v4/components/banner)
+A banner is a one-line, full color, full width container that can be used to communicate short snippets of information to users. Banners are un-intrusive, non-dismissible, and only support link buttons.
```
abigaeljamie, line 4:
```suggestion
A **banner** is a one-line, full-color, full-width container that can be used to communicate short snippets of information to users. Banners are un-intrusive and non-dismissible, and they only support link buttons.
```

abigaeljamie: Added bolding for component names to align with the bolding formatting we use in design guidelines. Also, should "un-intrusive" be "nonintrusive" ? I keep seeing the latter come up in Google searchers.

beta-component-promotions/2021.06-promotions/2021.06-promotions.md
```diff
@@ -0,0 +1,42 @@
...
+
+![banner](./img/banner.png)
+
+### [Button: inoperableEvents](https://www.patternfly.org/v4/components/button)
+Button prop. An array of events to prevent when the button is in an aria-disabled state.
```
abigaeljamie, line 9: What does this mean? I don't want to suggest any edits unless I know the meaning so that I don't sacrifice accuracy at all :)

beta-component-promotions/2021.06-promotions/2021.06-promotions.md
```diff
@@ -0,0 +1,42 @@
...
+
+### [Button: isAriaDisabled](https://www.patternfly.org/v4/components/button)
+Button prop. Adds disabled styling and communicates that the button is disabled using the aria-disabled html attribute
```
abigaeljamie, line 12:
```suggestion
Button prop. Adds disabled styling and communicates that the button is disabled using the aria-disabled HTML attribute.
``` 
Some lines have periods, and others don't. So I'm just adding periods for consistency 👍

beta-component-promotions/2021.06-promotions/2021.06-promotions.md
```diff
@@ -0,0 +1,42 @@
...
+
+### [Description list](https://www.patternfly.org/v4/components/description-list)
+A description list contains terms and their corresponding descriptions.
```
abigaeljamie, line 15:
```suggestion
A **description list** contains terms and their corresponding descriptions.
```

beta-component-promotions/2021.06-promotions/2021.06-promotions.md
```diff
@@ -0,0 +1,42 @@
...
+
+![description list](./img/description-list.png)
+
+### [FileUpload](https://www.patternfly.org/v4/components/file-upload)
+A file upload component uploads the contents of a file into the browser for further processing. Once uploaded, content may optionally be edited. While the component can upload any file, it is primarily designed for use with text files where the context can easily be displayed in the text area.
```
abigaeljamie, line 20:
```suggestion
A **file upload** component uploads the contents of a file into the browser for further processing. Once contents are uploaded, you have the option of editing them. While the component can upload any file, it's primarily designed for text files where the context can easily be displayed in the text area.
```

beta-component-promotions/2021.06-promotions/2021.06-promotions.md
```diff
@@ -0,0 +1,42 @@
...
+
+![file upload](./img/file-upload.png)
+
+### [Hint](https://www.patternfly.org/v4/components/hint)
+A hint is in-app messaging that provides a one-step reminder, explanation, or call to action for a page or modal.
```
abigaeljamie, line 25:
```suggestion
A **hint** is in-app messaging that provides a one-step reminder, explanation, or call to action for a page or modal.
```

beta-component-promotions/2021.06-promotions/2021.06-promotions.md
```diff
@@ -0,0 +1,42 @@
...
+
+![hint](./img/hint.png)
+
+### [NotificationDrawer](https://www.patternfly.org/v4/components/notification-drawer)
+A notification drawer is a place where users can view and manage notifications that are generated by the system. It allows them to see a list of notifications without the need to navigate away from the current screen.
```
abigaeljamie, line 30:
```suggestion
A **notification drawer** is a place where users can view and manage notifications that are generated by the system. It allows them to see a list of notifications without having to navigate away from the current screen.
```

beta-component-promotions/2021.06-promotions/2021.06-promotions.md
```diff
@@ -0,0 +1,42 @@
...
+
+![notification drawer](./img/notification-drawer.png)
+
+### [SimpleList](https://www.patternfly.org/v4/components/simple-list)
+A simple list provides a list of selectable items that can be shown within a page. Each item is described by a text label. The list may be divided into logical sections by introducing group headers.
```
abigaeljamie, line 35:
```suggestion
A **simple list** provides a list of selectable items that can be shown within a page. Each item is described by a text label. You can divide the list into logical sections using group headers.
```

beta-component-promotions/2021.06-promotions/2021.06-promotions.md
```diff
@@ -0,0 +1,42 @@
...
+
+![simple list](./img/simple-list.png)
+
+### [Tile](https://www.patternfly.org/v4/components/tile)
+A tile is a form of selection that can be used in place of a radio button and is commonly used in forms.
```
abigaeljamie, line 40:
```suggestion
A **tile** is a form of selection that can be used in place of a radio button. It's commonly used in forms.
```

packages/theme-patternfly-org/templates/mdx.js
evwilkin: That sounds great 👍

beta-component-promotions/2021.06-promotions/2021.06-promotions.md
nicolethoen: :) This is a hard one...
`inoperableEvents` is a prop the consumer can pass to the `Button` component. It is an array of strings which represent the html events which should be prevented whenever the button is in an `aria-disabled` state. I don't know the details of why this was added, I guess there are probably accessibility related reasons to use the `aria-disabled` attribute, but there are cases when it doesn't exactly behave the way the consumer expects, so we have given them an extra way to control the button's behavior in this case.

evwilkin (dismissed):
LGTM - thanks!

nicolethoen:
@abigaeljamie I think i've addressed all your comments. Only one follow up question :)

beta-component-promotions/2021.06-promotions/2021.06-promotions.md
abigaeljamie: Ah OK, that makes sense. Thanks!

PR #2473, opened 2021-05-05
fix(JumpLinks): use JumpLinks component on Org

### LINKED ISSUE #2310: Update doc jump links to be keyboard accessible
Currently the Table of contents jump links in our documentation shift focus when selected by clicking or pressing enter, but they do not shift keyboard focus. This makes our pages slightly more difficult to navigate by keyboard for keyboard users and for testing. It would improve the experience to have the focus shift to that section or to the next interactive element in that section.

evwilkin:
Closes #2310 

This PR converts the custom implementation of jump links/scrollspy to use the new JumpLinks component that they transformed into.  This resolves the issue of selecting a link on the keyboard not transferring focus to that part of the page.


evwilkin:
@redallen is there a way to use JumpLinks in the Org build?  I see it's failing due to `document.getElementById is not a function`, not sure how to work around this.

redallen:
Is it working locally? I'm not sure why the prerendering would fail, it's the same prerendering that is used in the React repo. I'll update https://github.com/patternfly/patternfly-react/pull/5733 so it for sure can prerender.

evwilkin:
@redallen thanks for taking a look - it runs locally, but `yarn build` fails with the same error 🤔

redallen:
AH it doesn't fail upstream because I only use the scroll spy in a demo. It's best to wait for https://github.com/patternfly/patternfly-react/pull/5733 to fix it. Polyfilling `document.getElementById` in `prerender.js` will still break.

I suppose a hacky workaround could be to pass `hasScrollSpy={!process.env.PRERENDER}` but this should really be fixed upstream...

redallen (changes requested):
Navigating between pages breaks :/

jump-links

evwilkin:
> Navigating between pages breaks :/
> 

Created https://github.com/patternfly/patternfly-react/issues/5753 to address this.

redallen (approved):
Good stuff!

jessiehuff (commented):
Yayyy! Exciting! 🎉 😄

PR #2474, opened 2021-05-06
fix(codesandbox): root height

### LINKED ISSUE #2304: JumpLinks in the full page example CodeSandbox do not "stick"
In the JumpLinks component example, when viewed in the CodeSandbox, the jump links do not stick to the page.
https://www.patternfly.org/v4/components/jump-links/react-demos

It appears that `#root { height: 100%; }` is missing.

redallen:
Closes #2304


evwilkin (approved):
LGTM

PR #2475, opened 2021-05-06
addition about fas vs far

### LINKED ISSUE #2105: Provide clarity around how to include icons that don't ship with patternfly
https://www.patternfly.org/v4/design-guidelines/styles/icons#all-icons

It isn't clear that PF core only ships `fas` icons, and we have `far` icons listed in the icons table. It would be good to include some instructions on how to include `far` icons into a product so that you can use `<i className="far [insert-icon-name]"></i>` to include an icon.

Here's a rough draft:

## FAS vs FAR

The PF icon set supports use of any FAS (font awesome solid) icons. Therefore, designers may use any FAS icon in their designs. If an FAR icon is needed, developers must include the FAR icon(s) themselves. There are a few ways of doing so:

1. Hosting it yourself (https://fontawesome.com/how-to-use/on-the-web/setup/hosting-font-awesome-yourself)
1. Including via a package manager (https://fontawesome.com/how-to-use/on-the-web/setup/using-package-managers) 
1. Linking to a CDN (https://cdnjs.com/libraries/font-awesome)
1. Use the SVG code directly from fontawesome.com (though this requires attribution)

mmenestr:
closes #2105

### DIFF
--- a/packages/v4/src/content/design-guidelines/styles/icons/icons.md
+++ b/packages/v4/src/content/design-guidelines/styles/icons/icons.md
@@ -110,7 +110,7 @@ Use this for 'fa' regular icons: `<i className="far [insert-icon-name]"></i>`<br
 If you're looking to copy React for an icon:<br />
 Use this for 'pficon' icons: `import { [insert-icon-name] } from '@patternfly/react-icons'`
 
-## Font Awesome solid (FAS) vs Font Awesome regular (FAR)
+### Font Awesome solid (FAS) vs Font Awesome regular (FAR)
 The PF icon set supports use of any FAS (font awesome solid) icons. Therefore, designers may use any FAS icon in their designs. If an FAR icon is needed, developers must include the FAR icon(s) themselves. There are a few ways of doing so:
 
 1. [Hosting it yourself](https://fontawesome.com/how-to-use/on-the-web/setup/hosting-font-awesome-yourself)
--- a/packages/v4/src/content/design-guidelines/styles/icons/icons.md
+++ b/packages/v4/src/content/design-guidelines/styles/icons/icons.md
@@ -116,7 +116,7 @@ The PF icon set supports use of any FAS (font awesome solid) icons. Therefore, d
 1. [Hosting it yourself](https://fontawesome.com/how-to-use/on-the-web/setup/hosting-font-awesome-yourself)
 2. [Including via a package manager](https://fontawesome.com/how-to-use/on-the-web/setup/using-package-managers)
 3. [Linking to a CDN](https://cdnjs.com/libraries/font-awesome)
-4. Use the SVG code directly from fontawesome.com (though this requires attribution)
+4. Use the SVG code directly from [fontawesome.com](https://fontawesome.com) (though this requires attribution)
 
 <IconsTable />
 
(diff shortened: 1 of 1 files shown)


packages/v4/src/content/design-guidelines/styles/icons/icons.md
```diff
@@ -110,6 +110,14 @@ Use this for 'fa' regular icons: `<i className="far [insert-icon-name]"></i>`<br
 If you're looking to copy React for an icon:<br />
 Use this for 'pficon' icons: `import { [insert-icon-name] } from '@patternfly/react-icons'`
 
+## Font Awesome solid (FAS) vs Font Awesome regular (FAR)
+The PF icon set supports use of any FAS (font awesome solid) icons. Therefore, designers may use any FAS icon in their designs. If an FAR icon is needed, developers must include the FAR icon(s) themselves. There are a few ways of doing so:
```
mcoker, line 114: The tricky part here is that this guidance is for users using the FA webfont that ships with PF core. If they're using `@patternfly/react-icons`, they get any icon that has the icon name under the "React" table column, eg, `AngleDoubleLeftIcon`, so they don't need to do anything special to use any of those icons.

I guess we could start this paragraph with

> The Font Awesome icons included with PatternFly core are FAS (Font Awesome solid) webfont icons. If you're using icons via `<i class="[fa, far, fas] [insert-icon-name]"></i>`, designers and developers may use any FAS icon in their designs. If an FAR icon is needed, developers must include the FAR icon(s) themselves. There are a few ways of doing so:

packages/v4/src/content/design-guidelines/styles/icons/icons.md
```diff
@@ -110,6 +110,14 @@ Use this for 'fa' regular icons: `<i className="far [insert-icon-name]"></i>`<br
 If you're looking to copy React for an icon:<br />
```
mcoker, line 110: I can't comment on the lines above this, but they reference:

`<i className="pf-icon [insert-icon-name]"></i>`

Seems like that is a mix of HTML and JSX? Wouldn't it either be `<i class="pf-icon [insert-icon-name]"></i>` or `<i className="pf-icon [insert-icon-name]" />`?  Since we're referencing writing HTML, I would probably write it as `<i class="pf-icon [insert-icon-name]"></i>` cc @tlabaj

mcoker (changes requested):
Left some comments. I think it's fine as is but probably worth clarifying that the bit about including other variations of FA (like FAR) is for users not using `@patternfly/react-icons`.

packages/v4/src/content/design-guidelines/styles/icons/icons.md
mmenestr: I made the other updates - just waiting on Titani for this one as I can't advise!

mcoker: This looks great! I'd like to make the use of `class` and `className` consistent though. I think using `class` in place of `className` on this page should be fine, we write regular HTML in the [a11y guide page](https://patternfly-org-pr-2475-v4.surge.sh/v4/developer-resources/accessibility-guide), too. wdty @redallen? Would love a second opinion!

redallen: `class` is for HTML and `className` is for React.

mmenestr: @mcoker fixed class name if you want to give this another look!

PR #2491, opened 2021-05-14
fix(docs): Fix codesandbox broken/missing images

### LINKED ISSUE #2422: Codesandbox links for Card component can't find images
On the Card component, any of the examples/demos with images has a broken Codesandbox because it can't find the images. On HTML examples, it just results in broken images, but on React examples it means the Codesandbox has errors and doesn't compile.

evwilkin:
Closes #2422 

This PR fixes Codesandbox showing broken images for HTML examples with images and not building for React examples with images by replacing the imports/relative hrefs with links back to the PF.org images.

- updated codesandbox template to point back to pf.org hosted images
- updated webpack config to use file-loader for gif|svg


redallen (approved):
Very clever!

PR #2503, opened 2021-05-21
add guidance for tooltip on hover delay

### LINKED ISSUE #2450: Add guidance in tooltips guidelines
https://github.com/patternfly/patternfly-react/pull/5654

Want to suggest adding delays to tooltips so they don't show up immediately on hover.

Note to self: Zack "only added it into the <TableText> component. other <Tooltip>s still have the old behavior"

mmenestr:
https://github.com/patternfly/patternfly-react/pull/5746

closes #2450

### DIFF
--- a/packages/v4/src/content/design-guidelines/components/tooltip/tooltip.md
+++ b/packages/v4/src/content/design-guidelines/components/tooltip/tooltip.md
@@ -21,7 +21,7 @@ Use tooltips to:
 <img src="./img/tooltip-disabled-button.png" alt= 'disabled edit button with a tooltip about access' width='1500'/>
 
 ## Tooltip delay on hover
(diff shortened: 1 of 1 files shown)


mcarrano (commented):
@mmenestr I think what you have here is OK, but maybe say something about why this is recommended and is the best approach in most cases.

mmenestr:
@mcarrano how does this sound:

"By default, tooltips have a 300ms entry delay when hovering over an object. This is the recommended delay time to avoid users being overwhelmed with tooltips in pages where they are abundant. However, you may customize this delay via prop. For example, if you have a tooltip on an icon, and you would like for it to appear immediately on hover, you may set the delay to 0 ms."

mcarrano (approved):
I think this is fine @mmenestr . I will merge.

PR #2509, opened 2021-06-02
change rocket icon

### LINKED ISSUE #2508: Replace empty state icon in design guidelines to other rocket
Reasoning  - got the following comment on a user test about 404 errors that happened to use the current empty state rocket. Therefore, we decided it was best to change it to a more cartoony rocket for now.

mmenestr:
closes #2508 

@mcarrano just updated the rocket icon used in the getting started empty state image due to a comment (described in iss #2508 ). Michael and I decided it was probably best/easiest to just switch to a different version of a rocket icon.


mcarrano (changes requested):
I'm good with changing this. I never really liked that space shuttle icon anyway. But we should also update this image in that case:

Screen Shot 2021-06-02 at 1 56 35 PM

mmenestr:
@mcarrano yes! I think you may have opened the link before my second push went through because I did change the image - do you mind checking again?

mcarrano (approved):
Looks great. Thanks @mmenestr . I will merge.

PR #2510, opened 2021-06-02
Adding design guidelines for drag and drop

lwrigh:
This PR features the first draft of adding drag and drop design guidelines to PatternFly. If there are any other drag and drop use cases that should be added let know if I missed any.


evwilkin:
@mmenestr it looks like this PR has been open for a bit, adding you but feel free to reassign for review as needed 😄

mmenestr:
Closing because we are going a different route with these guidelines to make them more broad.

PR #2524, opened 2021-06-16
edit icon usage table

### LINKED ISSUE #2499: Update usage guidelines for confirmation modals
The information here is not correct and conflicts with the examples shown above: https://www.patternfly.org/v4/components/modal/design-guidelines/#icon-use-in-modal-dialogs

IMO, the red (!) icon should only be used to report errors, not as a confirmation for actions. The examples showing use of the yellow warning icon to confirm potentially destructive actions in the content above on this page are correct.

mmenestr:
closes #2499 

@mcarrano could you review this and let me know if you think the changes I made make sense in addressing the issue. Or should we just remove the row all together?


mcarrano (changes requested):
@mmenestr I think that the text that you changed is fine. But in reviewing this more closely, under Usage for the red (!) we say:

> **Add to confirmation dialogs or passive dialogs to indicate the highest level of urgency and importance.**

I would argue that we should never use this for confirmation dialogs, only error dialogs and maybe passive dialogs where an error condition is being reported? What do you think?

mmenestr:
@mcarrano yeah, in that case should we just remove that row all together?

mcarrano:
We could remove it. Although we do use the error icon for error dialogs, which are mentioned as one of the common dialog types.

mmenestr:
@mcarrano ok how bout now? I just changed the wording to be focused on errors!

mcarrano (approved):
Looks good @mmenestr . I will merge.

PR #2525, opened 2021-06-16
creating jump link guidelines

### LINKED ISSUE #2199: Create jump links design guidelines
Create design guidelines for this new component: https://patternfly-org-pr-2268-v4.surge.sh/v4/components/jump-links

mmenestr:
closes #2199 

@mcarrano @ajacobs21e do you guys mind reviewing this? Allie, I added a tiny section on top of what you wrote to point out the different "elements" (ie: active vs inactive state).


ajacobs21e:
@mmenestr thank you for bugging me and finishing this!

* the first screenshot of the active / inactive sections is really huge. 

 >  Inactive section: other sections of the page that the user is not currently on.

* do you want to add in here that these are clickable?

> Additionally, they can optionally include subsections.

*  "additionally" and "optionally" seem redundant. "They can include optional subsections"?

>  Jump links can be displayed left aligned or centered

* Should this say "the jump links component" instead?

* space needed between the two mobile screenshots

mmenestr:
@ajacobs21e made updates based on your comments if you want to check again!

mcarrano (commented):
I think what you have here is fine as far as telling me how jump thinks work, however I would expect more guidance about when and why to use jump links vs other styles of navigation. Usually jump links are connected to heading styles on a page. Should the jump link labels match those headings? Not sure, but I think these are the types of usage questions that designers might have.

ajacobs21e:
@mcarrano the difference to me with jump links vs other navigation is that it's in page navigation. 

What about adding this to usage?
> Jump links are sticky as the user jumps or scrolls through content to provide context and a persistent way to navigate through the page. 

Do we want the jump links to only be to headings? I said sections to be more flexible. In the two places I am using this though the jump links are tied to page headings. @abigaeljamie what do you think?

mcarrano:
Maybe that's a more technical question that @nicolethoen @redallen or @mcoker can answer. What can jump links actually jump to? Do there need to be anchor links embedded in the page? <H> tags? "Sections" might be misleading to developers because there's a particular thing called a page section and I don't think we mean that.

redallen:
@mcarrano Jump links jump to any DOM element with an `id=`, usually a heading.

mmenestr:
@mcarrano In that case, would it work to just add an extra section to the guidelines to say this:

"Jump links provide a way to navigate through sections on a page without scrolling. The highlighted active section changes as the user scrolls to the next section. **Each jump link section should match to section headings on the page."**

In terms of using jump links as opposed to another component, the only comparable component I see is tabs - do you think it's worth adding a line that says:

"Jump links differ from tabs in that they are used to navigate a user to different sections on the same page, while tabs are used to navigate between views of a page."

Let me know what you all think!

mcarrano:
> "Jump links provide a way to navigate through sections on a page without scrolling. The highlighted active section changes as the user scrolls to the next section. **Each jump link section should match a section heading on the page."**

I think this is fine

> "Jump links differ from tabs in that they are used to navigate a user to different sections on the same page, while tabs are used to navigate between views of a page."

Maybe this is not needed. I think the new point is that Jump Links are useful for navigating within a long scrolling page.

abigaeljamie:
Sorry for being late to the party! I was out of the office last week on Thursday and Friday. 

+1 to @mcarrano 's comment—that description sounds good to me 👍

mmenestr:
@mcarrano just made the updates, please take another look and merge if it all looks good to you!

ajacobs21e:
>Jump links provide a way to navigate through sections on a page without scrolling. The highlighted active section changes as the user scrolls to the next section. Each jump link section should match a section heading on the page.

Can this be simplified to

>Jump links provide a way to navigate through section headings on a page without scrolling. The highlighted active section changes as the user scrolls to the next section. 

Either way LGTM!

ajacobs21e (dismissed):

mcarrano (approved):
Looks great. Thanks @mmenestr . I will merge.

PR #2541, opened 2021-06-29
sentence case and fix code blocks

redallen:
Fix some odd code styling and use sentence case for titles:
image


jessiehuff (approved):
Thanks! 😄

Issue #2187, opened 2020-10-08
Edit toolbar guidelines to address "toggle group" mentions


mmenestr:
Right now there is a whole section on "toggle groups" in toolbar, but this is confusing, because they are referring to something different than the toggle group component. Thinking we should re-visit the "toggle groups" naming in toolbar to something else - like filter groups or action group...
https://www.patternfly.org/v4/components/toolbar/design-guidelines#toggle-groups

@mcarrano what are your thoughts on this?

mcarrano:
@mmenestr I admit this is confusing. But I chose this name to be in sync with the component documentation here: https://www.patternfly.org/v4/components/toolbar#examples-with-toggle-groups-and-filters  So I'd rather maintain this consistency.

@mcoker @mattnolting do you think this is something we should consider renaming down the road?

mcoker:
@mcarrano @mmenestr Agree, this is confusing. I wonder how well this would play out if we introduce a toggle group component to a toolbar, and we need to create a BEM element for the toolbar to manage it. I'm wondering if we should rename the core toggle group to something like [button group](https://getbootstrap.com/docs/4.5/components/button-group/) or [toggle button group](https://material-ui.com/components/toggle-button/).

mcarrano:
I would be in favor of that @mcoker . Can we do that without a breaking change? I remember some discussion with @christiemolloy when we first introduced the component. It may have been "Toggle button group" originally but we felt that was too long. I think that "Button group" would work. @christiemolloy do you remember why we settled on Toggle group?

mcoker:
> Can we do that without a breaking change?

The toggle group component is still beta, so we can update it. I think it's a worthwhile change.

Also the toggle group component is currently just a toggle button group. Looking forward, I wonder if a "button group" could be something more generic as a way to group buttons/actions, and there is a toggle group variant. We could also move the toggle buttons in the toggle group currently to the button component as a "toggle" variant. I could see using that in components where we have button groups with variants. That may also be able to be used in place of the action list component/layout we have an issue for. WDYT?

mattnolting:
Agree, it's confusing. I think something more explicit like `__collapsible-filters` would avoid the `__group` nomenclature altogether, bearing in mind that individual `__item` s as well as `__group` s are used/placed inside `__collapsible-filters`. It should be a discrete type of container rather than a group, with a modifier. The reason it's called `pf-m-toggle-group` is that I named it before `pf-c-toggle-group` existed and naming is hard..

[...]

PR #2554, opened 2021-07-07
Clarified toggle group definition

mcarrano:
Updated the Toolbar design guidelines to try and clarify the definition of toggle groups in the context of a toolbar and differentiate from the toggle group component. Also updated the subheading levels in the Elements section so they show in the TOC.

Closes: #2187


mmenestr (approved):
Looks good! Merging

PR #2556, opened 2021-07-07
docs(button): add a11y documentation

### LINKED ISSUE #2557: Add a11y documentation for Button

bitwiseor:
A11y documentation for the Button PF component
Closes #2557

### DIFF
--- a/packages/v4/src/content/accessibility/button/button.md
+++ b/packages/v4/src/content/accessibility/button/button.md
@@ -3,9 +3,6 @@ id: Button
 section: components
 ---
 
-# Button
-Accessibility
-
 A **button** is a box area or text that communicates and triggers user actions when clicked or selected. Buttons are interactive elements.
 
 **Keyboard users** should be able to focus on the button using **Tab** to move forward and **Tab + Shift** to move backwards through interactive elements. They should be able to select a focused button using **Space** or **Enter**.
@@ -20,10 +17,10 @@ A **button** is a box area or text that communicates and triggers user actions w
 - Screen readers may read buttons to users out of context. For example, screen reader users can navigate a page specifically by form controls. So button text should be unique and easily understood on its own. Refer to PatternFly's accessibility guide for more information.
 
 The following props/attributes have been added for you or are customizable in PatternFly:
-| Reason used | React prop used to customize | HTML attribute | React component that it should be applied to | Which HTML element it appears on in markup |
-| -- | -- | -- | -- | -- |
-| Adds accessible text to the button if the button doesn’t contain descriptive text on its own. | `aria-label` | aria-label="[button or icon description]" | Button | .pf-c-button.pf-m-plain |
-| Indicates that a button is unavailable to use but doesn’t prevent keyboard or screen reader interactions. Used when a disabled button provides interactive elements like a tooltip. | `aria-disabled` | aria-disabled="true" | Button | button.pf-c-button |
+| Reason used | React prop used to customize | React component that it should be applied to | Which HTML element it appears on in markup |
+| -- | -- | -- | -- |
+| Adds accessible text to the button if the button doesn’t contain descriptive text on its own. | `aria-label` | Button | .pf-c-button.pf-m-plain |
+| Indicates that a button is unavailable to use but doesn’t prevent keyboard or screen reader interactions. Used when a disabled button provides interactive elements like a tooltip. | `aria-disabled` | Button | button.pf-c-button |
 
 
 
--- a/packages/v4/src/content/accessibility/button/button.md
+++ b/packages/v4/src/content/accessibility/button/button.md
@@ -1,0 +1,35 @@
+---
+id: Button
+section: components
+---
+
+# Button
+Accessibility
+
+A **button** is a box area or text that communicates and triggers user actions when clicked or selected. Buttons are interactive elements.
+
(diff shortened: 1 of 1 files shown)


packages/v4/src/content/accessibility/button/button.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Keyboard users** should be able to focus on the button using **Tab** to move forward and **Tab + Shift** to move backwards through interactive elements. They should be able to select a focused button using **Space** or **Enter**.
+
+**Screen reader users** should be able to navigate to the button and it should have text that can be read by a screen reader, or alternative text if it only contains an icon. For alternative text, `aria-label` is the most common choice.
+
+
+## Accessibility application:
+
+- A button containing only an icon without supporting text should be labeled with the `aria-label` prop.
+- A regular disabled button is not focusable, but an aria-disabled button is. Aria-disabled should be used when a disabled button provides interactive elements (like a tooltip).
```
jessiehuff, line 16: A nit, but I think we should use `isAriaDisabled` here instead of "Aria-disabled" to indicate which prop consumers would use.

packages/v4/src/content/accessibility/button/button.md
```diff
@@ -0,0 +1,32 @@
...
+The following props/attributes have been added for you or are customizable in PatternFly:
+| Reason used | React prop used to customize | React component that it should be applied to | Which HTML element it appears on in markup |
+| -- | -- | -- | -- |
+| Adds accessible text to the button if the button doesn’t contain descriptive text on its own. | `aria-label` | Button | .pf-c-button.pf-m-plain |
+| Indicates that a button is unavailable to use but doesn’t prevent keyboard or screen reader interactions. Used when a disabled button provides interactive elements like a tooltip. | `aria-disabled` | Button | button.pf-c-button |
+
+
+
+
+Use [available design guidelines](https://docs.google.com/spreadsheets/d/1wLkKcW99lhkNXsClUa0ihkdl2W7G_eRTY0wGK6_ah60/edit#gid=0) to reference usage.
```
jessiehuff, line 28: Oh these links are part of my original draft notes. They were just to help me plan out what these a11y docs would say. I wasn't planning on including them in the final docs :)

packages/v4/src/content/accessibility/button/button.md
```diff
@@ -0,0 +1,32 @@
...
+
+
+## Accessibility application:
+
+- A button containing only an icon without supporting text should be labeled with the `aria-label` prop.
+- A regular disabled button is not focusable, but an aria-disabled button is. Aria-disabled should be used when a disabled button provides interactive elements (like a tooltip).
+- Screen readers may read buttons to users out of context. For example, screen reader users can navigate a page specifically by form controls. So button text should be unique and easily understood on its own. Refer to PatternFly's accessibility guide for more information.
+
+The following props/attributes have been added for you or are customizable in PatternFly:
+| Reason used | React prop used to customize | React component that it should be applied to | Which HTML element it appears on in markup |
```
jessiehuff, line 20: I think you need a space after your title for the table to show up correctly.

@afialkoff Do you have thoughts about the wording of these sections? I know that we talked about alignment with wording, but I wasn't sure what we landed on.

packages/v4/src/content/accessibility/button/button.md
```diff
@@ -0,0 +1,32 @@
...
+| -- | -- | -- | -- |
+| Adds accessible text to the button if the button doesn’t contain descriptive text on its own. | `aria-label` | Button | .pf-c-button.pf-m-plain |
```
jessiehuff, line 22: @nicolethoen You've reviewed a number of these PRs so far, what are your thoughts on the prop/attribute tables? Is there a certain order that you think looks best and props/attributes that are more/less helpful? Some of my other PRs have slightly different approaches to this.

jessiehuff (changes requested):

packages/v4/src/content/accessibility/button/button.md
nicolethoen: @jessiehuff I answered this question on the Avatar PR, but just to be sure you see it, "in general I really like the tables. If our react components have props that are intentionally built to allow users to customize accessibility, then they are good to highlight in this way. If the attributes are not react props, then we probably don't need to highlight that. They can always add additional html attributes to a component if they want to - there are practically infinite possibilities."

*I'll note though, that aria-label in the case of controls that are usually or always present in a component is probably a good feature to highlight in the table as well. But in the case of button, it is also explicitly a react prop so it's good here.

nicolethoen (approved):

packages/v4/src/content/accessibility/button/button.md
jessiehuff: Since none of these have been added for the user, I'm thinking maybe that we should change the line before the table to something like this: "The following props can be added or are customizable in PatternFly:"

packages/v4/src/content/accessibility/button/button.md
jessiehuff: Can you move the Reason used to the end of the table and renaming the prop column to just "React prop"? I'm thinking of having the prop used as the first column for all of these. (So the order for all of them would be React prop, React component, HTML element, Reason used.) Any of mine that already went in, I'm going to create a PR at the end to clean it all up and make these consistent. :)

nicolethoen (approved):

jessiehuff (approved):
LGTM!

nicolethoen (approved):

PR #2561, opened 2021-07-08
docs(label): add a11y documentation

### LINKED ISSUE #2584: Add a11y documentation for Label

bitwiseor:
closes #2584


nicolethoen (dismissed):

jessiehuff (commented):
This looks good, though I'd add info about the `closeBtnProps` and how consumers can use it to customize the props on the close button of the [label](https://www.patternfly.org/v4/components/label/#props) (for example, change the `aria-label` on the close button which would be really helpful for things like internationalization.)

In order to get it to work, you would pass something like `closeBtnProps={{'aria-label': 'new label'}}` to the Label component. We could put this into the typical prop table or make it a bullet. I'm open to either as long as it's communicated clearly.

jessiehuff (approved):
LGTM! :)

nicolethoen (approved):

PR #2563, opened 2021-07-08
docs(skip-to-content): add a11y documentation

### LINKED ISSUE #2600: Add a11y documentation for Skip to content

bitwiseor:
closes #2600

### DIFF
--- a/packages/v4/src/content/accessibility/skip-to-content/skip-to-content.md
+++ b/packages/v4/src/content/accessibility/skip-to-content/skip-to-content.md
@@ -11,3 +11,4 @@ A **skip to content** is a component that allows screen reader and keyboard user
 
 ## To make skip to content accessible:
 - Ensure that the skip to content href is properly linked to the main content of the page. You may have to add context to the href if there are any query parameters in the URL.
+- If you're using the page component with the skip to content, we add a tabindex="-1" for you. If you are using the skip to content on its own, give the target of the skip to content link a tabindex of -1. Some browsers require this attribute to make this component work consistently.
--- a/packages/v4/src/content/accessibility/skip-to-content/skip-to-content.md
+++ b/packages/v4/src/content/accessibility/skip-to-content/skip-to-content.md
@@ -1,0 +1,13 @@
+---
+id: Skip to content
+section: components
+---
+
+A **skip to content** is a component that allows screen reader and keyboard users to bypass navigation rather than tabbing through it.
+
+**Keyboard users** should be able to find the skip to content using **Tab** at the start of a page with navigation. They should be able to select the skip to content using **Enter**.
+
+**Screen reader users** should also be able to navigate to skip to content using **Tab** and select it to skip past the navigation of a page.
+
+## To make skip to content accessible:
+- Make sure the skip to content can be accessed with **Tab** 
--- a/packages/v4/src/content/accessibility/skip-to-content/skip-to-content.md
+++ b/packages/v4/src/content/accessibility/skip-to-content/skip-to-content.md
@@ -5,12 +5,9 @@ section: components
 
 A **skip to content** is a component that allows screen reader and keyboard users to bypass navigation rather than tabbing through it.
 
-**Keyboard users** should be able to find the skip to content using **Tab** at the start of a page with navigation. They should be able to select the skip to content using **Enter**.
+**Keyboard users** should be able to find the skip to content using **Tab** at the start of a page with navigation. They should be able to select the skip to content using **Enter**. A skip to content does not trap tab focus. It is the first tabbable element on a page, but can be tabbed past and into the navigation of a page if the user prefers.
 
 **Screen reader users** should also be able to navigate to skip to content using **Tab** and select it to skip past the navigation of a page.
 
-A skip to content does not trap tab focus. It is the first tabbable element on a page, but can be tabbed past and into the navigation of a page if the user prefers.
-
 ## To make skip to content accessible:
-- Make sure the skip to content can be accessed with **Tab** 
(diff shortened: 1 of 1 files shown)


nicolethoen (commented):
Does the skip to content component trap tab focus when entering the page by default? I cannot tell from this writing if the consumer would need to wire it up themself or if this desired behavior is built in?

packages/v4/src/content/accessibility/skip-to-content/skip-to-content.md
```diff
@@ -0,0 +1,16 @@
...
+
+**Keyboard users** should be able to find the skip to content using **Tab** at the start of a page with navigation. They should be able to select the skip to content using **Enter**.
+
+**Screen reader users** should also be able to navigate to skip to content using **Tab** and select it to skip past the navigation of a page.
+
+A skip to content does not trap tab focus. It is the first tabbable element on a page, but can be tabbed past and into the navigation of a page if the user prefers.
+
+## To make skip to content accessible:
+- Make sure the skip to content can be accessed with **Tab** 
+- Give the skip to content a tabindex of -1 so that it can receive programmatic focus for keyboard and screen reader users. For more info on this tabindex, see [the WebAIM page for tabindex](https://webaim.org/techniques/keyboard/tabindex#zero-negative-one).
```
nicolethoen, line 16: I may be completely backwards in my understanding, but if I give it the skip to content a tabindex of -1, wouldn't that mean that I could NOT tab to it anymore? Isn't that the opposite of what we want?

bitwiseor: You're right! The target of the skip link should have tabindex -1 to make it work consistently on IE and Chrome, but the element itself needs to be focusable.

nicolethoen (approved):

packages/v4/src/content/accessibility/skip-to-content/skip-to-content.md
```diff
@@ -0,0 +1,16 @@
...
+section: components
+---
+
+A **skip to content** is a component that allows screen reader and keyboard users to bypass navigation rather than tabbing through it.
+
+**Keyboard users** should be able to find the skip to content using **Tab** at the start of a page with navigation. They should be able to select the skip to content using **Enter**.
+
+**Screen reader users** should also be able to navigate to skip to content using **Tab** and select it to skip past the navigation of a page.
+
+A skip to content does not trap tab focus. It is the first tabbable element on a page, but can be tabbed past and into the navigation of a page if the user prefers.
```
jessiehuff, line 12: I would personally move this up to the keyboard section since it's all about keyboard functionality.

packages/v4/src/content/accessibility/skip-to-content/skip-to-content.md
```diff
@@ -0,0 +1,16 @@
...
+
+## To make skip to content accessible:
+- Make sure the skip to content can be accessed with **Tab**
```
jessiehuff, line 15: I don't know if I'd add this in the recommendation section because technically every component should be tested. I think this section should be more about when they're creating it.

packages/v4/src/content/accessibility/skip-to-content/skip-to-content.md
```diff
@@ -0,0 +1,16 @@
...
+
+**Keyboard users** should be able to find the skip to content using **Tab** at the start of a page with navigation. They should be able to select the skip to content using **Enter**.
+
+**Screen reader users** should also be able to navigate to skip to content using **Tab** and select it to skip past the navigation of a page.
+
+A skip to content does not trap tab focus. It is the first tabbable element on a page, but can be tabbed past and into the navigation of a page if the user prefers.
+
+## To make skip to content accessible:
+- Make sure the skip to content can be accessed with **Tab** 
+- Give the target of the skip to content link a tabindex of -1. Internet Explorer and Chrome require this attribute to make this component work consistently.
```
jessiehuff, line 16: I don't think this one (the tabindex) is an absolute requirement? It doesn't seem to be included in the [WCAG techniques](https://www.w3.org/TR/WCAG20-TECHS/G1.html) or the example they include. Curious of your thoughts on this one though or which resource you found. Maybe it's totally fine here but thought I'd see which resource you're referencing for this. 

Looking at the skip to content example, I think the big thing is just making sure it's hooked up correctly with a proper href like `href="#main-content"` on the SkipToContent and having it point to the id of the main content. It's a pretty easy set up if they're using the Page component, and there is the possibility that they'll have to add context to the href if they have any query parameters in the URL. That might be something we should mention as well. For example, when I [added the skip to content to OpenShift](https://github.com/openshift/console/pull/6596/files), it was a fairly straightforward fix, but it needed context in the href.

bitwiseor: https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/mouse-and-keyboard-events/skip-to-main-content/

This is where I found that info; it might be outdated though, so I can remove it!

jessiehuff: Based on our discussion, I think it would make sense to add a bullet point that says that if you're using the Page component with the Skip to Content, we add a `tabindex="-1"` for you, but if you are using the Skip to Content on its own, you will need to add it to the target (you can keep some of your previous wording). 

I realized in our meeting that the reason I didn't need to add the tabIndex to OpenShift is that it's included when using the Page and Skip to Content together, which I did. 🙂

jessiehuff (approved):
LGTM!

nicolethoen (approved):

PR #2569, opened 2021-07-08
docs(page): add a11y documentation

### LINKED ISSUE #2601: Add a11y documentation for Page

bitwiseor:
closes #2601

### DIFF
--- a/packages/v4/src/content/accessibility/page/page.md
+++ b/packages/v4/src/content/accessibility/page/page.md
@@ -29,4 +29,4 @@ The following props/attributes have been added for you or are props/attributes t
 | An id to use for the [role="main"] element | `mainContainerId` |  | Page |
 | A tabIndex to use for the [role="main"] element. Make this value null to unset it | `mainTabIndex` |  | Page |
 | Skip to content component for the page | `skipToContent` |  | Page |
-| Value for the role on the \<main\> element | `role` |  | Page |
+| Value for the role on the `<main>` element | `role` |  | Page |
--- a/packages/v4/src/content/accessibility/page/page.md
+++ b/packages/v4/src/content/accessibility/page/page.md
@@ -16,7 +16,7 @@ Consider using a **skip to content** component on your page so that assistive te
 ## To make a page accessible:
 - Use roles to identify the main region and the banner region of a page.
 - Use aria-expanded on to indicate the state of the visibility of expandable content.
-- Use a tab index of -1 to give the main region of a page programmatic focus.
+- Use a tabindex of -1, as an entire page component should not be focusable by tabbing (though individual components within it may be).
 - For every element that does not contain text, add an `aria-label` attribute that contains alternative text.
 - If using notifications in the masthead, refer to notification badge accessibility guidelines.
 
--- a/packages/v4/src/content/accessibility/page/page.md
+++ b/packages/v4/src/content/accessibility/page/page.md
@@ -21,12 +21,12 @@ Consider using a **skip to content** component on your page so that assistive te
 
 The following props/attributes have been added for you or are props/attributes that can be customized:
 
-| Reason used | React prop used to customize | HTML Attribute | React component that it should be applied to |
+| React prop | React component that it should be applied to | Which HTML element it appears on in markup | Reason used |
 | -- | -- | -- | -- |
-| Labels the navigation toggle button | `aria-label` | aria-label="Global Navigation" | PageHeader |
-| Identifies the element controlled by the toggle | `aria-controls` |  | PageHeader |
-| Labels the main section | `mainAriaLabel` |  | Page |
-| An id to use for the [role="main"] element | `mainContainerId` |  | Page |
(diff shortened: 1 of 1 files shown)


nicolethoen (commented):
I'm confused about this suggestion: Use a tab index of -1 to give the main region of a page programmatic focus. What is the scenario you're imagining this would be used? How does -1 allow for programmatic focus?

bitwiseor:
I think I was confused on what the tabindex meant when I first wrote this. The page component shouldn't be focusable (so -1 keeps it out of the tab order), but components within it such as navigation elements or skip to content may be focusable.

nicolethoen (commented):
The update about tabindex makes way more sense, thank you!

However, the Page component is just a div right? So by default it wouldn't try to capture focus? In this case a tabindex of -1 is likely unnecessary?

packages/v4/src/content/accessibility/page/page.md
```diff
@@ -0,0 +1,33 @@
...
+**Keyboard users** should be able to navigate through the masthead of the page and into other interactive elements of the page using **Tab** and **Shift + Tab** to move forward and backward through interactive elements. 
+
+**Screen reader users** should be able to navigate through and interact with the masthead of the page using the same interactions as keyboard users. Elements of a page that do not have visible descriptive text, such as toggle icons, should have alternative text for screen readers.
+
+Consider using a **skip to content** component on your page so that assistive technology users don’t need to make their way through the entire navigation menu each time they go to another page.
+
+## To make a page accessible:
+- Use roles to identify the main region and the banner region of a page.
+- Use aria-expanded on to indicate the state of the visibility of expandable content.
+- Use a tabindex of -1, as an entire page component should not be focusable by tabbing (though individual components within it may be).
```
jessiehuff, line 19: Since we add this for them, I don't think we should include it here. I think it will cause more confusion than anything.

packages/v4/src/content/accessibility/page/page.md
```diff
@@ -0,0 +1,33 @@
...
+- For every element that does not contain text, add an `aria-label` attribute that contains alternative text.
+- If using notifications in the masthead, refer to notification badge accessibility guidelines.
```
jessiehuff, line 21: I'm thinking that for this section we should reword your first bullet to be about semantic elements and roles, maybe labeling next (for example, if there are two `<nav>` items then they should be labeled), and then maybe these two points? We could probably even link [this tutorial](https://www.w3.org/WAI/tutorials/page-structure/content/) or use it for any points we might want to add.

packages/v4/src/content/accessibility/page/page.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+| Reason used | React prop used to customize | HTML Attribute | React component that it should be applied to | Which HTML element it appears on in markup |
+| -- | -- | -- | -- | -- |
+| Labels the navigation toggle button | `aria-label` | aria-label="Global Navigation" | PageHeader |  |
+| Identifies the element controlled by the toggle | `aria-controls` |  |  | .pf-c-page__header-brand-toggle > .pf-c-button |
+| Allows the main region to receive programmatic focus |  | tabindex="-1" |  | .pf-c-page__main |
+| Provides a hook for sending focus to new content. |  | id="[id]" |  | .pf-c-page__main |
+| Indicates that the expandable content is visible and the current state of the contents. |  | aria-expanded="true/false" |  | .pf-c-page__header-brand-toggle > .pf-c-button |
+
```
jessiehuff, line 33: For this section, I don't think we should include any HTML attributes that we've added for them. After talking to the leads and a few developers, I think that adding these could end up confusing developers and make them think that they should add them onto what is already there. If they are interested in learning about the ones that are already there, they can still visit the a11y section in the html tab. (This means we'd take out the tabindex, id, and aria-expanded in this table.)

I think the following props should probably be included in this table instead:

- mainAriaLabel
- mainContainerId
- mainTabIndex
- skipToContent
- role

I didn't include (aria-label and aria-controls since you already have them defined :) )

jessiehuff: They're just missing the HTML section. I also just noticed that this one is labeled differently? Instead of "Which HTML element it appears on in markup", it's just "HTML element". Maybe we could change this to match the others, and then after what's in the Reason used section, just add "defaults to "Global Navigation".

For the "Which HTML element it appears on in markup" section, you could add the HTML elements below:
For `aria-label` : .pf-c-button.pf-m-plain
`aria-controls` : .pf-c-button.pf-m-plain
`mainAriaLabel` : ".pf-c-page__main
`mainContainerId` : .pf-c-page__main
`mainTabIndex` : .pf-c-page__main (in the reason used add, "defaults to -1")
`skipToContent` : .pf-c-button.pf-m-primary.pf-c-skip-to-content
`role`: .pf-c-page__main

jessiehuff (approved):
LGTM!

nicolethoen (approved):

PR #2570, opened 2021-07-08
docs(tree-view): add a11y documentation

### LINKED ISSUE #2599: Add a11y documentation for Tree view

bitwiseor:
closes #2599

### DIFF
--- a/packages/v4/src/content/accessibility/tree-view/tree-view.md
+++ b/packages/v4/src/content/accessibility/tree-view/tree-view.md
@@ -1,0 +1,33 @@
+---
+id: Tree view
+section: components
+---
+
+A **tree view** is a structure that displays data in a hierarchical view. It can be used in a dropdown, drawer, primary-detail, modal, or wizard. Tree views are often used when grouping similar decisions, action items, or navigation elements together.
+
+A default tree view provides a tree structure with items and arrows to expand or collapse child items. Every tree view can be extended with these optional capabilities.
+
+**Keyboard users** should be able to interact with and navigate through the interactive tree view. Initial focus should be set on the first focusable element in the tree view. The user should be able to focus on one element of the tree with **Tab**. Once focused on a tree item, the user should also be able to move between visible nodes using **Up arrow** and **Down arrow**. **Left arrow** can be used to collapse an expanded node, or move to the parent of an already-collapsed or childless node. **Right arrow** should expand the focused node if it is collapsed but has child nodes, move to the first child of an already-expanded node, or do nothing on an end node.
+
+Keyboard interactions for tree view:
+
+|  |  |
+| -- | -- |
+| Tab | Focus on the first (or previously selected) element of the tree |
+| Up arrow | Move up through visible nodes |
+| Down arrow | Move down through visible nodes |
+| Left arrow | Collapse focused node if expanded, or move to the parent of a collapsed/childless node |
+| Right arrow | Expand focused node if collapsed, move to first child if already expanded, or do nothing if end node |
+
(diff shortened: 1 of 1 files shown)


nicolethoen (approved):

packages/v4/src/content/accessibility/tree-view/tree-view.md
```diff
@@ -0,0 +1,33 @@
...
+| Down arrow | Move down through visible nodes |
+| Left arrow | Collapse focused node if expanded, or move to the parent of a collapsed/childless node |
+| Right arrow | Expand focused node if collapsed, move to first child if already expanded, or do nothing if end node |
+
+**Screen reader users** should be able to interact with the tree view using their mouse or keyboard. Nodes of a tree view should have descriptive text labels whenever possible. If a tree view node only contains a graphic (such as an icon), it should have alternative text: `aria-label` is the most common way to add this. A tree view should be identified as such when the screen reader first focuses on it. When users navigate to a new node, indicate its branch level , number of child nodes, and number of sibling nodes.
+
+## To make tree view accessible:
+- Use aria roles to identify the tree (“tree”), the nodes in the tree (“treeitem”), and sets of child nodes (“group”).
+- Indicate a selected node with aria-selected.
+- **For every parent node**, add an `aria-expanded` attribute that says whether the node is collapsed or expanded. This will make the tree view more screen reader friendly.
```
jessiehuff, line 27: We've actually added all of these for them so they shouldn't have to do any of this. :) 

Some things that could be noted are specific customizations that can or should be made depending on the variant: 
- The search variant could probably use an aria-label describing what the user can search like our example:
`searchProps={{ id: 'input-search', name: 'search-input', 'aria-label': 'Search input example' }}` (trying to exemplify the aria-label specifically)
- The checkbox variant, each of the checkboxes could have an aria-label or should be associated: `checkProps: { 'aria-label': 'app-launcher-check', checked: false }` (Note that both this and the previous bullet have to be passed down to where that prop needs to be: i.e. the search or the check)
- The actions variant should have an `aria-label` if they're using buttons without visible text

packages/v4/src/content/accessibility/tree-view/tree-view.md
```diff
@@ -0,0 +1,33 @@
...
+
+The following props/attributes have been added for you or are customizable in PatternFly:
+
+| Reason used | React prop used to customize | HTML Attribute | React component that it should be applied to | Which HTML element it appears on in markup |
+| -- | -- | -- | -- | -- |
```
jessiehuff, line 32: You could probably include the `searchProps` and `checkProps` in here with how to pass down the `aria-label`. 

This all looks really good though!! 😄

nicolethoen:
@bitwiseor there are still some comments to address on this PR :)

nicolethoen (dismissed):

packages/v4/src/content/accessibility/tree-view/tree-view.md
jessiehuff: Super close!! Just a few more details to edit to make things clearer. It looks like for the `checkProps`  the HTML element that it's applied to is the "checkbox input". (Though I don't see any classes to identify it with so maybe just state it's the checkbox input?) I would move the "defaulted to { checked: false }" to the reason used? Open to suggestions on where you think it belongs best though. :) And with the `searchProps`, the HTML element that it's applied to is the ".pf-c-form-control pf-m-search" search input.

bitwiseor: Updated, and I updated the table headers to match the other component docs as well.

nicolethoen (approved):

jessiehuff (approved):
LGTM! :)

PR #2577, opened 2021-07-09
docs(Card): Add a11y documentation for card

### LINKED ISSUE #2576: Add a11y documentation for Card

jessiehuff:
Closes #2576


nicolethoen (commented):
I think explaining where the `titleId` for `aria-labelledby` should come from would be helpful. Is it always required for the expandable variant? Is there a specific component this could or should be referring to?

nicolethoen:
@jessiehuff  I still have some comments on this PR :)

jessiehuff:
> I think explaining where the `titleId` for `aria-labelledby` should come from would be helpful. Is it always required for the expandable variant? Is there a specific component this could or should be referring to?

Really great catch, @nicolethoen!! I'll update it to add that. 😄

nicolethoen (approved):

PR #2586, opened 2021-07-09
fix(tsdocgen): interface normalizer

redallen:
Check v4/components/slider/#sliderstepobject:
image


mturley:
@redallen I wonder if it would be easy to turn the cells that reference denormalized interfaces into links that scroll you to the interface's table?

redallen:
It's not easy, but it's possible.

mturley:
@redallen I opened https://github.com/patternfly/patternfly-org/issues/2608 to capture this, maybe we can make it happen :)

PR #2614, opened 2021-07-21
feat(propsTable): Convert interface names referenced in prop types on the same page into links

### LINKED ISSUE #2608: Turn normalized prop type interface references into links
Following up on https://github.com/patternfly/patternfly-org/pull/2586, I noticed [this related comment](https://github.com/patternfly/patternfly-react/pull/5984#pullrequestreview-705561006). While having these interfaces broken out like this is very helpful, I agree that it becomes a little difficult to scroll around and look up related types. It would be great if we could turn these interface names into links so that you can jump from a component's prop types to the definition of one of the interfaces easily.

cc @redallen @dlabrecq

mturley:
Resolves #2608.

@redallen I had a hunch this would work and couldn't help myself.

Followup to #2586. When rendering each row of the `PropsTable`, this new `PropTypeWithLinks` component will look at the type name for that row and see if it contains full words matching any of the other component/interface names on the page. If so, it will split the type name into pieces isolating those matches and render the matched segments as links using the existing `slugger()` helper to generate the ids. This `slugger()` is what the `AutoLinkHeader` already uses to generate the ids of the referenced prop tables, so we know the ids being linked will always be consistent.

See the prop type for `customSteps` at v4/components/slider/#slider:

tUm5lnawHl

Several examples can also be seen in the new v4/components/table/ which has lots of these normalized interfaces.

### DIFF
--- a/packages/theme-patternfly-org/components/propsTable/propTypeWithLinks.js
+++ b/packages/theme-patternfly-org/components/propsTable/propTypeWithLinks.js
@@ -1,0 +1,33 @@
+import React from "react";
+import { slugger } from "../../helpers/slugger";
+
+export const PropTypeWithLinks = ({ type, allPropComponents }) => {
+  // If this type contains the names of any other components on this page as full words, turn them into links.
+  const referencedComponentNames = allPropComponents
+    .filter((component) => type.match(new RegExp(`\\b${component.name}\\b`)))
+    .map((component) => component.name);
+  if (referencedComponentNames.length > 0) {
+    return (
+      <React.Fragment>
+        {type
+          .split(new RegExp(`\\b(${referencedComponentNames.join("|")})\\b`))
+          .map((segment) =>
+            referencedComponentNames.includes(segment) ? (
+              <a
+                key={segment}
+                onClick={(event) => {
+                  event.preventDefault();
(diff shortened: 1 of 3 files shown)


redallen (approved):
Awesome! Next up: linking across components or to external sites with type defs?

mturley:
FYI @redallen I originally implemented the link with a simple href instead of using the `onClick` with `scrollIntoView`, but that put a hash in the URL which then broke the table of contents due to https://github.com/patternfly/patternfly-org/issues/2615. If we address that before merging this, we could clean it up by using an href. I'm also not sure about the accessibility of these links.

mturley:
> Awesome! Next up: linking across components or to external sites with type defs?

That would be really cool! I'm not sure where to start there since this component already had the convenient `propComponents` in scope. That sounds like a much bigger project :D

redallen (approved):

mturley:
FYI @nicolethoen @dlabrecq I did the thing :)

dlabrecq (commented):
Sweet, this looks great!

evwilkin:
@mturley late to the game but this is a great help, thanks for the contribution!

mturley:
My pleasure, thanks @evwilkin !

PR #2624, opened 2021-08-04
fix(externalwatcher): watch for adding files

### LINKED ISSUE #2622: Adding a new TS example to a MD file while the dev server is running does not add the new TS file to the watcher
Here's a corner case that we'll only hit when writing fresh new component examples using the ```ts file="filename" syntax established by https://github.com/patternfly/patternfly-org/pull/2607.

If I start the dev server in patternfly-react, then create a new TS file and add it to one of our markdown files, it reloads and renders the example from the new file properly (because the markdown file was the last thing I changed and it was already being tracked by the file watcher). However, if I then modify the new TS file without modifying anything else, the dev server doesn't automatically recompile. I have to restart the server for it to notice the new file.

Discussing with @redallen, I found that the list of files to be watched is gathered in `theme-patternfly-org/scripts/md/parseMD.js:280`, which only runs once when the dev server starts. He says he has a fix and asked me to open this to track it.

redallen:
Closes #2622


mturley:
That was quick! thanks @redallen

Issue #1953, opened 2020-07-16
Add guidance around custom alerts to Alerts docs


gdoyle1:
Add section about using custom alerts, specifically about having the ability to change the icon, and color? of the alert to whatever fits your needs best, if it's not already covered in the given options.

mmenestr:
Add section about using custom alerts, specifically about having the ability to change the icon, and color? of the alert to whatever fits your needs best, if it's not already covered in the given options.

mmenestr:
This is still not in React - so leaving this on hold for now. There is also an issue for this in the design-kit that hasn't been completed either bc it's not in React either.
Tracked here: https://github.com/patternfly/patternfly/issues/3257

PR #2638, opened 2021-08-09
Adding custom alert guidance

afialkoff:
Adding a "Customizing alerts" section as per #1953 

@mmenestr should we also include a visual example of what custom alerts might look like, or does the link to the examples suffice?

Thanks for your patience on this PR! :)


mmenestr:
I don't think an image is necessary, but we can link to these examples: /components/alert#custom-icons!

mmenestr (approved):

PR #2644, opened 2021-08-11
added TaskIcon to icons table

### LINKED ISSUE #2531: Add task icon to icons table
In [patternfly#4165](https://github.com/patternfly/patternfly/issues/4165) , we added the task icon to the PF webfont. I think there will be a follow-up issue to get it added to the React icons as well, since that only added it to Core.

We also want to add a new row to the icons table on the Icons page for this. Here's the information needed:

Style: pf-icon
Name: Task
React_name: Not created yet but will be called `taskicon`
Contextual_usage: Indicates the ability to open a task drawer. Also can represent tasks or activity.
Tooltip_label: Tasks

evwilkin:
Closes #2531 

This PR adds the TaskIcon to the list of "All icons" on the website.

### DIFF
--- a/packages/v4/src/content/design-guidelines/styles/icons/icons.js
+++ b/packages/v4/src/content/design-guidelines/styles/icons/icons.js
@@ -689,6 +689,14 @@ export const iconsData = [
     "Contextual_usage": "Represents data view content in a dashboard",
   },
   {
+    "Style": "pf-icon",
+    "Name": "pf-icon-task",
+    "React_name": "TaskIcon",
+    "Type": "",
+    "Contextual_usage": "Indicates the ability to open a task drawer. Also can represent tasks or activity.",
+    "Label": "Tasks"
+  },
+  {
     "Style": "fas",
     "Name": "fa-th",
     "React_name": "ThIcon",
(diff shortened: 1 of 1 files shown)


bdellasc:
> @bdellasc you're my go-to for icons, any familiarity with this ask and if it should have a "Type"? A quick recap, we had the "Type" field visible originally (with values like Framework, Brand, Status, Object, Action) and though we've hid it from view these values are still searchable. If not, this should be good to go.

Do we know the reason was for hiding the "Type" field?

bdellasc:
@evwilkin I assume we'd still want to populate the field since they're still searchable. Looking at the original issue, my reflex response would be that it falls under the "Object" and "Framework" categories...have we had any other icons that fall under multiple categories?

evwilkin:
Thanks for taking a look @bdellasc!  We originally hid the "Type" column when adding the "Tooltip label" column in [this PR](https://github.com/patternfly/patternfly-org/pull/2379), I can't find the conversation leading to that decision but I believe due to the restricted width of the table there wasn't enough room for an additional column.

We don't have any other icons that fall under other categories currently, but that said there's no reason why we can't do that - I'll go ahead and add both of these types. 👍

bdellasc (dismissed):
Based on Gina's [Contextual usage description](https://github.com/patternfly/patternfly-org/issues/2531) please add "Object" and "Framework" as the types for this. Thanks!

nicolethoen (approved):

PR #2652, opened 2021-08-24
updated link to use history.pushState to avoid route change

### LINKED ISSUE #2615: After clicking on an AutoLinkHeader link icon, the right-side Table of Contents JumpLinks do not work
I noticed this while working on #2614. If you navigate to a specific section of a docs page using one of the permalinks created by `AutoLinkHeader`, e.g. https://www.patternfly.org/v4/components/slider#slider and then try to click one of the links in the table of contents on the right, they just scroll you to the top of the page instead of scrolling you to the matching section. These links behave normally if the URL does not have a hash.

cc @redallen

evwilkin:
Fixes #2615 

This PR replaces `window.location.hash` with `history.pushState` which updates the URL but does not trigger route change which was causing a re-render of the page and breaking the jumplinks.

This fixes the issue of clicking an autolinkheader (link icon beside a section/example heading corresponding with a jumplink) would break the jumplinks, always causing the first jumplink to show as active and no longer following the user's scrolling on the page.

### DIFF
--- a/packages/theme-patternfly-org/components/link/link.js
+++ b/packages/theme-patternfly-org/components/link/link.js
@@ -31,7 +31,8 @@ export const Link = ({
       if (referencedElement) {
         referencedElement.scrollIntoView();
       }
-      window.location.hash = url;
+      // update URL without triggering route change
+      history.pushState({}, '', url);
     };
   }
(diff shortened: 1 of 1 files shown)


redallen (approved):
Nice find!

PR #2669, opened 2021-08-31
fix(docs): update props table column widths & word breaks

### LINKED ISSUE #2645: Documentation width limit causes a lot of wrapping
Having the page width limited causes a lot of wrapping in documentation sections. It would be great to let those stretch further on a wider viewport.

evwilkin:
Closes #2645 

This PR:
- Increases the "Type" column on component props tables from 15% to 20% width to help with text wrapping.
- Adds `breakWord` transform to the "Description" column on component props tables to fix long strings causing table to exceed parent container's width, such as with the URL in the `locale` prop description for calendar month [seen here](https://www.patternfly.org/v4/components/calendar-month/#props).
- Removes the "Required" column from the react props tables, replacing this with a red asterisk beside required prop names.
  - For components with required props, instructional text at the top of the table explains that the asterisk marks required props, and those props are listed first in alphabetical order before the rest of the optional props.

### DIFF
--- a/packages/theme-patternfly-org/components/example/example.css
+++ b/packages/theme-patternfly-org/components/example/example.css
@@ -83,3 +83,7 @@
   --pf-c-badge--m-unread--Color: var(--pf-global--primary-color--100);
   border: 1px solid var(--pf-global--primary-color--100);
 }
+
+.ws-prop-required {
+  color: var(--pf-global--danger-color--100);
+}
--- a/packages/theme-patternfly-org/components/propsTable/propsTable.js
+++ b/packages/theme-patternfly-org/components/propsTable/propsTable.js
@@ -32,10 +32,7 @@ export const PropsTable = ({
         className="pf-u-mt-md pf-u-mb-lg"
         variant="compact"
         aria-label={title}
-        caption={(rows.some(prop => prop.required))
-          ? <div><span className="ws-prop-required">*</span>required</div>
-          : null
-        }
+        caption={<div><span className="ws-prop-required">*</span>required</div>}
         cells={columns}
         gridBreakPoint="grid-lg"
         rows={rows
--- a/packages/theme-patternfly-org/components/propsTable/propsTable.js
+++ b/packages/theme-patternfly-org/components/propsTable/propsTable.js
@@ -5,7 +5,8 @@ import {
   TableHeader,
   TableBody,
   fitContent,
-  cellWidth
+  cellWidth,
+  breakWord
 } from '@patternfly/react-table';
 import { AutoLinkHeader } from '../autoLinkHeader/autoLinkHeader';
 import { PropTypeWithLinks } from './propTypeWithLinks';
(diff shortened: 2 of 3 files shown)


evwilkin:
@srambach @mcarrano I ran with this and updated the width of the "Type" column slightly, while keeping the table the same max width.  This does take away 5% of width from the "Description" column, but that field is typically sentence text that is easier to wrap without much confusion, let me know what you think.

mcarrano (commented):
@evwilkin I think this definitely helps. Did you also increase the width of the Default column? I think that looks better, too. Although it still wrap awkwardly for longer strings.

I'm wondering if there is a way to optimize this even more. The "Required" column takes a lot of space and mostly says, "No." Is there another way that we can indicate what props are required?

srambach:
Thank you, I think those definitely help. What do you think about making the text size and padding on the HTML usage sections match the React ones to help to get more information on one screen without going wider? The HTML docs are not only a step up in font size, but also have quite a bit more padding around each entry.
image

mcarrano:
That's a good point @srambach . I don't see any reason these pages should be different.

evwilkin:
@srambach thanks for calling that out, those HTML tables were built differently and didn't have the `pf-m-compact` styling but are now updated.

@mcarrano that's a good idea, we can remove the "Required" column if we determine another way to indicate this.  We can easily highlight/bold the rows or prop names that are required, or add a flag or icon of some sort.  Do you or @mceledonia have any suggestions for a way to indicate a prop is required, without needing the ["Required" column](https://patternfly-org-pr-2669-v4.surge.sh/v4/components/about-modal#props)?

srambach:
That helps a little. I still have the overwhelming urge to grab it and stretch it out though! 

image

image

mcarrano:
@evwilkin maybe use the red asterisk following the prop name to indicate "required" like we do on forms? Bolding required prop names could also work. But in either case we probably want to add instructional text at the top of the table to explain.

mcarrano:
@evwilkin I think this is pretty clear with the red asterisk for required props. Maybe share with a few developers on the team to see if it's intuitive to them?

evwilkin:
@mcarrano thanks for the suggestion, feedback from devs was to move required props first and add screenreader support for the 'required' asterisk, otherwise they all gave a 👍 to this idea.

nicolethoen:
I think this looks good! My only question is whether every table should include a key that says *Required, even if it does not have any required props? I'm imagining that'd communicate that there are no required props a little more intuitively?

evwilkin:
@nicolethoen good thought - this should be built in already, [here's an example](https://patternfly-org-pr-2669-v4.surge.sh/v4/components/accordion#props).  Is that what you were thinking?

nicolethoen:
In this case, it's a little clearer that not having a star means a prop is optional because one of the tables on the page does have a required prop. However, I don't think it'd add too much noise to just ALWAYS add the required star key to the table. That way every table tells the user that required props would be starred.

evwilkin:
@nicolethoen thanks for clarifying, I definitely understood exactly the opposite of what you meant 😄  I see your point now, as a component without required props will now not have any mention on the page of how to even tell that the props are all optional - this makes sense, I'll revert the logic.

PR #2676, opened 2021-09-13
add typeof and keyof parsing

### LINKED ISSUE #2674: TypeScript examples parser does not recognize `keyof` syntax
Another syntax quirk discovered while working on https://github.com/patternfly/patternfly-react/pull/6168.

In the compound expandable example, I'm trying to use the keys in the `Repository` interface (the example data type) as the values of a `Record` which maps repo name to key name for storing the state of which cells are expanded:

```ts
interface Repository {
  name: string;
  branches: number;
  prs: number;
  workspaces: number;
  lastCommit: string;
}

// ...

const [expandedCells, setExpandedCells] = React.useState<Record<string, keyof Repository>>({});
```

That useState line is giving me `SyntaxError: Unexpected token`. I've broken this down to two pieces of syntax that it doesn't recognize: `keyof` and a type param with nested type params. Opened a separate issue for the latter: https://github.com/patternfly/patternfly-org/issues/2675

To isolate the `keyof` issue, I also tried:

```ts
type RepositoryColumnKey = keyof Repository;
```

which gives the same error on its own.

redallen:
Closes #2674

### DIFF
--- a/packages/theme-patternfly-org/helpers/acorn-typescript.js
+++ b/packages/theme-patternfly-org/helpers/acorn-typescript.js
@@ -381,6 +381,9 @@ module.exports = Parser => class TSParser extends Parser {
           case tsTypeOperator.infer:
             node = this.parseTSInferType()
             break
+          case tsTypeOperator.keyof:
+            node = this.parseTSKeyofType()
+            break
           default:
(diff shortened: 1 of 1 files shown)


mturley:
Thanks @redallen !

PR #2748, opened 2021-11-05
docs(beta): linked to beta promotion schedule in beta alert

### LINKED ISSUE #1541: Experimental features need information about time/criteria to become full-fledged
Those outside the PF team need to have an idea when a given experimental feature may become full-fledged, and/or what criteria must be met. I have received feedback from product team members that they didn't think they could use a feature they needed and didn't know when they'd be able to, so they'd begun to build their own.

evwilkin:
Closes #1541 

This PR adds a link to the beta promotion schedule within the beta alert at the top of beta components.

@nicolethoen let me know if you know of a better place to link to, or if we'd like to add a README to this linked page with a little context.

### DIFF
--- a/packages/theme-patternfly-org/templates/mdx.js
+++ b/packages/theme-patternfly-org/templates/mdx.js
@@ -54,7 +54,8 @@ const MDXChildTemplate = ({
       {beta && (
         <InlineAlert title="Beta feature">
           This Beta component is currently under review and is still open for further evolution. It is available for use in product.
-          Beta components are considered for promotion on a quarterly basis. Please join in and give us your feedback or submit any questions on the <a href="https://forum.patternfly.org/">PatternFly forum</a> or via <a href="//slack.patternfly.org/" target="_blank" rel="noopener noreferrer">Slack</a>  Learn more about Beta components <a href="https://github.com/patternfly/patternfly-org/blob/main/beta-component-promotion/README.md">here</a>.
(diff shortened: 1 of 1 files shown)


nicolethoen (commented):
I think this is probably the simplest way to address this.
The only other though I have is to consider adding a read-me to that linked directory with more information? I could type something up.

And maybe update the promotion directory names so that they are more meaningful to people besides the patternfly leads. Even if the directory names included the word candidates or components, they might be clearer.

mcarrano (commented):
I don't have a better idea, either. Although I do find it strange that we don't talk about beta anywhere on the site except for in these component headers. Should we?

evwilkin:
Per discussion, keeping on hold for @nicolethoen to create a README around Beta components that we'll point to from the beta alerts.

evwilkin:
@nicolethoen this is updated with the link to your readme, lmk what you think - thanks!

nicolethoen:
so - if you had kept the link the same, it'd still land in a place that displayed the README and allowed them to look at the directory with the highlights. Would that be better?

evwilkin:
> so - if you had kept the link the same, it'd still land in a place that displayed the README and allowed them to look at the directory with the highlights. Would that be better?

Ahh that's a good point, thanks for the suggestion - I'll revert that link!

evwilkin:
@nicolethoen this is now updated, thanks 👍

nicolethoen (approved):

Issue #2765, opened 2021-12-06
PR preview action failing


mcoker:
re: https://github.com/patternfly/patternfly-org/pull/2764

the error is

```
Run mv build/patternfly-org/v3 tmp
mv: cannot stat ‘build/patternfly-org/v3’: No such file or directory
Error: Process completed with exit code 1.
```

Looks like the v3 build doesn't happen based on a conditional, but this command always runs. If the v3 build doesn't run, we shouldn't run this `mv` command.

PR #2766, opened 2021-12-06
fix(pr-preview): updated failing job to run conditionally

mcoker:
fixes https://github.com/patternfly/patternfly-org/issues/2765


evwilkin (changes requested):
Can you move this update to [.github/workflows-src/pr-preview.yml](https://github.com/patternfly/patternfly-org/blob/main/.github/workflows-src/pr-preview.yml#L21) - it looks like that file is run through [generate-workflows](https://github.com/patternfly/patternfly-org/blob/main/.github/generate-workflows.js#L11) which creates this file.

mcoker:
@evwilkin good catch! Updated and reran the generate workflows script.

PR #2800, opened 2022-01-07
Numbers audit

wolfeallison:
Audited the components to make sure all of the numbers used a symbol instead of being written out


mmenestr (approved):
Looks great @wolfeallison !! Thanks for taking the time to do this, and looks like there are no issues with your PR, yay!

PR #2816, opened 2022-01-19
fix: TS to JS conversion - Ability to spread props within components & allow for empty tags (fragments)

### LINKED ISSUE #2815: Unable to spread props inside components or use empty tags for fragments within markdown examples
**Describe the issue. What is the expected and unexpected behavior?**
When writing markdown examples, when spreading props into components or using empty tags for fragments, the TS to JS conversions (using the ExampleToolbar) fail with either console and/or syntax errors. This is not expected to happen since both of these things are valid JSX regardless of whether we're in a typescript or javascript space.

jeffpuzzo:
Closes #2815

### DIFF
--- a/packages/ast-helpers/stringify.js
+++ b/packages/ast-helpers/stringify.js
@@ -212,7 +212,7 @@ const es2017GeneratorJSX = {
   },
   // {...props}
   JSXSpreadAttribute(node, state) {
-    state.write('...(');
+    state.write(' {...(');
     if (node.argument.type === 'LogicalExpression') {
       this[node.argument.left.type](node.argument.left, state);
       state.write(' ');
@@ -223,7 +223,7 @@ const es2017GeneratorJSX = {
     else {
       this[node.argument.type](node.argument, state);
     }
-    state.write(')');
+    state.write(')}');
   },
   // <></>
   JSXFragment(node, state) {
@@ -232,8 +232,17 @@ const es2017GeneratorJSX = {
         attributes: [],
         name: {
           type: 'JSXIdentifier',
-          name: 'React.Fragment'
-        }
+          name: ''
+        },
+        type: 'JSXOpeningElement'
+      },
+      closingElement: {
+        attributes: [],
+        name: {
+          type: 'JSXIdentifier',
+          name: ''
+        },
+        type: 'JSXClosingElement'
       },
       children: node.children
     }, state);
--- a/packages/ast-helpers/stringify.js
+++ b/packages/ast-helpers/stringify.js
@@ -214,11 +214,11 @@ const es2017GeneratorJSX = {
   JSXSpreadAttribute(node, state) {
     state.write(' {...');
     if (node.argument.type === 'LogicalExpression') {
-      this[node.argument.left.type](node.argument.left, state);
       state.write('(');
-      state.write(node.argument.operator);
-      state.write(')');
+      this[node.argument.left.type](node.argument.left, state);
+      state.write(` ${node.argument.operator} `);
       this[node.argument.right.type](node.argument.right, state);
+      state.write(')');
     }
     else {
       this[node.argument.type](node.argument, state);
--- a/packages/ast-helpers/stringify.js
+++ b/packages/ast-helpers/stringify.js
@@ -212,18 +212,18 @@ const es2017GeneratorJSX = {
   },
   // {...props}
   JSXSpreadAttribute(node, state) {
-    state.write(' {...(');
+    state.write(' {...');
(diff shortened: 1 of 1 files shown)


packages/ast-helpers/stringify.js
```diff
@@ -212,7 +212,7 @@ const es2017GeneratorJSX = {
   },
   // {...props}
   JSXSpreadAttribute(node, state) {
-    state.write('...(');
+    state.write(' {...(');
```
nicolethoen, line 215: is it possible that sometimes it needs the curly braces and sometime it doesn't?

jeffpuzzo: This part of the conversion is specifically for attribute spreading within JSX. I can't think of an example where we would not need the curly braces for this circumstance. I did test that spreading elsewhere in JSX does not enter this block of code, where say if you had something like <div>{[...someArray, ...someOtherArray]}</div>, this is not impacted.

packages/ast-helpers/stringify.js
```diff
@@ -232,8 +232,17 @@ const es2017GeneratorJSX = {
         attributes: [],
         name: {
           type: 'JSXIdentifier',
-          name: 'React.Fragment'
```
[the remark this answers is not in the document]
jeffpuzzo: I did try both, and both are still working. There is an additional block for handling `JSXFragment`s where `React.Fragment` is handled already; https://github.com/patternfly/patternfly-org/blob/8da725b01f35fa2f1e8534f70cb05b296bb5b213/packages/ast-helpers/stringify.js#L129

nicolethoen (approved):
I tested this in my react workspace - it LGTM

mturley (dismissed):
This AST stuff is largely dark magic to me, so I hesitate to say I'm 100% confident in any changes here (when this code was originally written, was it just never tested? in what circumstances would it have worked correctly without the curly braces I wonder?)

But, I tested a bunch of variants of spread syntax and they all seem to work fine for me, so ¯\\\_(ツ)_/¯ LGTM

packages/ast-helpers/stringify.js
DaoDaoNoCode, line 215: Would it be better to put the `(` here inside the if statement below? like:
```
state.write(' {...');
if (node.argument.type === 'LogicalExpression') {
  state.write('(');
  ......
  state.write(')');
}
......
state.write('}');

We only need the brace when there is a logic expression like `&&` or `||`, and in this way, the converted JS code will look better.

**Review by DaoDaoNoCode — COMMENTED** [review]
Leave a small comment here. Otherwise, all the fixes look good to me, nice job! @jeffpuzzo

File: packages/ast-helpers/stringify.js [hunk 3]
```diff
@@ -212,18 +212,18 @@ const es2017GeneratorJSX = {
   },
   // {...props}
   JSXSpreadAttribute(node, state) {
-    state.write('...(');
+    state.write(' {...');
     if (node.argument.type === 'LogicalExpression') {
       this[node.argument.left.type](node.argument.left, state);
-      state.write(' ');
+      state.write('(');
```
DaoDaoNoCode, line 218: Sorry for not explaining clearly. The `state.write('(');` should be above `this[node.argument.left.type](node.argument.left, state);` and the `state.write(')');` should be below `this[node.argument.right.type](node.argument.right, state);`, so it will wrap the whole expression in this way. And we should better keep these 2 `state.write(' ');` here because we still need the space between the props and the logical operator. So the whole if statement should be like:
```
if (node.argument.type === 'LogicalExpression') {
  state.write('(');
  this[node.argument.left.type](node.argument.left, state);
  state.write(' ');
  state.write(node.argument.operator);
  state.write(' ');
  this[node.argument.right.type](node.argument.right, state);
  state.write(')');
}
```
Or you can even combine the center 3 `state.write` together to be like:
```
if (node.argument.type === 'LogicalExpression') {
  state.write('(');
  this[node.argument.left.type](node.argument.left, state);
  state.write(` ${node.argument.operator} `);    // or state.write(' ' + node.argument.operator + ' '); 
  this[node.argument.right.type](node.argument.right, state);
  state.write(')');
}
```
The situation in this if statement is not very commonly seen, many generators don't even consider it, but it's great to keep it here. Hope this makes sense 🙂

DaoDaoNoCode (approved):
LGTM, nice job, thanks! 😁 @jeffpuzzo

nicolethoen (approved):

mturley (approved):

evwilkin (approved):
LGTM - thanks!

PR #2820, opened 2022-01-20
Moves masthead guidelines from Page to Masthead

### LINKED ISSUE #2818: Move masthead design guidance from Page component to new Masthead component
Masthead guidelines were included within the Page component guidelines before the new Masthead component existed. Propose to move this section into a new guidelines doc that lives with the Masthead component.

cc @mmenestr

mcarrano:
This PR simply removes the guidance on the masthead from the Page component and adds it to the Masthead component. I added a link from Page to Masthead to reference this. At the time that the Page guidelines were created, the Masthead component did not exist.

Closes #2818


mcarrano:
@mmenestr can you please review when you get a chance? What I've done is to remove the masthead guidelines from Page and add them to the new Masthead component. I added a link from the Page guidelines doc. Also updated images on both pages to switch the order of the avatar and user menu per our most recent guidance and implementation. Thanks.

mmenestr:
@mcarrano this looks good! Just have two comments:
1. Change the first line to say "A masthead" instead of "**the** masthead" to match the other component guidelines
2. I notice that in all the images, the help icon is unfilled. Would you mind editing that icon to use a filled question circle, instead of an outlined one?

mcarrano:
@mmenestr I will make the first update. But regarding the help icon. We discussed this a while back and I believe we settled on the unfilled version of the Help icon as the one to use. @mceledonia can you confirm that?

mceledonia:
Looks like most products are using a filled icon here, but we have an unfilled version shown in the masthead design guidelines. However, in the updated icon guidelines, we recommend the product approach with filled in the masthead and outlined for contextual help. 

I prefer this approach, I think filled has more visual balance with the other filled icons in the masthead and the outlined contextual version prevents things from getting too heavy when there are a lot of them occurring in a single page. 

image

mmenestr (approved):

mmenestr:
Merging - dealing with images later when we have more discussion around it.

PR #2823, opened 2022-01-24
Drag and drop design conventions

andrew-ronaldson:
This PR is for the design conventions of the drag and drop feature available in PatternFly. The mockups included show how drag and drop is used in simple lists, multiple lists and cards.

Sketch file available at https://www.sketch.com/s/e12ae663-fe50-4265-9c62-18bff6f50c61


packages/v4/src/content/design-guidelines/components/drag-and-drop/drag.md
```diff
@@ -0,0 +1,72 @@
...
+
+<img src="./img/Simple-list-drag-2-3.png" alt="Borders appear around the item and bounding box shows the area in which an item can be dragged."/>
+
+4. postDrag event: Once dropped, the items will be reordered based on the user’s action. The space left empty is then filled by the next item in the list.
+
+<img src="./img/Simple-list-drag-4.png" alt="postDrag drops item into new position and all border highlights are disabled."/>
+
+5. Error state: If the list item is dragged outside the bounding box the borders on the dragged item will switch to `$pf-color-red-100` and the cursor will change to an error state to indicate an invalid placement. If the user releases the cursor outside the bounding area the dragged item will return to its default position.
+
+<img src="./img/simplelist_5.png" alt="Dragging outside the bounding box shows an error state on the dragged item." width="617"/>
```
evwilkin, line 25: There doesn't appear to be any file named "simplelist_5.png"

andrew-ronaldson: Thanks for catching that. Just updated the filename

packages/v4/src/content/design-guidelines/components/drag-and-drop/drag.md
```diff
@@ -0,0 +1,72 @@
+---
+id: Drag and drop
+section: components
+---
+The **drag and drop** interaction can be used to reposition elements on screen into a layout that benefits the user. This gives the user more flexibility to arrange and/or group items without having to make code changes.
+
+## Usage
+### Simple list
+1. Draggable area: The drag and drop interaction can be triggered with click and hold over any part of the item. The fa-grip icon is used to show that the drag and drop interaction is available.
```
mmenestr, line 9: Could you bold every first few words before the semi colon when you have numbered lists?

So` 1. **Draggable area**: the drag....`

mmenestr: Also might be helpful to use the "code" styling for "fa-grip", so '&fa-grip'

packages/v4/src/content/design-guidelines/components/drag-and-drop/drag.md
```diff
@@ -0,0 +1,72 @@
...
+
+<img src="./img/Simple-list-drag-1.png" alt="Showing the area in which an item can be dragged."/>
+
+2. Bounding box: Upon click & hold a `$pf-color-blue-400` border will show the draggable area that is available.
```
mmenestr, line 13: Do css values like that usually start with a $?

packages/v4/src/content/design-guidelines/components/drag-and-drop/drag.md
```diff
@@ -0,0 +1,72 @@
...
+### Simple list
+1. Draggable area: The drag and drop interaction can be triggered with click and hold over any part of the item. The fa-grip icon is used to show that the drag and drop interaction is available.
+
+<img src="./img/Simple-list-drag-1.png" alt="Showing the area in which an item can be dragged."/>
+
+2. Bounding box: Upon click & hold a `$pf-color-blue-400` border will show the draggable area that is available. 
+
+3. onDrag event: The list item being dragged will also use a `$pf-color-blue-400` border to highlight it as the item being dragged and all other list items will switch to a disabled state. The space where the item is being dragged from will remain empty to indicate its original position in the list.
+
+<img src="./img/Simple-list-drag-2-3.png" alt="Borders appear around the item and bounding box shows the area in which an item can be dragged."/>
```
mmenestr, line 17: Could you add a width to all these images so they're less blown up on the screen/ The width should be the width in the artboard in sketch.

So you'd just add width="123"  to the image tag

packages/v4/src/content/design-guidelines/components/drag-and-drop/drag.md
```diff
@@ -0,0 +1,72 @@
...
+3. **onDrag event:** The list item being dragged will also use a `pf-color-blue-400` border to highlight it as the item being dragged and all other list items will switch to a disabled state. The space where the item is being dragged from will remain empty to indicate its original position in the list.
+
+<img src="./img/Simple-list-drag-2-3.png" width="683" alt="Borders appear around the item and bounding box shows the area in which an item can be dragged."/>
+
+4. **postDrag event:** Once dropped, the items will be reordered based on the user’s action. The space left empty is then filled by the next item in the list.
+
+<img src="./img/Simple-list-drag-4.png" width="588" alt="postDrag drops item into new position and all border highlights are disabled."/>
+
+5. **Error state:** If the list item is dragged outside the bounding box the borders on the dragged item will switch to `pf-color-red-100` and the cursor will change to an error state to indicate an invalid placement. If the user releases the cursor outside the bounding area the dragged item will return to its default position.
+a
```
mmenestr, line 24: typo thing here to delete. Also make sure there's a "page break" / blank line, between the bullet list and image

andrew-ronaldson:
just adding @nicolethoen to the review. We updated designs for the error states that are not captured on the React side. Should I file another issue for that?

packages/v4/src/content/design-guidelines/components/drag-and-drop/drag.md
```diff
@@ -0,0 +1,72 @@
...
+---
+The **drag and drop** interaction can be used to reposition elements on screen into a layout that benefits the user. This gives the user more flexibility to arrange and/or group items without having to make code changes.
+
+## Usage
+### Simple list
+1. **Draggable area:** The drag and drop interaction can be triggered with click and hold over any part of the item. The `fa-grip` icon is used to show that the drag and drop interaction is available.
+
+<img src="./img/Simple-list-drag-1.png" width="588" alt="Showing the area in which an item can be dragged."/>
+
+2. **Bounding box:** Upon click & hold a `pf-color-blue-400` border will show the draggable area that is available.
```
mcoker, line 13: https://github.com/patternfly/patternfly/blob/541f89d94d4921d5512e02b30f3e09ac01437281/src/patternfly/components/DragDrop/drag-drop.scss#L6

I would reference the CSS var here, instead of the sass var name, which is just equal to a hex value. That implies the border should always be whatever `--pf-global--active-color--100` is, including if we change our global active color from `blue-400` to `blue-300`, for example. It also works across light and dark theme (and other themes, presumably)

wdyt @mcarrano @mceledonia @mmenestr? Do we have a standard for what kind of value/var to reference in places like this?

```suggestion
2. **Bounding box:** Upon click & hold a `--pf-global--active-color--100` border will show the draggable area that is available. 
```

packages/v4/src/content/design-guidelines/components/drag-and-drop/drag.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Usage
+### Simple list
+1. **Draggable area:** The drag and drop interaction can be triggered with click and hold over any part of the item. The `fa-grip` icon is used to show that the drag and drop interaction is available.
+
+<img src="./img/Simple-list-drag-1.png" width="588" alt="Showing the area in which an item can be dragged."/>
+
+2. **Bounding box:** Upon click & hold a `pf-color-blue-400` border will show the draggable area that is available. 
+
+3. **onDrag event:** The list item being dragged will also use a `pf-color-blue-400` border to highlight it as the item being dragged and all other list items will switch to a disabled state. The space where the item is being dragged from will remain empty to indicate its original position in the list.
```
mcoker, line 15:
```suggestion
3. **onDrag event:** The list item being dragged will also use a `--pf-global--active-color--100` border to highlight it as the item being dragged and all other list items will switch to a disabled state. The space where the item is being dragged from will remain empty to indicate its original position in the list.
```

packages/v4/src/content/design-guidelines/components/drag-and-drop/drag.md
```diff
@@ -0,0 +1,72 @@
...
+
+<img src="./img/Simple-list-drag-2-3.png" width="683" alt="Borders appear around the item and bounding box shows the area in which an item can be dragged."/>
+
+4. **postDrag event:** Once dropped, the items will be reordered based on the user’s action. The space left empty is then filled by the next item in the list.
+
+<img src="./img/Simple-list-drag-4.png" width="588" alt="postDrag drops item into new position and all border highlights are disabled."/>
+
+5. **Error state:** If the list item is dragged outside the bounding box the borders on the dragged item will switch to `pf-color-red-100` and the cursor will change to an error state to indicate an invalid placement. If the user releases the cursor outside the bounding area the dragged item will return to its default position.
```
mcoker, line 23:
```suggestion
5. **Error state:** If the list item is dragged outside the bounding box the borders on the dragged item will switch to `--pf-global--danger-color--100` and the cursor will change to an error state to indicate an invalid placement. If the user releases the cursor outside the bounding area the dragged item will return to its default position.
```

packages/v4/src/content/design-guidelines/components/drag-and-drop/drag.md
```diff
@@ -0,0 +1,72 @@
...
+
+<img src="./img/Simple-list-drag-1.png" width="588" alt="Showing the area in which an item can be dragged."/>
+
+2. **Bounding box:** Upon click & hold a `pf-color-blue-400` border will show the draggable area that is available. 
+
+3. **onDrag event:** The list item being dragged will also use a `pf-color-blue-400` border to highlight it as the item being dragged and all other list items will switch to a disabled state. The space where the item is being dragged from will remain empty to indicate its original position in the list.
+
+<img src="./img/Simple-list-drag-2-3.png" width="683" alt="Borders appear around the item and bounding box shows the area in which an item can be dragged."/>
+
+4. **postDrag event:** Once dropped, the items will be reordered based on the user’s action. The space left empty is then filled by the next item in the list.
```
mcoker, line 19: On the event names, do we want to give the event name or a more descriptive phrase? And if we use the event name, do we want to try and pair with the HTML drag/drop API or something else? https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API

packages/v4/src/content/design-guidelines/components/drag-and-drop/drag.md
mcoker: "Error" state is kind of generic. It works, but FWIW we use a modifier of `.pf-m-drag-outside` to apply the styling for when you're dragging over an area that isn't droppable.

packages/v4/src/content/design-guidelines/components/drag-and-drop/drag.md
mmenestr: Also just checked what we have in the sketch library, and we use the css var there, so I think it definitely makes sense to use it here too! Good catch

mmenestr (approved):
Merging what we have - can address any needed changes later!

PR #2828, opened 2022-01-26
feat(docs): added click event tracking

### LINKED ISSUE #2753: Set up GA events on pf.org
Follow up to https://github.com/patternfly/patternfly-org/issues/2721
Depends on #2795 

Set up [events](https://support.google.com/analytics/answer/1033068?hl=en#zippy=%2Cin-this-article) to track user interactions across the website.  Some identified events:
- clicks on component page tabs
- clicks on example actions
- clicks on or scroll to props/documentation/css variables
- clicks on jump links
- clicks to expand sidenav sections
- others?

evwilkin:
Closes #2753 

This PR implements the `gtag` functionality added in #2795 to add GA click event tracking to the follow events on patternfly.org:

1. Clicks on tabs at the top of pages (ex: React/HTML/Accessibility tabs on component pages)
    - **name**: `tab_click`
    - **data**: tracks name of tab clicked

2. Clicks on code editor controls in example code
    - **name**: `code_editor_control_click`
    - **data**: tracks which control user clicked:
      - `TOGGLE_CODE`
        - when code expanded/collapsed
        - **value**: value of `0` for collapsed and `1` for expanded
      - `COPY_CODE`
      - `CODESANDBOX_LINK`
      - `FULLSCREEN_LINK`
      - `TS_TO_JS`
        - for examples written in Typescript (Table is a good example)
      - `RESET_CODE`
        - (button only appears when example code is edited)

3. Clicks on jump links
    - **name**: `jump_link_click` with uppercase name 
    - **data**: tracks ID of element the jump link scrolls to (uppercase)

4. Clicks to expand/collapse sidenav sections
    - **name**:`sidenav_section_click`
    - **data**: Tracks name of section clicked
    - **value**: value `0` when section is collapsed, `1` when section is expanded

5. Clicks on the "view source" link at the bottom of any page
    - **name**: `view_source_click`
    - **data**: This varies - will show the `HTML/React/Accessibility` tab if on component page, `DESIGN-GUIDELINES` in UX writing section.  Essentially the "source" that the data comes from.

Note that `Page Title` (ex: About modal) and `Page` (ex: /components/about-modal/html) are already tracked for these events in Google Analytics, so that we can see where on the site each event occurs.

TESTING:

<details>
<summary>Testing instructions</summary>

For testing: local dev environment does not trigger actions, but the PR preview link (below) does.  Open the `Network` tab & type in `collect` in the filter box (as seen in screenshot below) to filter out extra noise.  As you click each item listed above, click the new network call that appears and look in the `Payload` tab for the following info:
- `ec` should list `click_event`
- `ea` should show the `name` of the event you triggered
- `el` should show the `data` tracked for that event
- `ev` will only show if there is a `value` tracked, for both the `TOGGLE_CODE` and `sidenav_section_click` events.  This will show `0` when the item is collapsed and `1` when the item is expanded.
Testing note: additional network calls will show up which automatically track page views, the URL for those will still start with `collect?` but will not include the `t=event` seen below.

Screen Shot 2022-01-26 at 4 28 49 PM

</details>

### DIFF
--- a/packages/theme-patternfly-org/helpers/sendEvent.js
+++ b/packages/theme-patternfly-org/helpers/sendEvent.js
@@ -1,8 +1,0 @@
-export const sendEvent = (name, eventCategory, eventLabel, value) => {
-  gtag('event', name, {
-    'event_category': eventCategory,
-    'event_label': eventLabel,
-    // Optional non-negative integer
-    ...(value >= 0 && {value})
-  });
-};
--- a/packages/theme-patternfly-org/helpers/sendEvent.js
+++ b/packages/theme-patternfly-org/helpers/sendEvent.js
@@ -1,3 +1,8 @@
-export const sendEvent = (name, paramsObj) => {
-  gtag('event', name, paramsObj);
+export const sendEvent = (name, eventCategory, eventLabel, value) => {
+  gtag('event', name, {
+    'event_category': eventCategory,
+    'event_label': eventLabel,
+    // Optional non-negative integer
+    ...(value >= 0 && {value})
+  });
 };
--- a/packages/theme-patternfly-org/helpers/sendEvent.js
+++ b/packages/theme-patternfly-org/helpers/sendEvent.js
@@ -1,0 +1,3 @@
+export const sendEvent = (name, paramsObj) => {
+  gtag('event', name, paramsObj);
+};
--- a/packages/theme-patternfly-org/components/example/exampleToolbar.js
+++ b/packages/theme-patternfly-org/components/example/exampleToolbar.js
@@ -9,7 +9,7 @@ import CodeIcon from '@patternfly/react-icons/dist/esm/icons/code-icon';
 import AngleDoubleRightIcon from '@patternfly/react-icons/dist/esm/icons/angle-double-right-icon';
 import ReplyAllIcon from '@patternfly/react-icons/dist/esm/icons/reply-all-icon';
 import { copy } from '../../helpers';
-import { sendEvent } from '../../helpers';
+import { trackEvent, EventCategory } from '../../helpers';
 
 function getLanguage(lang) {
   if (lang === 'js') {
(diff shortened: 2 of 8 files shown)


packages/theme-patternfly-org/helpers/sendEvent.js
```diff
@@ -0,0 +1,8 @@
+export const sendEvent = (name, eventCategory, eventLabel, value) => {
```
jeffpuzzo, line 1: Just to play on the title of this PR, I wonder if this function makes more sense to be named `trackEvent`.

Also, a docstring would be useful here to describe each of the parameters.

jeffpuzzo: Also, is there any reason to not make this a typescript file so we can type each of the params as well?

jeffpuzzo:
Couple things that would be good to add here would be write some unit tests for `sendEvent` and to add `Enum`s for `name`,`eventCategory`, and eventLabel`. Or at least for those params that are repeatedly used in multiple locations.

wise-king-sullyman (commented):
I second Jeff's comments, but everything else looks good to me. Once those are addressed/discussed I'll be happy to approve.

packages/theme-patternfly-org/helpers/sendEvent.js
evwilkin: @jeffpuzzo thanks for the feedback!  I went ahead and renamed the function per your suggestion.

When converting to TS it broke the build, it looks like we don't actually have TS running in the Org repo right now.  That's outside of the scope of this issue, and of my expertise, so I've reverted it back to JS for now and updated the file to better document the function & params.

jeffpuzzo (approved):
Tested, and looks good to me! Something for other testers to keep in mind, and maybe something we document somewhere is that browser ad-blockers block the google analytics API. I had to disable mine to see the endpoints firing.

wise-king-sullyman (approved):
LGTM 🎉

packages/theme-patternfly-org/helpers/sendEvent.js
evwilkin: Good suggestion - created https://github.com/patternfly/patternfly-org/issues/2846 to track this 👍

PR #2839, opened 2022-02-07
chore(documentation): adds breakpoint information

srambach:
Add documentation to the developer resources about how the breakpoints are used.

### DIFF
--- a/packages/theme-patternfly-org/pages/global-css-variables.md
+++ b/packages/theme-patternfly-org/pages/global-css-variables.md
@@ -46,7 +46,7 @@ Where...
 
 PatternFly defines a number of standard breakpoints. These are always used as a min-width breakpoint; i.e. using the `-md` breakpoint would apply to everything at the `-md` width and wider.
 
-Breakpoint suffixes are used in utility classes and layouts as well as in many components. When available, `{-on-[breakpoint]}` will be shown as an available suffix in the usage section of the documentation. The breakpoint suffix is optional and if not specified, the base will be used. There are breakpoints for `sm`, `md`, `lg`, `xl`, and `2xl`. The values for the corresponding breakpoints are defined below in the global variables `—pf-global—breakpoint—[breakpoint]`.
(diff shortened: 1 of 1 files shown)


evwilkin (dismissed):
LGTM - thanks!

packages/theme-patternfly-org/pages/global-css-variables.md
```diff
@@ -42,6 +42,12 @@ Where...
 - `breakpoint` is a media query breakpoint such as `sm` for `$pf-global--breakpoint--xs`.
 - `pseudo-element` is either `before` or `after`.
 
+### Breakpoint variables and class suffixes
+
+PatternFly defines a number of standard breakpoints. These are always used as a min-width breakpoint; i.e. using the `-md` breakpoint would apply to everything at the `-md` width and wider.
```
mcoker, line 47: Dunno if it's worth noting but the table is the exception to this.

```suggestion
PatternFly defines a number of standard breakpoints. These are always used as a `min-width` breakpoint; i.e. using the `-md` breakpoint would apply to everything at the `-md` width and wider.
```

packages/theme-patternfly-org/pages/global-css-variables.md
```diff
@@ -42,6 +42,12 @@ Where...
...
+
+Breakpoint suffixes are used in utility classes and layouts as well as in many components. When available, `{-on-[breakpoint]}` will be shown as an available suffix in the usage section of the documentation. The breakpoint suffix is optional and if not specified, the base will be used. There are breakpoints for `sm`, `md`, `lg`, `xl`, and `2xl`. The values for the corresponding breakpoints are defined below in the global variables `—pf-global—breakpoint—[breakpoint]`.
```
mcoker, line 49:
```suggestion
Breakpoint suffixes are used in utility classes and layouts as well as in many components as a way to apply class styles at a specified breakpoint. When available, `{-on-[breakpoint]}` will be shown as an available suffix for the class in the usage section of the documentation. The breakpoint suffix is optional and if not specified, the base class will be used. There are breakpoints for `sm`, `md`, `lg`, `xl`, and `2xl`, and the values for the corresponding breakpoints are defined below in the global variables `—pf-global—breakpoint—[breakpoint]`.
```

mcoker (approved):
Looks great! A couple of comments.

packages/theme-patternfly-org/pages/global-css-variables.md
mattnolting: @srambach @mcoker the table is the exception bc we're changing the default presentational behavior and breaking the built in semantics by making it `display: grid`. It doesn't make sense to assign `display: table` to a `table`, so this is an appropriate application of `min-width` media query as the table resumes its normal behavior at whatever breakpoint.

mcoker (approved):
🚀

PR #2842, opened 2022-02-10
Committing password strength indicator design guidelines

pankajshivpuje:


packages/v4/src/content/design-guidelines/demos/password-strength/password-strength.md
```diff
@@ -0,0 +1,25 @@
...
+## Elements
+
+<img src="./img/password strength indicator weak.png" alt="password strength indicator weak"/>
+<br/>
+<img src="./img/password strength indicator medium.png" alt="password strength indicator medium"/>
+<br/>
+<img src="./img/password strength indicator strong.png" alt="password strength indicator strong"/>
+<br/>
+The password strength indicator is located on the top right corner of the input field. It displays the following strength types:
+- Weak
```
mmenestr, line 17: Could you change this to a numbered list to match the pointers in the images? so:

1. Weak
2. Medium
3. Strong

mmenestr (changes requested):
Hey @pankajshivpuje, this looks great, I left one comment, and then also have a comment about the images - they all look quite blurry. There are 2 things we do to make sure the images come out as sharp as possible:

1. Make sure you exported the images at 2x PNG in Sketch.
2. Add a width constraint to the image in the code so that they display in real size, and not 2 times their size. To do that, you just need to width="123" to your image tag so for example:

To know which width to input, you just need to check the width of the artboard for the image inside the sketch file! 

Ping me if this doesn't make sense hahah

mmenestr (approved):
Looks great @pankajshivpuje !! Merging

PR #2845, opened 2022-02-11
chore(docs): enabled algolia search analytics

### LINKED ISSUE #2754: Set up Algolia search term tracking through GA
Follow up to https://github.com/patternfly/patternfly-org/issues/2721

We should collect data on & [track search terms](https://www.algolia.com/doc/guides/getting-analytics/search-analytics/understand-reports/how-to/connect-search-with-google-analytics/) used on pf.org to better understand what users are looking for and if we can better map to relevant results or if there are gaps in features that we can look into.

evwilkin:
Closes #2754 

Rather than track searches through Google Analytics, this PR enables the default search analytics automatically tracked by Algolia & accessible through the user dashboard in our account.

No analytics were previously being tracked, but these have been enabled by simply updating our API key to the current key & providing the `appId` as well - this last step is not clearly laid out in the docsearch docs as a requirement for analytics but was key in getting this working.

Note: confirming it is okay for this API key to be publicly viewable, per Algolia:

> This is the public API key which can be safely used in your frontend code.This key is usable for search queries and it's also able to list the indices you've got access to.

Result showing test searches coming through:
Screen Shot 2022-02-11 at 11 24 14 AM

### DIFF
--- a/packages/v4/patternfly-docs.config.js
+++ b/packages/v4/patternfly-docs.config.js
@@ -3,7 +3,8 @@ module.exports = {
   pathPrefix: '/v4',
   googleAnalyticsID: 'UA-47523816-6',
   algolia: {
-    apiKey: '06941733239da4f8617d272cf2ed4d5c',
+    apiKey: 'a8fb1726b78594ff97a3418757514404',
+    appId: '79P4ZBH7A3',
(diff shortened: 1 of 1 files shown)


jschuler (approved):
LGTM

PR #2849, opened 2022-02-14
fix(docs): remove optional chain & breaking import

### LINKED ISSUE #2848: theme-patternfly-org bump won't build
See [https://github.com/patternfly/patternfly-react/runs/5186643938?check_suite_focus=true](https://github.com/patternfly/patternfly-react/runs/5186643938?check_suite_focus=true)

We need to both remove the optional chainer [here](https://github.com/patternfly/patternfly-org/pull/2828/files#diff-1be6808eec1786fee67671a6e5240b3cce95851792241caca56a37407f494fcbR69) and the import [here](https://github.com/patternfly/patternfly-org/pull/2795/files#diff-0551229a0942b8c39960bb3b6d1817d5019f5c3c467f38d184b1e7e4b50be3e3R8) (can replace with string)

evwilkin:
Closes #2848 

This PR:
- Removes the optional chain character `?.`
- Removes the import from outside of the package (replaces with string)

### DIFF
--- a/packages/theme-patternfly-org/components/sideNav/sideNav.js
+++ b/packages/theme-patternfly-org/components/sideNav/sideNav.js
@@ -66,7 +66,7 @@ export const SideNav = ({ groupedRoutes = {}, navItems = [] }) => {
                     className="ws-side-nav-group"
                     onClick={(event) => {
                       // Don't trigger for bubbled events from NavItems
-                      if (!event.target?.href) {
+                      if (!event.target.href) {
                         const isExpanded = event.currentTarget.classList.contains('pf-m-expanded');
                         // 1 === expand section, 0 === collapse section
                         trackEvent('sidenav_section_click', 'click_event', section, isExpanded ? 0 : 1);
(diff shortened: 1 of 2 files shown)


packages/theme-patternfly-org/components/sideNav/sideNav.js
```diff
@@ -66,7 +66,7 @@ export const SideNav = ({ groupedRoutes = {}, navItems = [] }) => {
                     className="ws-side-nav-group"
                     onClick={(event) => {
                       // Don't trigger for bubbled events from NavItems
-                      if (!event.target?.href) {
+                      if (!event.target.href) {
```
nicolethoen, line 69: I don't think these are equivalent. I think it should be `if (!event.target || !event.target.href)` for them to be equivalent. The `?` is preventing an exception being thrown if event.target is undefined.

evwilkin: @nicolethoen thanks for calling this out - I intentionally changed this, in reviewing I realized the `?.` wasn't really needed, as the `.href` property of `event.target` should always either be defined or return `undefined`

nicolethoen (approved):

PR #2864, opened 2022-03-03
fix: stop example widths from being wider than parent on mobile viewports

nicolethoen:
to test, make your viewport width very narrow.
before this change, notice the examples are spilling over to the right like in the screenshot below from the Table with actions example where the actions kabob is completely off screen to the right. After this change, the examples don't spill over.

before:
Screen Shot 2022-03-03 at 8 43 55 AM

### DIFF
--- a/packages/theme-patternfly-org/templates/mdx.css
+++ b/packages/theme-patternfly-org/templates/mdx.css
@@ -294,7 +294,7 @@
 
 .ws-mdx-content {
   flex-grow: 1;
-  width: 100%;
+  min-width: 0;
 }
 
 /* Design asked for this. A nice side effect is that the TOC doesn't "jump" between pages */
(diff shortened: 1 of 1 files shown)


mcarrano (dismissed):
Looks good to me!

evwilkin (approved):
LGTM - thanks!

packages/theme-patternfly-org/templates/mdx.css
```diff
@@ -294,6 +294,7 @@
 
 .ws-mdx-content {
   flex-grow: 1;
+  width: 100%;
```
mcoker, line 297: This works great, though the root issue is that the `min-width` of a flex child is `min-content` - the generally recommended way to fix that is to set `min-width: 0;`. 

```suggestion
  min-width: 0;
```

mcoker (approved):
🥳

PR #2867, opened 2022-03-08
chore(docs): added mobile jumplinks and example styling

### LINKED ISSUE #2843: Improvements to Component pages
The following are proposed updates to Component pages to make them more usable and readable. 

1) Make it possible to access the TOC after it hides from the right hand sidebar. This is a follow up to: https://github.com/patternfly/patternfly-design/issues/1096 Proposed design here: https://marvelapp.com/prototype/7bcja40

2) Update styling to better differentiate headings from example content. Proposed designs: https://marvelapp.com/prototype/h964d3h

evwilkin:
Closes #2843 

This PR:
- Unhides the jumplinks which are currently hidden at <=1450px and utilizes the component's mobile layout plus custom styles from designs in #2843 
- Removes background from example headers/controls
- Remove borders from CodeEditorControls below each example, except for the blue bottom border on hover/focus/active
- Remove border below `.pf-c-code-editor__header` above example code except when example code is expanded
- Simplifies copy function for example code to fix console error currently being thrown

### DIFF
--- a/packages/theme-patternfly-org/components/example/example.css
+++ b/packages/theme-patternfly-org/components/example/example.css
@@ -9,6 +9,12 @@
   border-bottom: var(--pf-global--BorderWidth--sm) solid #f8f8f8;
 }
 
+.ws-example > .ws-example-header,
+.pf-c-code-editor__header,
+#ws-react-c-card-modifiers > div:first-child {
+  background-color: var(--pf-global--BackgroundColor--200);
+}
+
 .ws-example-header > .ws-example-heading:not(:last-child) {
   margin-bottom: var(--pf-global--spacer--md);
 }
--- a/packages/theme-patternfly-org/components/example/example.css
+++ b/packages/theme-patternfly-org/components/example/example.css
@@ -9,6 +9,11 @@
   border-bottom: var(--pf-global--BorderWidth--sm) solid #f8f8f8;
 }
 
+.ws-example > .ws-example-header,
+.ws-example > .pf-c-code-editor > .pf-c-code-editor__header {
+  background-color: var(--pf-global--BackgroundColor--200);
+}
+
 .ws-example-header > .ws-example-heading:not(:last-child) {
   margin-bottom: var(--pf-global--spacer--md);
 }
--- a/packages/theme-patternfly-org/components/example/example.css
+++ b/packages/theme-patternfly-org/components/example/example.css
@@ -12,21 +12,24 @@
   margin-bottom: var(--pf-global--spacer--md);
 }
 
-.ws-code-editor > .pf-c-code-editor__header::before {
-  content: none;
+.ws-code-editor:not(.ws-example-code-expanded) > .pf-c-code-editor__header::before {
+  --pf-c-code-editor__header--before--BorderBottomWidth: 0;
 }
 
-.ws-code-editor-control::after {
-  display: none;
+.ws-code-editor-control {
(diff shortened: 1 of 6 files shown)


evwilkin:
@mcarrano @maryshak1996 a couple notes/questions on this PR:
1. The jumplinks component has a mobile layout ([demo](https://www.patternfly.org/v4/components/jump-links/react-demos/scrollspy-with-subsections/)) that seems to match the design, including being expandable without the need to remove an extra blue left border, so I've simply used the mobile layout of jumplinks with custom styling added rather than wrap it in the expandable disclosure.
2. Our examples are inconsistent when it comes to the modifiers - for example, the [card modifiers example](https://www.patternfly.org/v4/components/card#modifiers) includes these toggles above the example, while the [card with image and actions example](https://www.patternfly.org/v4/components/card#with-image-and-actions) includes a toggle after the example and [this tooltip example](https://www.patternfly.org/v4/components/jump-links/react-demos/scrollspy-with-subsections/) has a completely different look altogether.
    - Based on this, I'd suggest splitting this out into a bigger consideration for a common design/implementation to be used for these types of examples.
    - If we go this route, should we proceed with the example header/code editor controls having the grey background as seen, and leave the modifiers untouched for a follow-up issue?

evwilkin:
@mcarrano @maryshak1996 I've updated this PR per our discussion:
- Added bottom box shadow to mobile table of contents (large when collapsed, small when expanded)
- Examples header & code editor controls no longer have any background, match page color instead
- Code editor controls are now plain icon buttons with no borders/background

mcarrano:
@evwilkin thanks for making these updates. Overall I think this looks a whole lot better! I just have one question/nit. By making the tool buttons below the example plain buttons, they no longer display a selected state. You'll notice that in the existing site, when the code window is open the "<>JS" or "<>TS" button is shown as selected/active. I'm not sure that's a big deal, but I wanted to raise it as an issue. What do you think @maryshak1996 ?

In looking at this, I was also wondering how those are implemented now. Even though they function as buttons, they have dropdown styling that leads to them having a blue underline when open. Were these some type of custom button? Would we want to customize the plain button to give it a selected state?

evwilkin:
@mcarrano thanks for taking a look.  The controls beneath each example & the expandable code below them are the `CodeEditor` component [with custom controls](https://www.patternfly.org/v4/components/code-editor#with-custom-control).

The `CodeEditorControl`s output control variant buttons (`<Button variant="control">`), which apply the styling currently on our site.  My understanding was to remove all the borders on these buttons (which in turn removed the blue bottom border), but we could keep only that blue border if we'd like - I believe this would be a custom style for our site.

mcarrano:
@evwilkin thanks for the added info. Yes I do recall that we styled the code editor buttons that way. @maryshak1996 I'll defer to you, should we do any more customization of these buttons or do you think they are good as is?

maryshak1996:
@evwilkin @mcarrano if we could keep just the blue border to indicate active state, that would be ideal!

evwilkin:
@mcarrano @maryshak1996 per our conversation I've made 2 additional changes:
- removed the borders from the example CodeEditorControls but kept the blue bottom border on focus/hover/active
- Remove the bottom border beneath the examples' CodeEditorControls when the example code is collapsed.

mcarrano (approved):
This looks great @evwilkin . Thanks for making these updates!

maryshak1996 (approved):

PR #2869, opened 2022-03-09
fix(docs): fixed font awesome typos

### LINKED ISSUE #2832: Wrong documentation for using icons in getting started section
In https://github.com/patternfly/patternfly-org/blob/main/packages/v4/src/content/get-started/develop.md the Using Styles - Icons section appears to be a cut and paste of the Typography section, referring to Overpass. This should instead explain how the Font Awesome is included in Patternfly.

evwilkin:
Closes #2832 

This PR fixes typos in the getting started for developers page around font awesome implementation instructions.

### DIFF
--- a/packages/v4/src/content/get-started/develop.md
+++ b/packages/v4/src/content/get-started/develop.md
@@ -140,7 +140,7 @@ PatternFly uses Font Awesome 5. Font Awesome 5 can be utilized in two different 
 
   * **Used as a CDN**
 
-  If you wish to use the CDN for Font Awesome 5 rather than the default approach, you need to update the ```utilities/variables.scss``` file (from source ```node_modules/@patternfly/patternfly/```) and build PatternFly as part of your build process.
+  If you wish to use the CDN for Font Awesome 5 rather than the default approach, you need to update the ```sass-utilities/scss-variables.scss``` file (from source ```node_modules/@patternfly/patternfly/```) and build PatternFly as part of your build process.
 
   To use the CDN vs the standard build, update the ```sass-utilities/scss-variables.scss``` file as follows:
 
--- a/packages/v4/src/content/get-started/develop.md
+++ b/packages/v4/src/content/get-started/develop.md
@@ -114,21 +114,9 @@ Use these files to consume the library. The recommended consumption approach wil
 ### Using styles
 **Typography**
 
-PatternFly uses the Overpass font family. Overpass can be utilized in two different ways:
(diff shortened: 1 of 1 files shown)


evwilkin:
@mcarrano looping you in for this copy update - while both the typography & icons sections ([here](https://patternfly-org-pr-2869-v4.surge.sh/v4/get-started/develop#using-styles)) are nearly identical, causing the original issue, they actually are accurate as each calls out different sass files to update and different sass variables to change.  That said, there was a typo where "Overpass" was mentioned in the Font Awesome section which I've cleaned up.

mcarrano (changes requested):
@evwilkin the changes you made look fine, however I believe this statement is no longer true:

> PatternFly uses the Overpass font family. 

We changed the default to Red Hat font a while ago. Didn't we?

evwilkin:
@mcoker - it looks like we may be able to fix part of the Typography page that wasn't originally part of the scope of this issue.  We've switched to Red Hat Font but it looks like the option to load Overpass from CDN still exists (without an equivalent for Red Hat Font) - do you mind reviewing the Typography section here and helping determine what's best to communicate to users at this point?

- https://www.patternfly.org/v4/get-started/develop#using-styles

> @evwilkin the changes you made look fine, however I believe this statement is no longer true:
> 
> > PatternFly uses the Overpass font family.
> 
> We changed the default to Red Hat font a while ago. Didn't we?

mcoker:
@evwilkin @mcarrano I'm not entirely sure what to do with the "Using styles" section. Looking at it more, I think we could call out a lot of things here, not just how to import fonts and icons.

I don't believe we need to list the equivalent details about Red Hat font like we do Overpass. I've never been asked about serving the Red Hat font via CDN, either, so I don't think that's something we need to support now anyways.

If we want to keep something there for typography, we could mention that we use Red Hat font, maybe link to the repo for it - https://github.com/RedHatOfficial/RedHatFont. And possibly link to the typography guidelines page about the families, sizes, etc - https://www.patternfly.org/v4/guidelines/typography

And we could link to the Red Hat font page for some available opt-ins - https://www.patternfly.org/v4/developer-resources/red-hat-font. Though that's something we'll likely want to keep an eye on since the opt-ins will likely change in our next breaking change release and any details we include on this page will probably need to be updated.

mcarrano:
@evwilkin @mcoker I'll defer to you guys. Seems like this page is targeted at developers, so if this information is no longer relevant, then perhaps we remove it? I was just pointing out that the current information seems to be stale.

mcoker (approved):
Looks great!

Issue #1108, opened 2018-08-10
Tooltip - Design Doc


mcarrano:
Introduce design documentation for Tooltip.

Related issue: #113

jgiardino:
@matthewcarleton, @rachael-phillips and I talked about Tooltips in the a11y review meeting today. 

The predominant pattern for accessible tooltips is to use `aria-describedby` or `aria-labelledby` to associate the tooltip text with the element that has focus, but there are some gaps where this doesn't work, which are noted in https://github.com/patternfly/patternfly-react/issues/789. These gaps are mainly cases where tooltips are used with static elements.

The conclusion that we came to is that instead of trying to create a solution that addresses these accessibility gaps, we should instead define in the design documentation that using tooltips on static elements is not recommended, and include suggestions for alternate solutions.

I've listed below some ideas on what we could suggest instead. @matthewcarleton also mentioned some examples where his team recently revised similar designs to use an alternate solution, so it would be great to see what his team has come up with, too.

Alternate suggestions for tooltips on static elements:

- Display the tooltip contents inline next to the element. For example, an icon that represents status could have the status stated (specifically for status that is not obvious from the icon).
- Display the contents in a popover
- For recurring tooltips, especially when the same contents could be repeated in a tooltip more than 10 times, consider creating a legend (i.e. a single place where this information can be accessed). Creating additional tab stops for static elements creates an additional burden for keyboard-only users.
  - Note that this case often occurs where tooltips are used to describe icons, and that an accessible name for these icons is still needed for screen readers. The recommended method for these is to include screen-reader-only text adjacent to the icon element.

abigaeljamie:
If we decide to create a doc on this, can we please include these UX writing guidelines for tooltips?  

- Keep your tooltips clear and concise. Use the fewest number of words you can without sacrificing meaning. 
- If the tooltip is a full sentence, end it with a period.   
- If information is needed for a user to complete a task (like a password character requirement), don’t hide it in a tooltip. Display it on the page instead.

mcarrano:
@mmenestr Do you think it makes sense to just fold this into the embedded assistance document that you are currently working on? It already references tooltips.

mmenestr:
Yes, that definitely makes sense! Just added @abigaeljamie's writing guidelines in the doc.
Do you think we should also add anything about accessibility for tooltips in there or no?

Issue #2871, opened 2022-03-10
Update tooltip and popover design guidelines


mmenestr:
Will be changing the tooltip to only be used for icons
Will be changing the popover to work on hover AND click

Not implemented yet - so holding

cc: @abigaeljamie

mmenestr:
Still on hold - there is an issue somewhere about it but never worked on

PR #2892, opened 2022-03-31
docs(Tooltip): add a11y docs

### LINKED ISSUE #2596: Add a11y documentation for Tooltip

thatblindgeye:
Closes #2596

[Tooltip a11y tab](https://patternfly-org-pr-2892-v4.surge.sh/v4/components/tooltip/accessibility)

Current implementation for a11y docs on the tooltip. This is based off of the current changes made in https://github.com/patternfly/patternfly-react/pull/7332 and https://github.com/patternfly/patternfly-react/pull/7335.

Additionally, I added in some verbiage in regards to tooltips on static elements. In https://github.com/patternfly/patternfly-org/issues/1108#issuecomment-495262558 it was mentioned that tooltips on static elements shouldn't be recommended. However, utilizing the pattern of `aria-live` instead of `aria-describedby` does seem to resolve the gaps mentioned.

Original comment:

One proposal I want to put out there is whether we should include verbiage for when a tooltip *shouldn't* be announced to screen reader users. Specifically, if an element already has a label (such as via `aria-label`), and the tooltip contents is essentially the same thing, should the tooltip then be prevented from being announced to screen reader users to avoid redundant/duplicate announcements? Or would the better approach be to avoid adding an `aria-label` if a tooltip is to be attached (as we've seen with using React refs, tooltips may not always be announced, though I've noticed a possible fix could be using `aria-live` on tooltips attached via ref)?

Using the [Tooltip on icon example](https://www.patternfly.org/v4/components/tooltip#on-icon), if the example code was changed so that the button had `aria-label="Copy contents to clipboard"`, a screen reader could announce both that and the tooltip contents, which would be redundant information. This could also be trying to solve a problem that may not really exist or not be a significant issue, though.

### DIFF
--- a/packages/v4/patternfly-docs/content/accessibility/tooltip/tooltip.md
+++ b/packages/v4/patternfly-docs/content/accessibility/tooltip/tooltip.md
@@ -17,20 +17,19 @@ The preferred method of using the tooltip is by wrapping it around its trigger v
 
 To ensure the tooltip is used accessibly, follow these steps when applicable:
 
-- When adding a tooltip to a static element, such as a `span` or `div`, you must ensure:
-    - The static element has `tabindex="0"` passed in, so that keyboard users can place focus on it to trigger the tooltip, and there is some form of styling to visually let users know the element can be focused/hovered (usually an underline on the text).
-    - When using the `children` prop, `aria-live="polite"` and `aria="none"` are passed into the tooltip. Tooltips on static elements do not always get announced by assistive technologies by default, so passing in `aria-live` allows the tooltip contents to be announced while `aria="none"` ensures the tooltip does not get announced more than once.
-- When the tooltip content is expected or intended to dynamically update and the `children` prop is being used, you must also pass in `aria-live="polite"` and `aria="none"` to ensure the updated content gets announced to users and that the the tooltip does not get announced more than once.
-- When passing in the `id` prop and using the `children` prop you should also pass in `aria="none"`, and then either `aria-describedby` or `aria-labelledby` to the tooltips' trigger. The aria attribute passed into the trigger should have the tooltips' `id` prop passed in as a value.
+- Tooltips should not be added to static elements, such as a `div` or `span`.
+- When a tooltip should act as the primary label for an element, pass in `aria="labelledby"`. When a tooltip should act as supplementary information, keep the default `aria="describedby"`.
+    - This pattern should also be followed when manually passing in the `aria-labelledby` or `aria-describedby` attribute to a trigger.
+- When using the `reference` prop, you should also pass in the `id` prop and then manually pass in `aria-labelledby` or `aria-describedby` to the trigger. The aria attribute passed into the trigger should have the tooltip `id` passed in as a value.
+    - If you are instead using the `children` prop and you want to manually pass in `aria-describedby` or `aria-labelledby` to a trigger, you should also pass in `aria="none"` to the tooltip.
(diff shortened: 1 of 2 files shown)


nicolethoen (dismissed):

jessiehuff (dismissed):
Looks good! Thanks for working on this! 🙂

thatblindgeye:
@nicolethoen from the backlog meeting earlier, I noticed one issue regarding Popover and adding hover functionality. After digging through that, based on issue #2871 + conversation in the issue linked in it, do you think it would be better to update the a11y docs for Tooltips taking into account these intended changes to Tooltip + Popover? For the most part may just be a case of re-adding the recommendation to not use tooltips on static elements, and then later adding in any other verbiage the design guidelines end up including.

nicolethoen:
> For the most part may just be a case of re-adding the recommendation to not use tooltips on static elements, and then later adding in any other verbiage the design guidelines end up including.

I think I agree - we should include the recommendation of not using tooltips on static elements and then we can follow up if recommendations evolve. 
The only exception being if we already are following the 'yet to be written' recommendations in the examples/demos and then it would just mean that the design guidelines would be one step behind the a11y docs which is not the end of the world.

jessiehuff (commented):
I really love these updates! I think it's a good call out that you're differentiating labeling from supplementary information. 

I think Nicole makes a good suggestion as well. Adding small snippets/examples helps make the points clearer and more consumable.

thatblindgeye:
@nicolethoen @jessiehuff I updated the tooltip doc with the new template. I also added in some code blocks/examples for certain things throughout. For examples with specific props, they're not exactly inline in the table and instead placed in sub-sections, so let me know what you think of that sort of layout or if you have other ideas for how to go about them

jessiehuff (commented):
I like the new template for this, but I find the top section a little confusing between the React and HTML/CSS bullet points. At first I thought that you meant that everyone needs to add all of those things. For example, I saw the point about adding the role and thought, "isn't that already added for you?" Then I realized that you meant if you're just using the HTML/CSS, not that you need to add HTML/CSS. Admittedly I think previously I just included the React suggestions before and kept the HTML/CSS in the customization section. However, I can understand why it might make sense to keep at the top, but I'd be curious what others think. Do you think it would be helpful for us to separate that into two different categories at the top to keep it clear? Or maybe it was just me that got confused haha.

thatblindgeye:
>Or maybe it was just me that got confused haha.

I definitely think you raise a valid point. What do you think of something along the lines of this instead:

```
- Pass in `role="tooltip"` to the element acting as the tooltip component if using the HTML/CSS library
- Pass in `aria="labelledby"` to the tooltip component if using the PatternFly React library, or the `aria-labelledby` attribute to the trigger if using the HTML/CSS library, when the tooltip should act as the primary label for its trigger

or use nested bullets to separate the two libraries:

- Pass in `aria="labelledby"` to the tooltip component if using the PatternFly React library when the tooltip should act as the primary label for its trigger.
    - Pass in the `aria-labelledby` attribute to the trigger if using the HTML/CSS library
```

The "if using X library..." could also be placed first, e.g. "If using the HTML/CSS library, pass in `role="tooltip"`...".

nicolethoen:
I think I like your second suggestion more. maybe something like:

To implement an accessible PatternFly tooltip:
  - Avoid using tooltips on static elements such as a div or span, except in cases of truncation.

Using react:
- Pass in aria="labelledby" when the tooltip should act as the primary label for its trigger (with snippet)

Using HTML/CSS:
- Set the aria-labelledby attribute on the trigger when the tooltip should act as the primary label for its trigger. (code snippet)
- Pass in the aria-describedby attribute to the trigger (HTML/CSS) when the tooltip should act as supplementary information. (code snippet)

nicolethoen (approved):

Issue #2682, opened 2021-09-15
Add A11y docs for Progress Stepper component


nicolethoen:
Follow on for React issue: https://github.com/patternfly/patternfly-react/issues/6165

Issue #2814, opened 2022-01-18
Update badge a11y documentation


jessiehuff:
I think we should add a note to the badge a11y documentation about how the color could be important (i.e. if you're using colored badges to communicate something, you should find a way to communicate that to users with visual disabilities or screen reader users).

PR #2893, opened 2022-04-02
add/update component a11y docs

nicolethoen:
Closes:
#2682 
#2593 
#2824 
#2825 
#2814


packages/v4/src/content/developer-resources/accessibility-development.md
```diff
@@ -98,3 +98,19 @@ The recommended interaction pattern for the modal components like the modal or p
 - For screen reader users, the other contents on the page should be hidden from the screen reader.
 
 Based on the [screen reader / browser combinations](https://www.patternfly.org/v4/get-started/accessibility-guide#testing) we use for testing, we recommend applying `aria-hidden="true"` to the parent wrapping element of the page contents. Note that the modal element of the component must not be a descendent of this element with `aria-hidden="true"` and should be included as a sibling to this element.
+
+### Scrollable elements
+Any elements with horizontal or vertical scroll need to be accessible via keyboard. It may be necessary to ensure every 
+container with an overflow rule and content which could possibly grow larger than its parent's height to have 
+`tabindex=0`. This will ensure that parent can capture browser focus and be scrolled using the arrow keys. Some browsers
```
seanforyou23, line 105: When we add `tabIndex="0"` to an inactive element, like a div container for example, we turn this element into an interactive element - for interactive elements, it's important to ensure a well crafted accessible name exist.

If we only add tabIndex, and do not also ensure an accessible name is explicitly set, screen readers may not treat this content as expected

The main issue is that the container element with non-negative tabIndex will now be registered as an element in the form controls rotor menu, but the label for this element will be driven by whatever text content exist inside the container, which is variable and probably not suitable to use as a single label to name/describe the region's purpose.

That said, even if you're just using standard keyboard nav or screen reader cursor nav (and not rotor) the effect is the same.

Could we just also mention that when adding tabIndex in this scenario, it would also be good to add an aria-label that describes what the element is? I think that would do it..

packages/v4/src/content/accessibility/badge/badge.md
```diff
@@ -10,3 +10,7 @@ A **badge** is used to annotate other information like a label or an object name
 **Screen reader users** should be able to have a screen reader describe the contents of and purpose of a badge when it pops up, but not focus on it otherwise.
 
 No props need to be added or modified for badge accessibility.
+
+As a caution, badges can be styled using a variety of colors. Relying on color alone to communicate information causes barriers to
```
seanforyou23, line 14: I wonder if it's worth making a simple suggestion of what to do instead, in case users were planning to do this "rely on color alone"...

Since the badge doesn't get an accessible name and isn't focusable, maybe we can just suggest to ensure there is adequate contextual information provided in the surrounding UI parts to convey the unit of measure for the number in the badge in case it isn't obvious?

Or maybe could just mention to consider using [label with icon](https://patternfly-org-pr-2893-v4.surge.sh/v4/components/label) if the badge count alone isn't intuitive enough?

seanforyou23 (commented):
Good stuff! Love these doc enhancements :)

Left a few questions/comments, lemme know what you think!

packages/v4/src/content/accessibility/jump-links/jump-links.md
```diff
@@ -0,0 +1,26 @@
...
+- Keyboard users should be able to navigate to and through the jump links using **Tab** and **Shift + Tab**. 
+- Keyboard users should also be able to select a jump link using **Enter** and **Space**.
+- The jump links should be before the associated sections in the DOM order so that the user can use the jump links to jump to the desired section. 
+- When a user selects a jump link, the browser focus should move into the associated section so that the navigable elements in that associated section
+can be navigated to next without having to move focus through all the rest of the jump links and page.
+
+Additionally, it should be noted that jump links are often used on pages with so much content that the page scrolls. It should be true that
+**Up** and **Down** arrow keys should be able to scroll the content.
+
+**Screen reader users** should be made aware that the jump links are interactable, and which jump link is currently active using
```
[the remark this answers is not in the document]
thatblindgeye: It looks like the links in the Jump Links page aren't in VO rotor due to not having an `href` passed in. Once that gets passed in they show up. Though via the rotor there's no indication which is active, so adding the verbiage regarding "on focus" I'd agree could be worth adding for that.

packages/v4/src/content/accessibility/jump-links/jump-links.md
```diff
@@ -0,0 +1,26 @@
...
+**Up** and **Down** arrow keys should be able to scroll the content.
+
+**Screen reader users** should be made aware that the jump links are interactable, and which jump link is currently active using 
+`aria-current="location"` on the active jump link. 
+
+The following props/attributes have been added for you or are customizable in PatternFly:
+
+| React component | React prop      | Which HTML element it appears on in markup   | Explanation                                                   | 
+|-----------------|-----------------|----------------------------------------------|---------------------------------------------------------------|
+| JumpLinks       | aria-label      | nav.pf-c-jump-links aria-label               | Labels the nav to reflect the current nav description.        |
```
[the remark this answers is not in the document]
thatblindgeye: I'd be in agreement on this as well.

packages/v4/src/content/accessibility/label/label.md
```diff
@@ -14,3 +14,7 @@ The following props/attributes have been added for you or are customizable in Pa
 | React component | React prop | Which HTML element it appears on in markup | Explanation | 
 |---|---|---|---|
 | Label | closeBtnProps | .pf-c-label | Customize the props on the close button of the label. For example, to change the `aria-label`, pass `closeBtnProps={{'aria-label': 'new label'}}` to the Label component. |
+
+As a caution, labels can be styled using a variety of colors. Relying on color alone to communicate information causes barriers to
```
thatblindgeye, line 18: Just wanted to highlight this to go along with the comments from the Badge file as they might pertain here as well.

packages/v4/src/content/accessibility/progress-stepper/progress-stepper.md
```diff
@@ -0,0 +1,19 @@
...
+id: Progress stepper
+section: components
+---
+
+A **progress stepper** is used to provide information to the user about the state of their current progress on a multistep interaction.
+
+**Keyboard users** should be able to use **Tab** and **Shift + Tab** navigate to and between progress steps with popovers. Keyboard users should 
+be able to open and close the popover on a progress step using **Enter** or **Space**.
+
+**Screen reader users** should be able to discern the label, status, and optional description of every progress step.
```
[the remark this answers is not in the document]
thatblindgeye: For NVDA/JAWS it might just be called "elements list"? Based on this screen shot of keyboard shortcuts from Deque, the task in the second column aligns with what the rotor does in VO (the first two columns are the same for their pages on NVDA and JAWS)

NVDA elements list shortcut

That said I think I'd agree in maybe explicitly stating in what way the info should be discernible via screen reader, re: screen reader/VO focus vs rotor/element list. This might involve doing a run-through of current a11y docs as well, but could be easier at this stage.

packages/v4/src/content/accessibility/progress-stepper/progress-stepper.md
```diff
@@ -0,0 +1,19 @@
...
+---
+
+A **progress stepper** is used to provide information to the user about the state of their current progress on a multistep interaction.
+
+**Keyboard users** should be able to use **Tab** and **Shift + Tab** navigate to and between progress steps with popovers. Keyboard users should 
+be able to open and close the popover on a progress step using **Enter** or **Space**.
+
+**Screen reader users** should be able to discern the label, status, and optional description of every progress step. 
+The aria-label provided to each step should indicate all information being communicated by the progress step's icon, including the variant and the completed status.
+Screen reader users should also be able to discern whether a step has an associated popover they can trigger and the contents of that popover.
```
thatblindgeye, line 13: The previous line is also tied into this comment, but I wanted to keep this separate from the comments already shared on that one.

Could lines 11-13 of the file for screen reader users be linked more to a single thought, rather than sort of coming off as somewhat separate ones? Roughly something along the lines of:

```suggestion
**Screen reader users** should be able to discern the label, status, and optional description of every progress step via the `aria-label` prop. This should include all information being communicated by the progress step's icon, such as the variant and the completed status, as well as whether the step has an associated popover they can trigger. Screen reader users should also be able to access the contents of any popover associated with a step.
```

Of course taking into account if any changes are made based on the comments above.

packages/v4/src/content/accessibility/tabs/tabs.md
```diff
@@ -0,0 +1,27 @@
...
+id: Tabs
+section: components
+---
+
+**Tabs** allow users to navigate between views within the same page or context.
+
+**Keyboard users** should be able to use **Tab** and **Shift + Tab** navigate to and between tabs.
+Keyboard users should also be able to select tab using **Enter** and **Space**. Disabled tabs cannot receive
+browser focus. Whenever a disabled tab needs to be able to receive browser focus, such as in the case of
+disabled tabs with a tooltip, set `aria-disabled` to true.
```
thatblindgeye, line 11: Minor suggested tweaks here.

```suggestion
Keyboard users should also be able to select a tab using **Enter** and **Space**. Disabled tabs cannot receive
browser focus by default. Whenever a disabled tab needs to be able to receive browser focus, such as in the case of
disabled tabs with a tooltip, use the `isAriaDisabled` prop instead of `isDisabled`.
```

thatblindgeye (commented):
Awesome job on these. I think I mainly had comments on several things, with a few suggested changes sprinkled in.

packages/v4/src/content/accessibility/jump-links/jump-links.md
```diff
@@ -0,0 +1,26 @@
+---
+id: Jump links
+section: components
+---
+
+**Jump links** provide a way to navigate through section headings on a page without scrolling.
+
+For **Keyboard users** there are a number of considerations which must be made concerning tab order and focus management.
+- Keyboard users should be able to navigate to and through the jump links using **Tab** and **Shift + Tab**. 
+- Keyboard users should also be able to select a jump link using **Enter** and **Space**.
```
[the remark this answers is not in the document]
nicolethoen: TIL

packages/v4/src/content/accessibility/jump-links/jump-links.md
```diff
@@ -0,0 +1,34 @@
...
+
+| React component | React prop      | Which HTML element it appears on in markup   | Explanation                                                   | 
+|-----------------|-----------------|----------------------------------------------|---------------------------------------------------------------|
+| JumpLinks       | aria-label      | nav.pf-c-jump-links aria-label               | Labels the nav to reflect the current nav description.        |
+| JumpLinks       | toggleAriaLabel | .pf-c-jump-links__toggle > button aria-label | Accessible label for the expandable jump links toggle button. |
+
+**Note:** Whenever multiple navigation elements are present on a page, an aria-label is highly recommended. Screen reader users
+navigating a page via an elements list or rotor menu will be unable to distinguish between the various navigation elements. Using an aria-label
+in these cases allows the user to differentiate between the navigation elements without navigating to and interacting with each one individually
+to determine its contents. The following image demonstrates the lack of information when aria-labels are not present:
```
thatblindgeye, line 31: No other comment other than 🎉 for this verbiage

thatblindgeye (approved):
Latest updates look good! Regarding a template for a11y docs, I think merging them in if there are no other issues would be fine, and simply applying the template to them once that is completed and approved. These additions/updates would be great to convey in the present rather than holding them off I think.

packages/v4/src/content/accessibility/progress-stepper/progress-stepper.md
```diff
@@ -0,0 +1,20 @@
...
+A **progress stepper** is used to provide information to the user about the state of their current progress on a multistep interaction.
+
+**Keyboard users** use standard keyboard navigation (**Tab** and **Shift + Tab**) to move to and between progress steps with popovers. Keyboard users should
+be able to open the popover on a progress step using **Enter** or **Space**, and close the popover using **Enter**, **Space**, or **Escape**.
+
+**Screen reader users** should be able to discern the label, status, and optional description of every progress step 
+via the `aria-label` prop. This should include all information being communicated by the progress step's icon, such as 
+the variant and the completed status, as well as whether the step has an associated popover they can trigger. Screen 
+reader users should also be able to access the contents of any popover associated with a step.
+
```
jessiehuff, line 15: So I think that we probably need to add an "Accessibility Application" section or whatever we're planning on calling it here. (I think we've called it different things on different a11y pages.) 
When we look at the progress stepper react demos, we're creating a live region that we're announcing the progress with. This should probably be explained in the a11y docs:
``` 
     <div className="pf-screen-reader" aria-live="polite">
          {steps[currentStep] && `On ${steps[currentStep].title}.`}
          {steps[currentStep - 1] && `${steps[currentStep - 1].title} was successful.`}
     </div>
     <ProgressStepper>
          {steps.map((step, index) => {
            let variant = 'pending';
            let ariaLabel = 'pending step';
            if (index < currentStep) {
              variant = 'success';
              ariaLabel = 'completed step, step with success';
            } else if (index === currentStep) {
              variant = 'info';
              ariaLabel = 'current step';
            } 
 ```

File: packages/v4/src/content/accessibility/badge/badge.md (hunk 2 above)
Comment (jessiehuff) [reply packages/v4/src/content/accessibility/badge/badge.md:14]: I agree with these above comments. When reading it, I also thought something more action oriented would be better like "Do not rely on color alone to communicate" because I think it's clearer for those skimming the docs. 

I think an example image could be powerful as well. Perhaps we could show badges or labels with just using color but through another lens (what it looks like when you have this eye condition, etc). I think Eric's idea of required labels being in red (but they all look the same to another user) would be really informative. I remember that was effective when we were talking about chart a11y.

File: packages/v4/src/content/developer-resources/accessibility-development.md [hunk 12]
```diff
@@ -98,3 +98,24 @@ The recommended interaction pattern for the modal components like the modal or p
...
+will automatically make any scrollable element focusable to ensure keyboard accessibility as long as they have a 
+non-negative tabindex.
+
+Additionally, when adding tabIndex in this scenario, it would also be beneficial to add an aria-label that describes the scrollable element.
```
thatblindgeye, line 109: Slight tweak suggestion:

```suggestion
Additionally, when adding tabIndex in this scenario, it can also be beneficial to add an aria-label that describes the scrollable element, similar to how a heading describes a section of content.
```

Thoughts being that maybe it will paint a clearer picture for how to use `aria-label` for this use-case.

packages/v4/src/content/accessibility/badge/badge.md
```diff
@@ -10,3 +10,23 @@ A **badge** is used to annotate other information like a label or an object name
...
+in Figure 1 with its lack of visual cues. Compare Figure 1 to Figure 2 where there is an icon
+used to indicate that the second badge is associated with actionable or unread content.
+
+Figure 1.
+
+![badges with no color and no icons](../../../images/badge-no-color-no-icons.png)
+
+Figure 2.
+
+![badges with no color with icons](../../../images/badge-no-color-with-icons.png)
```
thatblindgeye, line 32: Question: For these images, did you by chance use Chrome's vision deficiencies emulation?

Nitpick: I'd say images for a11y docs should either be placed in this `images` directory as done for this doc, or placed inside the corresponding component like the jump links has.

packages/v4/src/content/accessibility/expandable-section/expandable-section.md
```diff
@@ -0,0 +1,23 @@
...
+content. When the expandable content is detached from the expandable section toggle, the expandable content's `id` should be set as the value of the
+expandable section toggle's `aria-controls` attribute.
+
+The following props/attributes have been added for you and are customizable in PatternFly:
+
+| React component         | React prop            | Which HTML element it appears on in markup     | Explanation                                                                   | 
+|-------------------------|-----------------------|------------------------------------------------|-------------------------------------------------------------------------------|
+| ExpandableSection       | contentId             | .pf-c-expandable-section__content              | Id of the expandable section, associates this content with its toggle button. |
+| ExpandableSection       | toggleContent         | .pf-c-expandable-section__toggle-text          | Aria-label for the Context Selector Search button                             |
```
thatblindgeye, line 22: This item should be removed; it mentions context selector, but I also don't believe the expandable section `toggleContent` prop is a11y specific.

packages/v4/src/content/accessibility/progress-stepper/progress-stepper.md
```diff
@@ -0,0 +1,32 @@
...
+The following props/attributes have been added for you and are customizable in PatternFly:
+
+| React component | React prop  | Which HTML element it appears on in markup | Explanation                       | 
+|-----------------|-------------|--------------------------------------------|-----------------------------------|
+| ProgressStep    | aria-label  | .pf-c-progress-stepper__step               | Progress step for screen readers. |
```
thatblindgeye, line 20: For the explanation, maybe something more along the lines of, "Provides an accessible label describing the variant and status of the progress step," or something that conveys what the aria label is doing.

packages/v4/src/content/accessibility/progress-stepper/progress-stepper.md
```diff
@@ -0,0 +1,32 @@
...
+
+
+**Note:** When using a progress stepper in an application, it's also wise to consider how to best summarize the overall progress
+state to a screen reader user. The [basic progress stepper demo]('components/progress-stepper/react-demos#basic')
```
thatblindgeye, line 24: Not sure if it's just because it's a preview build, but this link doesn't redirect correctly.

thatblindgeye (commented):
Looks good! Some nitpicks/questions/comments below in specific areas. One general comment is in instances where attributes/props are mentioned outside of the props/attributes table, perhaps we should wrap those in backticks to differentiate them from other text.

packages/v4/src/content/developer-resources/accessibility-development.md
```diff
@@ -98,3 +98,24 @@ The recommended interaction pattern for the modal components like the modal or p
...
+Some screen reader users navigate page elements via an element list or rotor menu and the default label for these scrollable elements
+will be driven by whatever text content exists inside the container. This generated label may not be suitable to describe the regions purpose, and 
+would likely be better described with a simple aria-label.
```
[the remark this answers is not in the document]
nicolethoen: Maybe when we open and begin tackling an issue to address the use of aria-labels on divs or other places where that issue is flagged, we can also update these docs. As far as I can tell, axe is flagging it as 'needs review' and on the wizard examples, it make sense that - because of the rotor menu navigation as spaxman noted - these situational uses passes review (for me). WDYT?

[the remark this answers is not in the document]
thatblindgeye: That's a fair point. I suppose between aXe only telling us it isn't "well supported" and Jessie not having had any issues/not finding anything official (I also haven't really had much issues other than some potential JAWS behavior, but that would require more specific testing), it could be worth waiting to see if we hear about any issues/come across any resources about it down the line.

If we were to add anything now, something along the lines of "This may not be well supported in some circumstances, but because of X, Y, and Z we are opting to do this" might be better, but I'd be okay holding off on adding any verbiage about it for now. Worse case is one of us tests or comes across something a few days from now and update the docs then.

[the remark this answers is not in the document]
nicolethoen: I actually did a little more reading and found this blog: https://www.tpgi.com/short-note-on-improving-usability-of-scrollable-regions/

it lead me to read up on use of `role='region'` and using that role seems to remove the warnings. What do you think? 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role

packages/v4/src/content/accessibility/badge/badge.md
nicolethoen: I used a chrome extensions called 'A11y - color blindness empathy test'.

thatblindgeye (approved):
🎉 awesome work on all of these updates

jessiehuff (approved):
LGTM!  🎉 🏆 🎉

PR #2923, opened 2022-04-18
Fix nested type parameters in TS example

CooperRedhat:
Re: https://github.com/patternfly/patternfly-react/pull/7219

Fix nested type parameters

- acorn (and the acorn-typescript extension) treated `>>` as a single token
- There is likely a better way to do this, but this seems to fix the problem
  - Potential issue would be someone using `>>` in a bitwise operation, but very unlikely that bitwise operations are happening in our react examples


CooperRedhat:
Example code do not merge

PR #2931, opened 2022-04-20
re-write form sentence

### LINKED ISSUE #2930: Edit multi-column form guidelines
"Never design multi-column forms to read like a book, where the left column is completed first before moving on to the right. The form itself should still move and be filled from top to bottom, with related fields placed on the same line."

Instead, write "“Never design multi-column forms to read like a magazine or newspaper column, where the left column is completed first before moving on to the right. The form itself should still move and be filled from top to bottom, with related fields placed on the same line."

mmenestr:
closes #2930


jeperry:
LGTM!

jeperry (approved):
This looks good to me and clarifies the reading order of the form.

PR #2939, opened 2022-04-22
chore(deps): bump React to version 17

### LINKED ISSUE #2837: Chore: Update React dependency  from 16 to 17
Update to React 17

wise-king-sullyman:
Closes #2837

### DIFF
--- a/packages/theme-patternfly-org/package.json
+++ b/packages/theme-patternfly-org/package.json
@@ -70,7 +70,7 @@
     "webpack-merge": "^5.0.9"
   },
   "peerDependencies": {
-    "react": ">=16.8.0",
-    "react-dom": ">=16.8.0"
+    "react": ">=17.0.2",
+    "react-dom": ">=17.0.2"
   }
 }
--- a/packages/v4/package.json
+++ b/packages/v4/package.json
@@ -18,8 +18,8 @@
   },
   "dependencies": {
     "@patternfly/react-docs": "5.63.17",
-    "react": "^16.8.0",
-    "react-dom": "^16.8.0",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
     "theme-patternfly-org": "^0.11.28"
   },
   "devDependencies": {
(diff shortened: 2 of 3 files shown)


packages/v4/package.json
```diff
@@ -18,8 +18,8 @@
   },
   "dependencies": {
     "@patternfly/react-docs": "5.63.17",
-    "react": "^16.8.0",
```
evwilkin, line 21: Let's update these to match what's in `@patternfly-react/react-core` so we're not removing support for react & react-dom 16.8+
- https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/package.json#L60-L61

packages/theme-patternfly-org/package.json
```diff
@@ -70,7 +70,7 @@
     "webpack-merge": "^5.0.9"
   },
   "peerDependencies": {
-    "react": ">=16.8.0",
-    "react-dom": ">=16.8.0"
```
evwilkin, line 74: Let's update these to match @patternfly-react/react-core so we're maintaining support for 16.8+ while adding support for 17+
- https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/package.json#L60-L61

evwilkin (changes requested):
These deps should be updated to add support for 17.0+ but not remove support for 16.8+ as seen here:
- https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/package.json#L60-L61

evwilkin (dismissed):
LGTM - thanks @wise-king-sullyman !

evwilkin:
Hey @CooperRedhat @cdcabrera we had an issue open to bump Org to support React 17, we've mirrored the dependencies from the patternfly-react repo but replaced the [existing dependencies from the theme](https://github.com/patternfly/patternfly-org/pull/2939/files#diff-d25cba053d1632764148892418656a66f37348a7097b10d7640126c4cd655ae3L73-L74) to avoid React 18 for now as I'm unsure what the ramifications are.  Given the work you're doing in the theme already, any feedback here would be helpful before merging 👏

evwilkin (approved):
LGTM

PR #2961, opened 2022-05-11
fix(docs): resolve accessibility issues on static pages

### LINKED ISSUE #2921: website a11y effort: clean up simple fixes in static content pages
(i.e. getting started or the community pages)

wise-king-sullyman:
(potentially) Closes #2921 

This PR addresses all (non color contrast) issues I could find using Axe DevTools on the homepage, getting started pages, developer resources pages, accessibility pages, guidelines pages, UX writing pages, contribution pages, training pages, and the community page.

It also fixes an issue in the left nav that was appearing on all pages.

It will likely be beneficial for me to talk through the reasoning behind some of these changes during the review process, please let me know if you'd like to jump on a call while reviewing!

I plan to squash and rebase with an appropriate commit name before merging, but I wanted to leave individual commits in place, for now, to make reviewing easier.

Issues resolved include:
- Link being focusable while aria-hidden
- Aria-label applied to invalid elements
- Heading levels should start with `h1` and increment sequentially
- Link text should be descriptive and identical link text on the same page should do the same function
- Table headers were missing text
- IDs were repeated on the page on different elements
- Inputs were lacking labels
- `<dt>` tags did not have corresponding `<dd>` elements


packages/v4/src/content/design-guidelines/styles/colors/ColorFamily.js
```diff
@@ -54,6 +54,15 @@ export function ColorFamily({
             {title}
           </h3>
         </dt>
+        <dd
```
nicolethoen, line 57: I wonder if it'd fix the error and keep the UI clean if you change each of the subsequent `dt` elements into `dd` elements? so that each of the colors in each color family is a `dd` element? i think that'd satisfy the rule.

nicolethoen (commented):
Only one question. In general I think it looks great! :)

packages/v4/src/content/design-guidelines/styles/colors/ColorFamily.js
wise-king-sullyman: That does indeed seem to satisfy it, and is much clearner. Thank you!

nicolethoen (approved):

evwilkin (approved):
LGTM - thanks for all this work!  

I added a list of the common issues resolved here in the PR description, which may be beneficial for a quick reference as we continue to improve a11y on the website.

PR #2971, opened 2022-05-19
fix(toc): fixed jumplinks in TOC to resolve a11y nesting issue

### LINKED ISSUE #2969: Jump links contain invalid HTML
It is invalid HTML to nest `<ul>` as children of `<ul>`.
This could be added as a variant of Jump links in React (i.e. nested jump links) and then consumed by org (ideal) - or fixed in org specifically (faster).

this will solve HUNDREDS of the a11y issues on the website.

evwilkin:
closes #2969 

This PR:
- Mirrors the structure of the [nested jump links example](https://www.patternfly.org/v4/components/jump-links#expandable-vertical-with-subsection) from our docs to fix a11y error of a `<ul>` nested directly inside another `<ul>`.

Fix:
Previously we were either rendering `<JumpLinksList>` directly inside the `<JumpLinks>` component:
```
<JumpLinks>
  <JumpLinksItem><JumpLinksItem>
  <JumpLinksList><JumpLinksList/>
</JumpLinks>
```

We should always render a `<JumpLinksItem>` and place any needed `<JumpLinksList>`s within them where needed.
```
<JumpLinks>
  <JumpLinksItem></JumpLinksItem>
  <JumpLinksItem>
    <JumpLinksList><JumpLinksList/>
  </JumpLinksItem>
</JumpLinks>
```

### DIFF
--- a/packages/theme-patternfly-org/components/tableOfContents/tableOfContents.js
+++ b/packages/theme-patternfly-org/components/tableOfContents/tableOfContents.js
@@ -10,25 +10,57 @@ export const TableOfContents = ({ items }) => {
     const { innerWidth } = window;
     innerWidth !== width && setWidth(innerWidth);
   }
+  let wasSublistRendered = false;
 
-  const renderItem = (item, index) => {
-    return Array.isArray(item)
-      ? (
-        <JumpLinksList key={index} className="ws-toc-sublist">
-          {item.map(renderItem)}
+  const renderSublist = (item, nextItem, idx) => {
+    wasSublistRendered = true;
+    return (
+      <>
+        {item.text}
+        <JumpLinksList>
+          {nextItem.map(curItem => (
+            <JumpLinksItem
+              key={curItem.id}
+              href={`#${curItem.id}`}
+              className="ws-toc-item"
+              onKeyDown={updateWidth}
+              onMouseDown={updateWidth}
+              onClick={() => trackEvent('jump_link_click', 'click_event', curItem.id.toUpperCase())}
+            >
+              {curItem.text}
+            </JumpLinksItem>
+          ))}
         </JumpLinksList>
-      ) : (
+      </>
+    );
+  }
+
+  let jumpLinksItems = [];
+
+  const renderJumpLinksItems = () => {
+    items.forEach((item, index) => {
+      let cur = item;
+      let next = items[index + 1];
+      // Don't render empty <JumpLinksItem> for an array of sublist items
+      if (wasSublistRendered) {
+        wasSublistRendered = false;
+        return null;
+      }
+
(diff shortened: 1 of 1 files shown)


nicolethoen (approved):

packages/theme-patternfly-org/components/tableOfContents/tableOfContents.js
```diff
@@ -10,25 +10,57 @@ export const TableOfContents = ({ items }) => {
...
     innerWidth !== width && setWidth(innerWidth);
   }
+  let wasSublistRendered = false;
 
-  const renderItem = (item, index) => {
-    return Array.isArray(item)
-      ? (
-        <JumpLinksList key={index} className="ws-toc-sublist">
-          {item.map(renderItem)}
+  const renderSublist = (item, nextItem, idx) => {
```
thatblindgeye, line 15: For this, does `idx` need to be utilized now that the key attribute is no longer using the index, or could the index be omitted from the params here?

packages/theme-patternfly-org/components/tableOfContents/tableOfContents.js
```diff
@@ -10,25 +10,57 @@ export const TableOfContents = ({ items }) => {
...
-      ) : (
+      </>
+    );
+  }
+
+  let jumpLinksItems = [];
+
+  const renderJumpLinksItems = () => {
+    items.forEach((item, index) => {
+      let cur = item;
```
thatblindgeye, line 42: Should `cur` be utilized anywhere within this function, or could this line be removed?

thatblindgeye (commented):
Looks good in the preview! I just had a couple of questions below if you wouldn't mind clarifying as I wasn't 100% sure.

packages/theme-patternfly-org/components/tableOfContents/tableOfContents.js
evwilkin: You're correct, this was leftover from previous refactoring - removed this unused param 👍

packages/theme-patternfly-org/components/tableOfContents/tableOfContents.js
evwilkin: Same as above - removed this unused var 👍

evwilkin:
@thatblindgeye thanks for the feedback, I've removed the unused var/param you called out and made small name changes to other vars to be more descriptive in [my second commit](https://github.com/patternfly/patternfly-org/pull/2971/commits/e5f26aaf0f4c21326075bda2c89877a4e081496a):
- renamed `next` to `nextItem`
- renamed `nextItem` where it's passed as an array to `nextItemArr`

PR #3013, opened 2022-06-15
fix(docs): increased TOC z-index to 501 to fix overlapping examples

### LINKED ISSUE #2941: Bug - buttons overlap mobile jump links
The `.pf-c-toggle-group__button` overlaps the mobile jump links, [seen at this link](https://www.patternfly.org/v4/components/table#composable-actions) and in the screenshot below.  Note that the mobile jumplinks on the website `.ws-toc` have a `z-index: 1` while the overlapping button has a `z-index` of `100`.

A side note, the back to top button `.ws-back-to-top` has `z-index: 2` so it currently overlaps the mobile jumplinks but may need to be considered when adjusting any of these values.

Screen Shot 2022-04-25 at 8 56 46 AM

evwilkin:
Closes #2941 

This PR increases the z-index on the `.ws-toc` jumplinks to 501 to avoid overlapping examples

Note that 501 was chosen to ensure TOC displays over `.ws-preview__thumbnail-link::before, .ws-preview__thumbnail-link::after`.  There are a few elements that are assigned `ZIndex--2xl` in core where`z-index` equals 600, but I believe those elements should still overlap the TOC (for example, modal close button).  Other elements, such as `Menu` demos, have inline z-index of 9999 which cannot be overridden.

### DIFF
--- a/packages/theme-patternfly-org/components/tableOfContents/tableOfContents.css
+++ b/packages/theme-patternfly-org/components/tableOfContents/tableOfContents.css
@@ -4,7 +4,7 @@
   position: sticky;
   width: calc(100% + var(--pf-c-page__main-section--PaddingLeft) + var(--pf-c-page__main-section--PaddingRight));
   background-color: var(--pf-global--BackgroundColor--200);
-  z-index: 1;
+  z-index: 501;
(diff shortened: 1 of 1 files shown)


srambach (approved):
This looks good. I will note that popover and tooltip will overlap, but with a z-index of 9999 I don't think it's worth trying to beat them 😜

evwilkin:
Thanks for that callout - if we find additional issues outside of these `9999` z-index components we can revisit this to further refine 👍

PR #3017, opened 2022-06-16
Fix/2902 toc scroll

### LINKED ISSUE #2902: Responsive TOC on compoents page should scroll
If the TOC for a component page is too long to fit in the new responsive TOC dropdown nav, it's impossible to get to the bottom of the TOC. Should implement scrolling if the contents overflows the height of the viewport.

evwilkin:
closes #2902 

This PR fixes mobile jumplinks not scrolling when they run off the bottom of the page:

- Set max height on expanded jumplinks to ensure they don't stretch off the page
- Set overflow to auto to allow jumplinks to scroll within new max-height container
- Set jumplinks header to sticky & add background and z-index so it remains visible when jumplinks are scrolled


evwilkin:
@mcarrano interested on your input - there isn't a straightforward way to have the TOC both stick to the top of the page but also scroll and reveal the page behind, so @srambach suggested giving the jumplinks a max height and letting them scroll within that container, this way they're still sticky to the top but you can scroll them & see the page below them (or once they're collapsed).  

I picked an arbitrary height of `65vh` which allows them to cover most but not all of the page if they're expanded before they are anchored at the top of the screen, but this height can be adjusted as needed.

srambach (dismissed):

srambach:
I think this looks great - ready to go on your watch interface 😉

mcarrano (dismissed):
@evwilkin I agree. Evan though I'm not a big fan of double scroll bars, this works much better!

evwilkin:
@mcarrano thanks for calling out the double scrollbar - that was unintentional, we're actually hiding the scrollbar for jumplinks currently but as this PR updates the scroll behavior & container the style rule isn't applying correctly.  We _can_ hide the extra scrollbar, would this be preferable?

https://user-images.githubusercontent.com/8651509/174397283-563ae071-93e1-4015-bcac-0e9fd9ebd7c7.mov

mcarrano:
@evwilkin I like it. So if I understand correctly the scrollbar of the page it hidden when the jumplinks panel has focus? Yes, let's do it.

mcoker (dismissed):
LGTM! 

Have we considered using the sidebar component here?

mcarrano (approved):
Looks great. Thanks @evwilkin !

PR #3018, opened 2022-06-17
fix duplicate ids/labels in generated component doc elements

nicolethoen:
- add ids to css variable tables to avoid using default id values and conflicting with examples
- adds example name to code editor toolbar control labels on each example to make them unique

### DIFF
--- a/packages/theme-patternfly-org/components/example/exampleToolbar.js
+++ b/packages/theme-patternfly-org/components/example/exampleToolbar.js
@@ -35,7 +35,9 @@ export const ExampleToolbar = ({
   // Current code in editor
   code,
   // Callback to set code in parent component
-  setCode
+  setCode,
+  // Title of example used in creating unique labels
+  exampleTitle
 }) => {
   const [isEditorOpen, setIsEditorOpen] = React.useState(false);
   const [isCopied, setCopied] = React.useState(false);
@@ -56,6 +58,14 @@ export const ExampleToolbar = ({
   const fullscreenLabel = 'Open example in new window';
   const convertLabel = 'Convert example from Typescript to JavaScript';
   const undoAllLabel = 'Undo all changes';
+
+  const copyAriaLabel = `Copy ${exampleTitle} examle code to clipboard`;
+  const languageAriaLabel = `Toggle ${lang.toUpperCase()} code in ${exampleTitle} examle`;
+  const codesandboxAriaLabel = `Open ${exampleTitle} example in CodeSandbox`;
+  const fullscreenAriaLabel = `Open ${exampleTitle} example in new window`;
+  const convertAriaLabel = `Convert ${exampleTitle} example from Typescript to JavaScript`;
+  const undoAllAriaLabel = `Undo all changes to ${exampleTitle}`;
+
   const customControls = (
     <React.Fragment>
       <CodeEditorControl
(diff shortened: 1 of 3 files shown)


packages/theme-patternfly-org/components/example/exampleToolbar.js
```diff
@@ -56,6 +58,14 @@ export const ExampleToolbar = ({
   const fullscreenLabel = 'Open example in new window';
   const convertLabel = 'Convert example from Typescript to JavaScript';
   const undoAllLabel = 'Undo all changes';
+
+  const copyAriaLabel = `Copy ${exampleTitle} examle code to clipboard`;
```
evwilkin, line 62: `example` typo
```suggestion
  const copyAriaLabel = `Copy ${exampleTitle} example code to clipboard`;
```

packages/theme-patternfly-org/components/example/exampleToolbar.js
```diff
@@ -56,6 +58,14 @@ export const ExampleToolbar = ({
...
+  const languageAriaLabel = `Toggle ${lang.toUpperCase()} code in ${exampleTitle} examle`;
```
evwilkin, line 63: `example` typo
```suggestion
  const languageAriaLabel = `Toggle ${lang.toUpperCase()} code in ${exampleTitle} example`;
```

packages/theme-patternfly-org/components/example/exampleToolbar.js
```diff
@@ -56,6 +58,14 @@ export const ExampleToolbar = ({
...
+  const codesandboxAriaLabel = `Open ${exampleTitle} example in CodeSandbox`;
+  const fullscreenAriaLabel = `Open ${exampleTitle} example in new window`;
+  const convertAriaLabel = `Convert ${exampleTitle} example from Typescript to JavaScript`;
+  const undoAllAriaLabel = `Undo all changes to ${exampleTitle}`;
```
evwilkin, line 67: This is created but never used - it looks like it should be added in on line 168 below `aria-label={undoAllLabel}`

packages/theme-patternfly-org/components/example/exampleToolbar.js
```diff
@@ -56,6 +58,14 @@ export const ExampleToolbar = ({
   const fullscreenLabel = 'Open example in new window';
   const convertLabel = 'Convert example from Typescript to JavaScript';
   const undoAllLabel = 'Undo all changes';
+
+  const copyAriaLabel = `Copy ${exampleTitle} examle code to clipboard`;
+  const languageAriaLabel = `Toggle ${lang.toUpperCase()} code in ${exampleTitle} examle`;
+  const codesandboxAriaLabel = `Open ${exampleTitle} example in CodeSandbox`;
+  const fullscreenAriaLabel = `Open ${exampleTitle} example in new window`;
+  const convertAriaLabel = `Convert ${exampleTitle} example from Typescript to JavaScript`;
```
evwilkin, line 66: This is created but never used, it can be implemented on line 156 below `aria-label={convertLabel}`

evwilkin (changes requested):
A couple typos & 2 variables created but not implemented, but once those are resolved it looks like this is working well!  Thanks for diving deep into Org code for these updates 💪

nicolethoen:
updated - that's what I get for trying to type when I should be sleeping...

evwilkin (approved):
LGTM!

PR #3026, opened 2022-06-23
Update copyright year

andybraren:
A programmatic way to use the current year would probably be best, but for now just updating the year to 2022.


evwilkin (approved):
LGTM - thanks @andybraren 👏

PR #3051, opened 2022-07-06
jumplink

wolfeallison:


mmenestr:
Ok I just have a small change request which is to first show vertical jump links, and maybe mention that these are the recommended onces to use (?) and put horizontal second. What do you think?

mmenestr (approved):

PR #3073, opened 2022-07-12
rearrangin data list

### LINKED ISSUE #3072: Data list reordering!
Just changing the order and headings of this component!

ssalice9:
closes #3072 

No new content added, just rearranged to reflect template and added "spacing" heading

### DIFF
--- a/packages/v4/patternfly-docs/content/design-guidelines/components/data-list/data-list.md
+++ b/packages/v4/patternfly-docs/content/design-guidelines/components/data-list/data-list.md
@@ -23,17 +23,22 @@ The elements mentioned below are similar for a data list with compact or default
 ## Usage
 Think of each row in a data list as a container for some formatted content. In PatternFly, data list rows can accept any [layout](/layouts/bullseye) supported by the design system as long as all rows apply the same layout. Here are some common layouts that may be useful in a data list:
 
-* **[Grid](/layouts/grid/react):** Displays content in a responsive grid.
-* **[Level](/layouts/level/react):** Justifies content evenly over the width of the row and vertically aligns items in the center of the row.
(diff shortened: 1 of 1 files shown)


packages/v4/patternfly-docs/content/design-guidelines/components/data-list/data-list.md
```diff
@@ -21,7 +28,14 @@ Think of each row in a data list as a container for some formatted content. In P
 * **[Split](/layouts/split/react):** Distributes content evenly with a main content area in the center.
 * **[Flex](/layouts/flex/react):** Enables more customization control over the alignment and spacing options provided in the other layouts.
```
mmenestr, line 29: It looks like all the "layout" links have changed to not have "react" at the end of them - so just remove the last "react" from each url and you should be good there!

For example, it should be:  [Grid] (/layouts/grid)

packages/v4/patternfly-docs/content/design-guidelines/components/data-list/data-list.md
```diff
@@ -31,20 +45,7 @@ See [when to use compact vs. default spacing](#compact-vs-default-spacing) for m
```
mmenestr, line 45: The "when to use compact vs. default spacing" links needs to change, because it doesn't match the header right now
Try: (#when-to-use-a-data-list-vs-a-table) instead

mmenestr (changes requested):
Hey @ssalice9 just left two minor comments, and have one more which is - I wonder if the "When to use data list vs table" should be inside of the "usage" section instead of at the bottom of the doc like it is now?

ssalice9:
I agree with you on when to use the data list vs. table being under usage. So i moved it there and then got rid of the redundant bullets

PR #3106, opened 2022-07-29
Clean up after removing v3

### LINKED ISSUE #3048: Move patternfly 3 content out of main branch and host elsewhere
We should no longer need to build/deploy PatternFly 3 web content any longer - it can be archived somewhere and just linked to in case of necessity.

Steps to do this:

- [x] Move v3 content to pf3.patternfly.org in s3 bucket
- [x] Create patternfly-3 branch with just v3 content on patternfly/patternfly-org
- [x] Delete v3 content in packages and update package.json build steps to no longer refer to this content
- [x] Update site footer to point at pf3.patternfly.org for PatternFly 3
- [x] Clean up readme to remove any mention of pf3 building
- [ ] After #3047 is merged - remove any github actions referring to the old build steps in single PR
- [ ] Remove old package.json steps referring to pf3 once github actions are committed

nicolethoen:
Closes #3048 

Remove references to v3 in package.json, the side nav, and the redirects file.

Also turned off the dark theme switcher - which is only supposed on be on in the workspaces.

There was talk of no longer needing the 'v4' in the URL, I imagine we could update `scripts/writeV3Redirects.js` again to redirect a url with a v4 in it to the url without it if we update the config to no long append 'v4' in the urls. what do you think @evwilkin @dgutride ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,8 +11,6 @@
     "build": "yarn build:v4",
     "build:analyze": "yarn workspace patternfly-org-4 build:analyze && yarn copy:v4",
     "build:v4": "yarn workspace patternfly-org-4 build && yarn copy:v4",
-    "build:v3": "yarn copy:v3",
-    "copy:v3": "rm -rf build/patternfly-org/v3 && mkdir -p build/patternfly-org && cp -r packages/v3/readme.md build/patternfly-org/v3",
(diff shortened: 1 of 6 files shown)


evwilkin (approved):
LGTM - I'm okay with removing the `/v4` from the URL, though that may need a follow-up issue?

Issue #3125, opened 2022-08-11
(documentation-framework): Allow for "React next" component/prop usage


jeffpuzzo:
**Describe the enhancement**
As a part of an effort to expose examples inside of react-docs for components and types that are improved/modified and have the same names as the original component and types (temporarily), we'll need to modify the documentation framework to account for the overlapping names so there is no conflict when building examples via the documentation-framework package.

**Example:** 
`Wizard` has been updated and instead of renaming the updated Wizard to say `WizardComposable`, it will instead have the same name, but exist in a `next` sub-directory of `react-core` and be imported with `@patternfly/react-core/next`. This will also include a new tab name of `React next` which will appear to the immediate right of the `React` tab.

PR #3126, opened 2022-08-11
feat(doc-framework): Allow for 'next' core component usage alongside core components

jeffpuzzo:
**Closes**: #3125
**Depends on**: https://github.com/patternfly/patternfly-react/pull/7824

- Give "next" module components unique key names when building prop components.
- Filter out duplicate component names when building "scope" for the list of examples we render on any given documentation page (based on the "source" passed from react-docs).

### DIFF
--- a/packages/documentation-framework/components/example/example.js
+++ b/packages/documentation-framework/components/example/example.js
@@ -78,6 +78,8 @@ export const Example = ({
   liveContext,
   // Content that appears between h3 and code block to explain example
   children,
+  // Show dark theme switcher on full page examples
+  hasDarkThemeSwitcher = process.env.hasDarkThemeSwitcher
 }) => {
   if (isFullscreenPreview) {
     isFullscreen = false;
--- a/packages/documentation-framework/components/example/example.js
+++ b/packages/documentation-framework/components/example/example.js
@@ -1,7 +1,8 @@
 import React from 'react';
 import { useLocation } from '@reach/router';
-import { Badge, CodeBlock, CodeBlockCode, debounce, Switch } from '@patternfly/react-core';
+import { Badge, CodeBlock, CodeBlockCode, debounce } from '@patternfly/react-core';
 import * as reactCoreModule from '@patternfly/react-core';
+import * as reactCoreNextModule from '@patternfly/react-core/next';
 import * as reactTableModule from '@patternfly/react-table';
 import { css } from '@patternfly/react-styles';
 import { getParameters } from 'codesandbox/lib/api/define';
@@ -25,7 +26,7 @@ class ErrorBoundary extends React.Component {
     super(props);
     this.state = { error: null, errorInfo: null };
   }
-  
+
   componentDidCatch(error, errorInfo) {
     errorInfo._suppressLogging = true;
     this.setState({
@@ -39,13 +40,13 @@ class ErrorBoundary extends React.Component {
       this.setState({ error: null, errorInfo: null });
     }
   }
-  
+
   render() {
     if (this.state.errorInfo) {
       return errorComponent(this.state.error);
     }
     return this.props.children;
-  }  
+  }
 }
 
 // Props come from mdx-ast-to-mdx-hast.js
@@ -77,8 +78,6 @@ export const Example = ({
   liveContext,
   // Content that appears between h3 and code block to explain example
   children,
-  // Show dark theme switcher on full page examples
-  hasDarkThemeSwitcher = process.env.hasDarkThemeSwitcher
 }) => {
   if (isFullscreenPreview) {
     isFullscreen = false;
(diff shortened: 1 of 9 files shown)


packages/documentation-framework/components/example/example.js
```diff
@@ -98,12 +97,22 @@ export const Example = ({
   const [editorCode, setEditorCode] = React.useState(code);
   const loc = useLocation();
 
+  // Build unique list of module components for when core module names overlap with "next" module names.
+  const moduleComponents = Object.entries({ ...reactCoreModule, ...reactTableModule }).reduce((acc, [componentName, componentValue]) => {
```
jschuler, line 101: could this more simply be
```
const moduleComponents = {
   ...reactCoreModule,
   ...reactTableModule,
   ...(source === 'react-next' ? reactCoreNextModule : {})
}
```

packages/documentation-framework/scripts/md/parseMD.js
```diff
@@ -60,15 +60,17 @@ function toReactComponent(mdFilePath, source, buildMode) {
...
-            return true;
-          }
-          file.message(`Prop component ${propComponent} missing from tsDocgen`);
-          return false;
-        })
-        .map(propComponent => tsDocs[propComponent])
+      const propComponents = [...new Set(frontmatter.propComponents || [])].reduce((acc, componentName) => {
+        const name = getTsDocName(componentName, source === 'react-next');
+
+        if (tsDocs[name]) {
```
jschuler, line 66: Trying to understand. So if there's let's say ...next/components/Wizard/Wizard.tsx, then it exists as `tsDocs['Wizard-next']`?
And then it will get pushed to 
```
propComponents[
  {
    "name": "Wizard-next",
    "props": []
  }
]
```
?

packages/documentation-framework/scripts/md/typecheck.js
```diff
@@ -34,13 +34,18 @@ declare module '\\*.svg' {
 const reactStylesDir = path.join(require.resolve('@patternfly/react-styles/package.json'), '../');
 const reactStyles = glob.sync(path.join(reactStylesDir, 'css/**/*.d.ts'))
   .map(f => f.replace(reactStylesDir, '@patternfly/react-styles/').replace(/\.d.ts$/, ''));
+const defaultImports = [
+  'react',
+  '@reach/router',
+  '@patternfly/react-core/next',
```
jschuler, line 40: is this not added to versions.json?

packages/documentation-framework/scripts/webpack/webpack.base.config.js
```diff
@@ -117,6 +117,7 @@ module.exports = (_env, argv) => {
         'client-styles': path.resolve(process.cwd(), 'patternfly-docs/patternfly-docs.css.js'),
         './routes-client': path.resolve(process.cwd(), 'patternfly-docs/patternfly-docs.routes.js'),
         './routes-generated': path.resolve(process.cwd(), 'patternfly-docs/generated/index.js'),
+        '@patternfly/react-core/next': '@patternfly/react-core/dist/js/next/index.js',
```
jschuler, line 120: why is this needed?

jeffpuzzo: It shouldn't be after the next path exists in patternfly-react. This PR wasn't quite ready to be reviewed as it depends on the version of react-core that has the "next" path.

packages/documentation-framework/scripts/md/parseMD.js
jeffpuzzo: No, the name of the component would be "Wizard" still, but the props would be that of the appropriate component.

Here's an example generated output for the "Next" Wizard:
```
const pageData = {
  "id": "Wizard",
  "section": "components",
  "source": "react-next",
  "slug": "/components/wizard/react-next",
  "sourceLink": "https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/next/components/Wizard/examples/Wizard.md",
  "propComponents": [
    {
      "name": "Wizard",
```

packages/documentation-framework/scripts/md/typecheck.js
jeffpuzzo: I did not update that file. What are the expectations there and what is it used for exactly?

packages/documentation-framework/scripts/webpack/webpack.base.config.js
jeffpuzzo: Spoke with @jschuler in slack about this and we decided to keep this around until webpack is updated to v5 for the documentation-framework.

tlabaj (dismissed):
I am not familiar with the framework.  Looks good to me if @jschuler or @CooperRedhat are good with it.

packages/documentation-framework/components/example/example.js
```diff
@@ -77,8 +78,6 @@ export const Example = ({
   liveContext,
   // Content that appears between h3 and code block to explain example
   children,
-  // Show dark theme switcher on full page examples
-  hasDarkThemeSwitcher = process.env.hasDarkThemeSwitcher
```
nicolethoen, line 81: Why did you remove this? I think we want the dark theme switcher on full pages if the environment sets `hasDarkThemeSwitcher` to true.

And it is also seeming like the full page demos are not working in the review? 
https://patternfly-org-pr-3126-v4.surge.sh/v4/components/notification-drawer/react-demos/basic/
https://patternfly-org-pr-3126-v4.surge.sh/v4/components/page/react-demos/sticky-section-group-using-pageheader/

jeffpuzzo: Ah, not intentional. Will add this back.

jeffpuzzo: Both of those links point to their respective pages now fine it appears.

jschuler (approved):

PR #3147, opened 2022-08-22
Adding row striping to table docs

### LINKED ISSUE #2783: Add to table guidelines: mention ability to have "striped" tables
Want to add guidance on possibility of having "striped" tables, where every other row will differ in color. 
Probably would just add a small section to the table guidelines to address this + some guidance around when someone might want to do this.

Here is the demo for it: https://www.patternfly.org/v4/components/table/html#striped

original design issue: https://github.com/patternfly/patternfly-design/issues/1102#issuecomment-994907397

andrew-ronaldson:
closes #2783


mmenestr (changes requested):
Looks good, just two small things:
1. could you add a picture of the row striping 
2. can you change the "when to use row stripping in tables" header to just say "when to use" to match the other instances of it?

Thanks!!

mmenestr (dismissed):

mmenestr (approved):

PR #3153, opened 2022-08-23
docs(Card): update a11y doc to new template

### LINKED ISSUE #2809: Update card a11y docs to include selectable cards
Follow up to https://github.com/patternfly/patternfly-react/issues/6798
A11y docs need to be updated to account for selectable cards to explain how to set it up for use with a screen reader

### LINKED ISSUE #3112: Update a11y documentation for Card
Card

**Describe the enhancement**
Updates the a11y documentation to the new guidelines

N/A

**Any other information?**
N/A

thatblindgeye:
Closes #2809, closes #3112 

[Card a11y preview build](https://patternfly-org-pr-3153-v4.surge.sh/v4/components/card/accessibility/)

I included some verbiage about placing multiple cards within a list container, though I don't believe this is a convention we've shown in card examples. Would like to hear any other opinions on it.

### DIFF
--- a/packages/v4/patternfly-docs/content/accessibility/card/card.md
+++ b/packages/v4/patternfly-docs/content/accessibility/card/card.md
@@ -33,6 +33,9 @@ At a minimum, a card should meet the following criteria:
   <ListItem>
     <Checkbox id="card-a11y-checkbox-5" label={<span>If there's no surrounding context, a card list container has its own accessible name via the <code class="ws-code">aria-label</code> attribute.</span>} description="An example of surrounding context would be a heading element that directly precedes the group of cards." />
   </ListItem>
+  <ListItem>
+    <Checkbox id="card-a11y-checkbox-6" label={<span>If building a single-selectable card for a <a href="/demos/primary-detail">primary detail view</a>, the card has an associated hidden checkbox input.</span>} description={<span>This notifies users navigating via screen readers and similar assistive technologies that a checkbox can be selected and of its current selected status. See the <a href="/components/card#selectable-card-accessibility-features">React selectable card accessibility features example</a> or <a href="/components/card/html#selectable-with-a-hidden-input-for-improved-screen-reader-accessibility">HTML selectable with hidden input example</a> for implementation information.</span>} />
+  </ListItem>
 </List>
 
 ## React customization
--- a/packages/v4/patternfly-docs/content/accessibility/card/card.md
+++ b/packages/v4/patternfly-docs/content/accessibility/card/card.md
@@ -9,10 +9,9 @@ import { Checkbox, List, ListItem } from '@patternfly/react-core';
 
 To implement an accessible PatternFly **card**:
 
-- Provide an accessible name for the card via the `aria-label` or `aria-labelledby` attribute.
+- Provide an accessible name for the card via the `aria-label` attribute, or by linking to a card title via the `aria-labelledby` attribute.
 - Ensure any other PatternFly components within a card follow their own accessibility documentation.
-- Place a set of multiple cards inside of a list container, with a list item wrapped around each card.
-- Give a card list container its own accessible name via the `aria-label` attribute if there is no other surrounding context.
+- Place cards inside of a list element whenever displaying a set of multiple cards, with a list item wrapped around each card. This list element must also have its own accessible name via the `aria-label` attribute if there is no other surrounding context.
 
 ## Testing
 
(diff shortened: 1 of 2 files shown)


packages/v4/patternfly-docs/content/accessibility/card/card.md
```diff
@@ -3,21 +3,109 @@ id: Card
...
+  </ListItem>
+</List>
+
+## React customization
+
+Various React props have been provided for more fine-tuned control over accessibility.
+
+| Prop | Applied to | Reason | 
+|---|---|---|
+| `aria-label` or `aria-labelledby` | `Card` | Adds an accessible name to the card. If the card has a `CardTitle`, the `aria-labelledby` prop should be passed in with a value of the `CardTitle`'s `id` prop. Otherwise the `aria-label` prop should be passed in to act as a title for the card. |
```
nicolethoen, line 45: It may be worth noting here or in the prop description below it that if they use the `div` element for decorative reasons, they should not use an aria-label, since divs should have a role if they use an aria-label.

packages/v4/patternfly-docs/content/accessibility/card/card.md
```diff
@@ -3,21 +3,109 @@ id: Card
...
+| `isExpanded` | `Card` | Adds styling to a card toggle to visually indicate whether the expandable content is expanded or collapsed. When this prop is passed in, `onExpand` and `toggleButtonsProps` must be passed into the `CardHeader` sub-component. |
+| `isSelectable` or `isSelectableRaised` | `Card` | Adds styling on hover or focus to visually indicate that a card is selectable, and adds the `tabindex="0"` attribute so that the card can be navigated to with standard keyboard navigation. |
+| `isSelected` | `Card` | Adds styling to visually indicate that a card is currently selected. When the `hasSelectableInput` prop is passed in, this prop will also set the selectable input's `checked` attribute. |
+| `selectableInputAriaLabel` | `Card` | Adds an accessible name to the hidden input when `hasSelectableInput` is passed in and there is no `CardTitle` sub-component within the card. The value passed into this prop should act as a title for the card. |
+| `component` | `CardBody` | Sets the base element for the sub-component (`div` element by default). |
```
nicolethoen, line 53: It's not clear from this description when/how the `component` prop has any affect on a11y.

packages/v4/patternfly-docs/content/accessibility/card/card.md
```diff
@@ -3,21 +3,109 @@ id: Card
...
+| `component` | `CardFooter` | Sets the base element for the sub-component (`div` element by default). |
+| `onExpand` | `CardHeader` | Adds a toggle button with a caret icon to control whether a card is expanded or collapsed, and to visually indicate the card is expandable. <br/><br/> If an expandable card does not include a `CardTitle`, there should be some other visual context in the `CardHeader` to convey the content of the card to users that rely on sight, such as an image. |
+| `toggleButtonProps` | `CardHeader` | Sets props on the card's toggle. **Required** when a card is expandable. |
```
nicolethoen, line 56: It's not clear from this description when/how the `toggleButtonProps` prop has any affect on a11y.

nicolethoen: I see now that you have more explanation about that below. Maybe you should note that a11y considerations for toggleButtonProps can be found below - possibly even link to that section header?

packages/v4/patternfly-docs/content/accessibility/card/card.md
```diff
@@ -3,21 +3,109 @@ id: Card
...
+
+## HTML/CSS customization
+
+Various HTML attributes and PatternFly classes can be used for more fine-tuned control over accessibility.
+
+| Attribute or class | Applied to | Reason | 
+|---|---|---|
+| `aria-label` or `aria-labelledby` | `.pf-c-card` | Adds an accessible name to the card. If the card has a `.pf-c-card__title`, the `aria-labelledby` attribute should be passed in with a value of the card title's `id` attribute. Otherwise the `aria-label` attribute should be passed in to act as a title for the card. |
+| `.pf-m-non-selectable-raised` | `.pf-c-card` | Adds styling to visually indicate that a selectable card is disabled. |
+| `.pf-m-selectable-raised` or `.pf-m-selectable` | `.pf-c-card` | Adds styling to visually indicate that a card is selectable. |
```
nicolethoen, line 88: I dont think these style related classes are necessary on this page.

nicolethoen (changes requested):

packages/v4/patternfly-docs/content/accessibility/card/card.md
thatblindgeye: What do you think of, "If you primarily want to use a card for styling or layout purposes, it's recommended to pass in a value of "div" to this prop. When doing so, the `aria-label` or `aria-labelledby` prop should be omitted unless a `role` prop is also passed in."

Possible downside is that we then don't elaborate on what `role` should be passed in which may or may not be an issue I suppose. Could instead mention something along the lines of, "...should be omitted, as these props are not well supported on `div` elements without a `role`".

packages/v4/patternfly-docs/content/accessibility/card/card.md
thatblindgeye: Yeah, this and the line below I'm trying to think of how to expand on. The most I could really think of was after looking at some articles that went into accessible Card-like components was essentially not wrapping the entire card in an `<a>` element (so in this case passing `component="a"` to CardBody or CardFooter). Though we would probably just then have to explicitly mention all the elements to be wary of passing in for these props...

I could add on, "... When customizing this prop, be sure to not set the base element of this sub-component as an interactive element such as a button or anchor". Or maybe just mention being sure to not nest interactive elements inside of interactive elements?

I'm also fine with removing these lines for now as well, though.

packages/v4/patternfly-docs/content/accessibility/card/card.md
thatblindgeye: So for these, the main reason I included them was because without them it would make it difficult for users to understand that they can interact with the cards via pointer devices/which card is currently selected, and they're not styles that are automatically applied on hover. Though this could bring into question to what extent style-related classes should be mentioned on a11y docs (if at all).

packages/v4/patternfly-docs/content/accessibility/card/card.md
nicolethoen: I think i'd opt for the second - it's clearer and gives them a hint about how to fix it further, but in those cases, they likely would need to do more research since they've diverged from the intended case so far.

packages/v4/patternfly-docs/content/accessibility/card/card.md
nicolethoen: If they nest interactive elements, it should flag an a11y error if they are doing any automated testing, so I think it might be okay to leave these off.

nicolethoen (approved):
Great stuff!

packages/v4/patternfly-docs/content/accessibility/card/card.md
```diff
@@ -3,21 +3,102 @@ id: Card
...
+
+## Accessibility
+
+To implement an accessible PatternFly **card**:
+
+- Provide an accessible name for the card via the `aria-label` attribute, or by linking to a card title via the `aria-labelledby` attribute.
+- Ensure any other PatternFly components within a card follow their own accessibility documentation.
+- Place cards inside of a list element whenever displaying a set of multiple cards, with a list item wrapped around each card. This list element must also have its own accessible name via the `aria-label` attribute if there is no other surrounding context.
+
+## Testing
```
jessiehuff, line 16: We should probably include testing for a selectable card. I imagine that we could pull this [selectable card a11y content from the examples page](https://www.patternfly.org/v4/components/card#selectable-card-accessibility-features) and place it on this a11y page? Or should we keep it on the examples page?

thatblindgeye: Honestly I think it can work on either page (or both). We could trim down the example verbiage a little and move some of it over to the a11y page, so that the example doesn't get *as* into the a11y reasoning behind using those props.

I'm not opposed to that, but would a selectable card be something that would be considered the "minimum" for cards, or something that isn't exactly a "standard" Card and more something consumers would continue reading in the React Customization section for their particular use-case?

jessiehuff: I think I was imagining adding one check for testing that kind of variant, but I could see us also just including it further down like we do with [Toast Alerts](https://www.patternfly.org/v4/components/alert/accessibility#toast-alerts) since it's a specific variant. I guess that's a good question - should we include testing for specific variants or just include a separate section? Have we already talked about that before?

nicolethoen: Yeah, I think it makes sense on thinking about it to add one more testing step that starts with clear context (like the ones that start with 'If there's no surrounding context...'
The selectable testing step could start with 'If building a selectable card....' or something like that

thatblindgeye: Sounds good! I'm not sure if it was discussed previously, but moving forward it can make sense to add a checkbox test for these kinds of variants. Each section would still kinda build upon the previous in a "No information about specific variant > A short mention to test for a specific variant > More details behind the prop/attribute for the variant > even more info if needed about the variant" sort of way.

thatblindgeye:
@nicolethoen @jessiehuff Latest update adds a test checkbox for selectable cards. I added linked in the description for the checkbox to the HTML and React examples, but wanted your thoughts on that. Sort of an alternative to what I had previously said:

> We could trim down the example verbiage a little and move some of it over to the a11y page, so that the example doesn't get as into the a11y reasoning behind using those props.

nicolethoen (approved):

jessiehuff (approved):

PR #3181, opened 2022-09-13
Feat/2992 view all components

### LINKED ISSUE #2992: Component catalog page
Create a new page that allows users to browse components in a card/catalog view. See designs here: https://marvelapp.com/prototype/hjbd30h/screen/86590461

evwilkin:
Closes #2992 

This PR adds a component gallery listed as "View all components" at the top of the components sidenav section, which contains a gallery view (illustrations only) and a list view (illustrations + summary text) both of which link directly to component pages on click.  These new components are built in a way that by changing just the `section` and `subsection` props you are able to dynamically generate a new gallery for any nav grouping.

Notes:
- Added ability for consumer to choose whether to show or hide subsections within a sidenav section (ex: `components => menus` expandable section would be listed as "menus" in the gallery & link to first page within that expandable section).
- Added ability for consumer to include subsection items in the gallery (ex: `components => menus` components - Dropdown, Menu, Select, etc -  will be included in the gallery.  They're grouped separately in the sidenav tertiary nav for organization, but can still be listed individually in the gallery view).
- Search only filters by component name.  Searching through component summary text would be possible, but as this text is only shown in list view and is hidden in gallery view it could lead to confusion around results in gallery view based on text descriptions that don't exist on the page.

Design and implementation has changed significantly throughout build, final designs here:  [https://github.com/patternfly/patternfly-org/issues/2992#issuecomment-1310457997](https://github.com/patternfly/patternfly-org/issues/2992#issuecomment-1310457997)

<details>
<summary>TODO:</summary>

- [x] sidenav.js: replace hard code placement of "View all components" with weighting in .md file
- [x] remove componentsData being passed to docs framework, keep within v4
  - [ ] Complete for sake of this issue, can still consider moving summary text  & illustration location into .md frontmatter and removing need for componentsData file altogether.
- [x] create .md file in v4 which imports componentGallery from docs-framework
- [x] remove component illustrations being accessible from webpack.base.config, move logic back to v4 to be passed as prop to componentGallery
- [x] fix list view preview images are different widths
- [x] Review Beta Label implementation - custom CSS currently to match placement to designs, and Beta label takes up space even when not present to avoid image being resized due to presence of label.
- [x] Split out "sectionGallery" into smaller more composable components in the case that consumer wants to use part but not all of the gallery options.
- [x] Allow user to customize text through props

</details>


evwilkin:
@maryshak1996 I've put up a rough draft of this page to confirm some of the details before cleaning up for a final review, you can see it here:
- https://patternfly-org-pr-3181-v4.surge.sh/v4/components/view-all-components

Questions:
1. All of our website pages have a max-width of 832px - the [mockup shows 3 card columns + the sidebar](https://marvelapp.com/prototype/hjbd30h/screen/86590461) but with this width limitation it looks to fit 2 columns + the sidebar.
2. Is it correct that the sidebar should run the entire height of the page?  Note that if you're near the bottom of the page and select something like the Truncate component, you then have to scroll all the way back to the top of the page to view the component info.
3. I added a toggle at the top showing what this would look like implemented as a drawer rather than a sidebar.  The pro is that it doesn't replace one of the columns of cards as it slides in front of them, but the con is the same issue where the data in the drawer is only visible when the user is at the top of the page.
  - We could fix this issue by simply making the sidebar sticky once it hits the top of the page, so it would follow the user down the screen.  If we wanted to use the drawer I would need to look into how to accomplish something similar.
 
@mcarrano FYI

maryshak1996:
Hey @evwilkin , so exciting to see this coming together! To your questions:

1. 2 columns + sidebar seems fine to me!
2. I think that the sidebar should run the height of the user's display. In my mind, the toolbar and sidebar should be stick to the top and right of the display, and just the contents of the card catalog body should be scrollable. This way the contents of the sidebar are always visible and the search bar is always accessible. If having the toolbar be sticky to the top is out of the scope, I think that that is okay, but I definitely think it's important that the sidebar be sticky with it's contents always viewable.
3. I don't have much of a preference of 'sidebar' vs ' drawer', but I do like that the sidebar doesn't cover up any content -- feels a bit more scannable that way.

packages/v4/patternfly-docs/pages/componentGallery.css
```diff
@@ -0,0 +1,48 @@
+/* Toolbar styles */
```
mcoker, line 1: It seems like some of this could be mitigated by being able to remove padding from `#main-content` on the workspace, and probably putting this element higher up in the DOM so it isn't a child of the mdx-content elements that manage the example/sidebar layout. Probably move up and replace the `.ws-mdx-child-template` element? Then we wouldn't have to account for the `--pf-c-page__main-section--Padding[...]` vars

packages/v4/patternfly-docs/pages/componentGallery.css
```diff
@@ -0,0 +1,48 @@
...
+.ws-component-gallery > .pf-c-toolbar > .pf-c-toolbar__content {
+  --pf-c-toolbar__content--PaddingLeft: var(--pf-c-page__main-section--PaddingLeft);
+}
+
+/* Hide searchinput "x" button in favor of "Reset" text */
+.ws-component-gallery > .pf-c-toolbar .pf-c-text-input-group__utilities {
+  display: none;
+}
+
+/* Sidebar styles */
```
mcoker, line 19: This is impressive CSS! And it looks really good matching the design, working with what you have. That said, I'd love to use this and make some updates to the sidebar component.

* You should be able to have a divider between the sections. You can do that now manually, but you have to manage the orientation, it doesn't work with `.pf-m-panel-right` and doesn't work with a sticky panel in a stack layout (though I'd argue you shouldn't use a border there anyways).
* You should be able to add "body" elements to the sidebar panel and make those sticky so that the sidebar panel can remain full height and the things in it can be sticky.
* The body elements re: ^ should support padding and color variations - or somehow integrate with page sections better, as it's likely you have a sidebar as a full width layout like this and you want to recreate page sections (backgrounds, spacing/padding, etc).

I think that would go a long way here. I've added that to this issue - https://github.com/patternfly/patternfly/issues/5131

packages/v4/patternfly-docs/pages/componentGallery.js
```diff
@@ -0,0 +1,139 @@
...
+                {drawerPanelData.title}
+              </span>
+            </Text>
+            <Text>
+              { drawerPanelData.id ? <SummaryComponent id={drawerPanelData.id}/> : null }
+            </Text>
+          </TextContent>
+          <Button className="ws-view-component" component={(props) => <Link {...props} to={drawerPanelData.slug} />}>View component</Button>
+        </SidebarPanel>
+        <SidebarContent>
```
mcoker, line 98: this should get rid of the white background behind the cards

```suggestion
        <SidebarContent hasNoBackground >
```

mcoker (changes requested):
This is really cool! I left some comments. The only other thing is the links in the preview are missing `/v4`, but I don't know if that's just a PR build thing or not?

mcarrano:
@evwilkin this is looking good, but one thing I noticed besides the issues above is that the description in the sidebar and the search result can get out of sync. See below:

Screen Shot 2022-10-06 at 3 16 42 PM

This is a result of searching on "tree." You'll notice that the description on the right still references About modal which is the first item in the list. I wonder if the sidebar should not display anything until the user explicitly makes a selection. What do you think @maryshak1996 ?

maryshak1996:
Hey all! This is looking really great @evwilkin @mcoker! I see some comments around the sizing of the content area, so I apologize if what I'm about to say is redundant to any of that, but here are a couple things I notice:

1. The toolbar should be flush against the header section (where it says 'All components')
2. Can the card area w/ panel not go full width? It looks strange to have so much space to the right of the panel.

@mcarrano , that's a really good point about the search not affecting the content of the panel. I would support only having the panel appearing once a card is clicked. In that case, it would make sense to make the panel dismissable (since it no longer would be persistent on the page).

mcarrano:
I was just thinking about the process for adding new components. These will require a new illustration to be created and added to the illustrations folder in `patternfly-docs`. We do have a procedure for that and will need to include in our design process, but it does raise a question. What happens if a new component is published and the illustration is unavailable? Do we need or want a default placeholder? Interested in your thoughts @evwilkin @maryshak1996

maryshak1996:
@mcarrano , I think that having a default placeholder image would be a good thing to at least have in our back pockets here. I'm happy to take that task on (I've got some cycles right now to do this) :) I just opened #3222 to track that work.

mcarrano:
That sounds good. Thanks @maryshak1996 !

srambach:
Apologies for being a bit late to the game, but I did want to provide some things I noticed when I first saw this.

- I'd expected this catalog to be wider than 2 items if there's room. It's still a lot of scrolling and I was hoping for something easy to visually scan and compare components - but even at full height of my big screen I can only see most of just 8 at a time.

- I expected that the short description and link straight to the component would be in the card. When browsing, I have to look back and forth from the component thumbnails to the text. And, It's now an additional click to select the component and then a move over to the button to click to get to the page. If I use search to bring up one, I automatically click the card but that doesn't get me anywhere - I still have to go click the View component button. (I realize this is slightly in opposition to my first point since it would take up more room but I still think it's important to have everything together and scannable.)

- Nit - when the page first loads, the "About modal" information is in the sidebar but the card isn't selected.

mcarrano:
@evwilkin I am bothered by the fact that the selected card can leave the view when the sidebar appears and the number of columns gets reduced. I wonder if it makes more sense to always have the sidebar visible and keep it empty if nothing is selected (or design an empty state).

I do wonder how much effort it's worth putting into the sidebar implementation if we will eventually abandon this for something else. There definitely seemed to be some discontent about this approach at the meeting. We should discuss.

@maryshak1996 what do you think?

evwilkin:
@mcarrano I agree - always showing the sidebar, or holding that space so the cards don't overflow into it when it's not displayed, would make for a more predictable user experience.

maryshak1996:
Hey there @mcarrano @evwilkin, I think that these are all valid points and I'm not particularly attached to any particular interaction or design with this. Happy to chat about this!

maryshak1996:
@evwilkin just noticing something a little strange about the interaction between scrolling and the details sidebar. If I'm scrolled down to the bottom of the page and click on one of the last items, like 'Tree view' for example, you can't see any content in the sidebar and you have to know to scroll up just a smidge to see anything. Here are some screenshots I captured:
Screen Shot 2022-11-04 at 12 49 19 PM
Screen Shot 2022-11-04 at 12 49 31 PM

srambach:
I love it! I love the list view with the descriptions there. I do see some wonkiness in the hover with the sticky toolbar though. 
Also, should the links work on the preview site, or I vaguely recall there's something about the `/v4`?

https://user-images.githubusercontent.com/19825616/201372460-3c8e4ccb-d251-4aee-89c1-bccdde98d96e.mp4

evwilkin:
Thanks @srambach !  I'm still finding bugs but that toolbar/hover interaction isn't one I'd caught yet, good callout as we just added that hover effect in.  I'll look at the URL - it should work locally (correct) and in the preview link (incorrect) but I think I've got a fix queued up for this.

Off topic - the gold browser tabs in your video are wild!

packages/v4/patternfly-docs/pages/componentGallery.css
evwilkin: Agreed this would have made this code much easier, but requires structural changes to how `documentation-framework` scaffolds docs based on wrapper elements & inserted parsed markdown docs.  This is outside of the scope of this issue, but may be a future enhancement we look into and evaluate other repercussions especially now that this code is being used outside of just PF docs.

packages/v4/patternfly-docs/pages/componentGallery.js
```diff
@@ -0,0 +1,139 @@
...
+          <ToolbarGroup>
+            <ToolbarItem alignment={{default: 'alignLeft'}}>
+              <Button variant="link" onClick={() => setSearchTerm('')} isInline>Reset</Button>
+            </ToolbarItem>
+          </ToolbarGroup>
+          <ToolbarGroup alignment={{default: 'alignRight'}} spacer={{default: 'spacerMd', md: 'spacerNone'}}>
+            <ToolbarItem alignment={{default: 'alignRight'}}>
+              <Text component={TextVariants.small}>{filteredComponents.length} items</Text>
+            </ToolbarItem>
+          </ToolbarGroup>
```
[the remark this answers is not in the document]
evwilkin: Thank you!  Pulled in and tested, looks better than what I had up 👍

mcarrano (commented):
At first glance this looks great @evwilkin !  @mmenestr @maryshak1996 @lboehling @edonehoo perhaps also take a look.

mmenestr:
@evwilkin looks great!! My only comment is, are we going to call that nav item "View all components"? It feels weird to me to have an "action" as a nav item. Maybe we could simplify it to "All components" (?) Curious what others think!

nicolethoen (dismissed):
This LGTM. The number of style overrides applied to the Toolbar specifically feel like a little bit of a red flag - I wonder if we could use Toolbar props to a similar effect. But that can be a follow up eventually. For now this is good.

evwilkin:
@mmenestr I'm good with "All components", @edonehoo what do you think?  This (as it stands) is different from the landing pages we reviewed for different sections, so I think the "All components" link name indicates that (vs. something like "Overview")

maryshak1996:
This looks great! I think that @mmenestr 's comment around removing the action text makes sense. I have two other notes ...
1. It'd be good to have a way sort of separate the 'All components' item and the rest of the components (since this menu item is the only one that isn't, itself, a component). Could we add a horizontal divider for something like this? (See screenshot below)
Screenshot 2023-04-14 at 10 00 46 AM
2. Not sure if this is the correct PR to comment on this, but I just noticed something funky behavior that happens with the 'All components' page header when the left nav is collapsed (See screenshot below). Basically, the header section just doesn't grow to fill the space to the right. 
Screenshot 2023-04-14 at 9 55 17 AM

edonehoo:
This looks fantastic. I'm also team "All components"! I think it makes sense for it to be named differently and, to me, "Overview" suggests that there will be top level details regarding components, rather than a catalog of everything.

wise-king-sullyman (approved):

mcarrano (approved):
Looks great!

PR #3200, opened 2022-09-19
docs(Modal): move aria-label from <i> to <span>

### LINKED ISSUE #3198: Modal design guidelines: aria-label cannot be placed on i elements
Work towards #2920 

The icons on the Modal design guidelines page should not have an `aria-label` attribute on the `<i>` elements, and instead should be moved to the `<span>` wrapping them.

RuiBrian:
Closes #3198

Moved the `aria-label` properties from child `<i>` elements to parent `<span>` elements.

However, this then brings up the issue where the `role` property must be assigned a value. I've assigned `role="img"` to the <span> elements for the time being, let me know if these should be some other value or if there is another solution.


thatblindgeye (approved):
This looks great! Good choice with adding `role="img"` as well 🎉 

An alternative would be keeping the `aria-label` on that `<i>` element and adding `role="img"` there, but semantically both solutions should be equal.

PR #3205, opened 2022-09-21
fix: remove lookbacks in regex to support safari 16

nicolethoen:
I spent some time parsing these regexp trying to figure out if the lookbacks were necessary. Everything seems to work without them and they are [not supported in Safari](https://caniuse.com/js-regexp-lookbehind).

### DIFF
--- a/packages/documentation-framework/helpers/codesandbox.js
+++ b/packages/documentation-framework/helpers/codesandbox.js
@@ -138,7 +138,7 @@ function getReactParams(title, code, scope, lang, relativeImports) {
     code = code.replace(imgImportMatch[0], `const ${imgName} = "https://www.patternfly.org/v4${scope[imgName]}"`);
   }
 
-  const relImportRegex = /(?<=import[\s*{])([\w*{}\n\r\t, ]+)(?=[\s*]from\s["']([\.\/]+.*)["'])/gm;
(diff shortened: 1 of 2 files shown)


mcoker (approved):
Verified the codesandbox links seem to work fine, and checked a few with pages with relative import paths and it looks the same as what's on .org now.

@evwilkin if it's just a matter of pointing to .org for relative paths, have you looked at the `<base>` element? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base

Here's a codesandbox as a demo. Technically it's supposed to go in `<head>` but it seems like it's being removed from `<head>` when the app renders so I just put it in `<body>` - https://codesandbox.io/s/jovial-hawking-t77v8i?file=/index.html

mcoker:
@nicolethoen @evwilkin can we try and squeeze this one in? Any dev workspace on the current docs-framework fails to load on safari.

evwilkin:
@mcoker that's a new one for me, thanks!  It looks like that could work, but there may be some additional work around these imports as pointing to our Org site still seems fragile - ideally we can eventually get passing the static files over to Codesandbox working so we can avoid these workarounds, for now this update for Safari support looks good to move forward with.

evwilkin (approved):
LGTM

PR #3224, opened 2022-10-11
fix(sideNavLayout):replace use of PageHeader with Masthead

### LINKED ISSUE #3053: Use masthead instead of page header
Can we update this to use the masthead?

https://github.com/patternfly/patternfly-org/blob/9380bcb7a683ec4758f5ac144f214f26a63c0d0f/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js#L211-L225

nicolethoen:
Closes #3053

### DIFF
--- a/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
+++ b/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
@@ -36,10 +36,6 @@
   --pf-c-input-group--BackgroundColor: none;
 }
 
-.ws-masthead .pf-c-nav.pf-m-horizontal {
-  width: 100%;
-}
-
 /* TODO: PageHeaderTools on small viewports */
 @media (max-width: 430px) {
   .ws-page-header > .pf-c-page__header-brand > .pf-c-page__header-brand-link {
--- a/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
+++ b/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
@@ -38,12 +38,9 @@
 #ws-global-search {
   background-color: transparent;
   border: none;
-  /* For icon */
-  padding-left: var(--pf-global--spacer--xl);
-  width: 200px;
 }
 .ws-hide-search-input .algolia-autocomplete,
-.ws-hide-search-input #ws-global-search {
+.ws-hide-search-input {
   display: none !important;
 }
 .ws-toggle-search {
--- a/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
+++ b/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
@@ -61,7 +61,7 @@
   border-bottom: 2px solid #fff;
 }
 /* For tiny (200px) viewports */
-.ws-masthead .pf-c-toolbar  {
+.ws-masthead .pf-c-toolbar__item  {
   min-width: 0;
 }
 .ws-github-pageheader {
--- a/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
+++ b/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
@@ -68,3 +68,7 @@
 .ws-github-pageheader {
   color: var(--pf-global--Color--light-100) !important;
 }
+
+.ws-masthead .pf-c-toolbar  {
+  min-width: 0;
+}
--- a/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
+++ b/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
@@ -36,14 +36,6 @@
   --pf-c-input-group--BackgroundColor: none;
 }
 
-/* TODO: PageHeaderTools on small viewports */
-@media (max-width: 430px) {
-  .ws-page-header > .pf-c-page__header-brand > .pf-c-page__header-brand-link {
-    display: none;
-    visibility: hidden;
-  }
-}
-
 /* For tiny (200px) viewports */
 .ws-masthead .pf-c-toolbar__item  {
   min-width: 0;
--- a/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
+++ b/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
@@ -44,26 +44,7 @@
   }
 }
 
-/* Version toggle */
-.ws-org-version-toggle {
-  color: var(--pf-global--Color--light-100);
-}
-.ws-org-version-toggle::before {
-  content: none !important;
-}
-.ws-org-version-toggle:hover {
-  border-bottom: 1px solid #fff;
-}
-.ws-org-version-toggle-expanded::before {
-  border-left: none;
-  border-right: none;
-  border-top: none;
-  border-bottom: 2px solid #fff;
-}
 /* For tiny (200px) viewports */
 .ws-masthead .pf-c-toolbar__item  {
   min-width: 0;
-}
-.ws-github-pageheader {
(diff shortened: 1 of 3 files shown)


evwilkin:
@nicolethoen pushed my updates to implement the expandable SearchInput with algolia, and was also able to rip out several stale CSS rules 👍

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -45,111 +55,106 @@ const HeaderTools = ({
...
+    <Toolbar isFullHeight>
+      <ToolbarContent>
+        {topNavItems.length > 0 && (
+          <ToolbarItem className="pf-m-overflow-container">
+            <TopNav navItems={topNavItems} />
+          </ToolbarItem>
+        )}
+        <ToolbarGroup
+          alignment={{ default: 'alignRight' }}
+          spacer={{ default: 'spacerNone', md: 'spacerMd' }}
```
mattnolting, line 84:
```suggestion
          spacerItems={{ default: 'spacerNone', md: 'spacerMd' }}
```

mattnolting (changes requested):
Looking good! Saw a few things we can clean up..

Looks to me like this dropdown text shouldn't wrap a mobile views, rather it should truncate. 

Current:
Screen Shot 2022-11-01 at 10 19 41 AM

Updated:
Screen Shot 2022-11-01 at 10 25 33 AM

And it looks like we need to add `min-width: 0` to `.pf-c-toolbar__item`, so replace https://github.com/patternfly/patternfly-org/pull/3224/files#diff-3512b9f9de6b73548461009275d974f097637d70bde1c1b724d30d5a45e8b146R64-R66

```
.ws-masthead .pf-c-toolbar  {
  min-width: 0;
}
```

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
```diff
@@ -95,10 +61,9 @@
   border-bottom: 2px solid #fff;
 }
 /* For tiny (200px) viewports */
-.ws-org-version-switcher .pf-c-dropdown__toggle  >.pf-c-dropdown__toggle-text {
-  white-space: normal;
+.ws-masthead .pf-c-toolbar  {
```
mattnolting, line 64: Sorry, need to target `.ws-masthead .pf-c-toolbar__item` here

mattnolting (dismissed):
Love it! LPTM 🎉

evwilkin (dismissed):
LGTM

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -45,111 +55,105 @@ const HeaderTools = ({
...
+              target="top"
+              aria-label="Link to PatternFly GitHub page"
+              className="ws-github-pageheader"
+            >
+              <GithubIcon />
+            </Button>
+          </ToolbarItem>
+          {hasVersionSwitcher && (
+            <ToolbarItem>
+              <Dropdown
```
mcoker, line 118: I believe this should be the full height variation, that's what we're doing with dropdowns in the masthead like this AFAIK. Also there are custom styles on this making it shift around on hover and when it's expanded - IMO it would be much better as a plain ol' dropdown without the custom stuff. 

Nov-17-2022 18-59-29

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -209,20 +210,31 @@ export const SideNavLayout = ({ children, groupedRoutes, navOpen: navOpenProp })
...
-  );
+    <Masthead className="ws-masthead">
+      <MastheadToggle>
+        <PageToggleButton variant="plain" aria-label="Global navigation">
+          <BarsIcon />
+        </PageToggleButton>
+      </MastheadToggle>
+      <MastheadMain>
+        <MastheadBrand href={prurl || pathPrefix || '/'}>
+          {prnum ? `PR #${prnum}` : <Brand src={logo} alt="Patternfly Logo" />}
```
mcoker, line 221: It may be worth considering using the logo from core, it's bigger and looks better IMO - https://github.com/patternfly/patternfly/blob/5eee883c9aadfa4834798de2ee3c00dfa8a37a6a/src/patternfly/demos/Masthead/masthead-template.hbs#L15-L20

mcoker (approved):
LGTM. Left some comments, but they're pre-existing, so they don't need to block this PR if we want to merge it.

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
nicolethoen: Do you know offhand the best way to do that? Is it already something I can import and use or should I copy and paste an svg from core here?

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
mcoker: @mcarrano @mceledonia wdyt? ^

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
mcoker: On org anyways, the core demo images are at `/assets/images/...`. So they're already there somewhere :D 

I'm not sure if there is a best practice around `import`ing the file and referencing it like most of the react examples do, but you could just do it like this which is pretty simple

```
var logoMd = '/assets/images/logo__pf--reverse-on-md.svg';
var logo = '/assets/images/logo__pf--reverse--base.svg';
var logoBase = '/assets/images/logo__pf--reverse--base.png';

<Brand src={logoBase} alt="PatternFly logo" widths={{ default: '180px', md: '220px' }}>
  <source media="(min-width: 768px)" srcSet={logoMd} />
  <source srcSet={logo} />
 </Brand>
```

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -209,20 +210,36 @@ export const SideNavLayout = ({ children, groupedRoutes, navOpen: navOpenProp })
...
+    <Masthead className="ws-masthead">
+      <MastheadToggle>
+        <PageToggleButton variant="plain" aria-label="Global navigation">
+          <BarsIcon />
+        </PageToggleButton>
+      </MastheadToggle>
+      <MastheadMain>
+        <MastheadBrand href={prurl || pathPrefix || '/'}>
+          {prnum ? `PR #${prnum}` : (
+            <Brand src={logoBase} alt="PatternFly logo" widths={{ default: '180px', md: '220px' }}>
```
mcoker, line 222: oops! I gave you the wrong breakpoint. This will match core.

```suggestion
            <Brand src={logoBase} alt="PatternFly logo" widths={{ default: '180px', '2xl': '220px' }}>
```

mcoker (changes requested):
Just one little nit about the brand image width.

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
```diff
@@ -27,78 +27,24 @@
...
-    display: none;
-    visibility: hidden;
-  }
-}
 @media (max-width: 430px) {
   .ws-page-header > .pf-c-page__header-brand > .pf-c-page__header-brand-link {
     display: none;
     visibility: hidden;
   }
 }
```
mcoker: Just noticed this - do we need it still?

mcoker (approved):
🚀

nicolethoen:
I think this is ready to merge, @evwilkin ?

evwilkin (approved):
LGTM

PR #3237, opened 2022-10-17
fix(components): width limit the description section

### LINKED ISSUE #3184: Some page sections should be width limited
Form component docs screenshot

The description probably shouldn't span the browser width like that. There may be other places on the site that could have the same issue. `isWidthLimited` should fix it, and that prop works on `<Page[Section/Breadcrumb/Navigation]>`

srambach:
Fixes #3184 
Adds IsWidthLimited to the description section, but since that adds a `pf-c-page__main-body` with its own padding, I had to remove the utility class on the page section and instead directly modify the variable to remove top padding on `pf-c-page__main-body`.

### DIFF
--- a/packages/documentation-framework/components/footer/footer.js
+++ b/packages/documentation-framework/components/footer/footer.js
@@ -13,7 +13,7 @@ import logo from '../../layouts/logo.svg';
 
 export const Footer = () => (
   <React.Fragment>
-    <PageSection isWidthLimited key="footer-1" className="ws-org-pfsite-l-footer">
+    <PageSection key="footer-1" className="ws-org-pfsite-l-footer">
       <Grid>
         <GridItem
           sm={12}
@@ -212,7 +212,7 @@ export const Footer = () => (
         </GridItem>
       </Grid>
     </PageSection>
-    <PageSection isWidthLimited key="footer-2" className="ws-org-pfsite-l-footer-dark pf-m-no-fill">
+    <PageSection key="footer-2" className="ws-org-pfsite-l-footer-dark pf-m-no-fill">
       <Grid className="pf-u-py-xl-on-sm pf-u-py-0-on-md pf-u-align-items-center">
         <GridItem md={2} mdOffset={1}>
           <Link
--- a/packages/documentation-framework/components/footer/footer.js
+++ b/packages/documentation-framework/components/footer/footer.js
@@ -13,7 +13,7 @@ import logo from '../../layouts/logo.svg';
 
 export const Footer = () => (
   <React.Fragment>
-    <PageSection key="footer-1" className="ws-org-pfsite-l-footer">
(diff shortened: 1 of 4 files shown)


evwilkin (dismissed):
LGTM - thanks!

mcoker (changes requested):
This looks great. I reviewed with @mceledonia and @lboehling and agreed the title and description should just use our default content/text styling. Then we can remove the extra section with the inline padding style. 

Screen Shot 2022-10-19 at 10 05 38 AM

WDYT about the [community page](https://patternfly-org-pr-3237-v4.surge.sh/v4/community) and the footers?

Screen Shot 2022-10-18 at 3 59 05 PM

Screen Shot 2022-10-18 at 3 58 53 PM

packages/documentation-framework/components/footer/footer.js
```diff
@@ -13,7 +13,7 @@ import logo from '../../layouts/logo.svg';
 
 export const Footer = () => (
   <React.Fragment>
-    <PageSection key="footer-1" className="ws-org-pfsite-l-footer">
+    <PageSection isWidthLimited key="footer-1" className="ws-org-pfsite-l-footer">
```
mcoker, line 16: Looks like this is creating some padding action for this block since there are custom padding styles defined here https://github.com/patternfly/patternfly-org/blob/5f9afc258ddb1bf29f3e9cb1610bb566b81ada24/packages/documentation-framework/components/footer/footer.css#L1-L9

Easiest fix if we want to leave it the way it is would probably just be to set the right vars there, since the body element inherits the main-section's padding vars.

So like:

```
--pf-c-page__main-section--PaddingTop: var(--pf-global--spacer--xl);
--pf-c-page__main-section--PaddingRight: var(--pf-global--spacer--2xl);
--pf-c-page__main-section--PaddingBottom: var(--pf-global--spacer--2xl);
--pf-c-page__main-section--PaddingLeft: var(--pf-global--spacer--2xl);
```

mcoker (changes requested):
Looks awesome, just one thing it looks like.

mcoker (commented):
Just poking around, I think this looks a little odd on the home page since the home page content is centered. Maybe that's OK?

Screen Shot 2022-10-21 at 4 37 51 PM

Here is what it looks like if we use the `isCenterAligned` prop

Screen Shot 2022-10-21 at 4 38 28 PM

This is what it looks like with that set on component pages - wdyt @mceledonia @lboehling?

Screen Shot 2022-10-21 at 4 37 42 PM

FWIW it's currently centered on component pages, it just grows to fill the viewport width.

Screen Shot 2022-10-21 at 4 42 44 PM

mcoker (approved):
lgtm!

evwilkin (approved):
LGTM - thanks!

PR #3252, opened 2022-10-24
chore(background-image): reformat a11y docs

### LINKED ISSUE #3251: Background image

srambach:
Fixes #3251 

Since we mention using alt text, I added it to the HTML section. However, there isn't a React prop for it, so I didn't add it..is that something that should be added?

### DIFF
--- a/packages/v4/patternfly-docs/content/accessibility/background-image/background-image.md
+++ b/packages/v4/patternfly-docs/content/accessibility/background-image/background-image.md
@@ -9,8 +9,8 @@ import { Checkbox, List, ListItem } from '@patternfly/react-core';
 
 To implement an accessible PatternFly **background image**:
 
-- Check that there is sufficient color contrast between different parts of the background image and any text that will be on the image. Automated testing tools don't often pick up background image colors, so they may not be able to tell you if contrast is sufficient for users with visual disabilities.
--  Using alternative text so that a screen reader is able to describe the background image when the user navigates to a page is beneficial.
+- Check that there is sufficient color contrast between different parts of the background image and any text that will be on the image. 
+-  A background image should not be used as the sole method of conveying important information.
 
 ## Testing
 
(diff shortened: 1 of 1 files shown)


nicolethoen (dismissed):
I dont think you need to list `alt` under react props. A consumer can still pass the `alt` html attribute to the react component.

thatblindgeye (commented):
In regards to the `alt` attribute, there is the following issue regarding LoginPage and its use of BackgroundImage under the hood: https://github.com/patternfly/patternfly-react/issues/7867

Since the component is using an `<svg>` rather than an ``, and since the component would most likely only be decorative, we could probably omit the `alt` attribute completely in the docs. [Deque university](https://dequeuniversity.com/rules/axe/4.2/svg-img-alt_) has a page on alt text for SVGs.

packages/v4/patternfly-docs/content/accessibility/background-image/background-image.md
```diff
@@ -3,11 +3,36 @@ id: Background image
...
+import { Checkbox, List, ListItem } from '@patternfly/react-core';
 
-**Keyboard users** should not be able to interact with or navigate to the background image, as it is not an interactive element.
+## Accessibility
 
-**Screen reader users** do not need to be able to interact with the background image. However, using alternative text for the screen reader to be able to describe the background image when the user navigates to a page is beneficial.
+To implement an accessible PatternFly **background image**:
 
-## Accessibility application:
 - Check that there is sufficient color contrast between different parts of the background image and any text that will be on the image. Automated testing tools don't often pick up background image colors, so they may not be able to tell you if contrast is sufficient for users with visual disabilities.
```
thatblindgeye, line 12: Could be worth adding this item as a testing checkbox as well so that it's in both sections. Doing that we could shorten this bullet by pulling out the second sentence ("Automated testing tools...") and placing it as the test item description.

packages/v4/patternfly-docs/content/accessibility/background-image/background-image.md
```diff
@@ -3,11 +3,36 @@ id: Background image
...
+## Testing
+
+ At a minimum, a background image should meet the following criteria:
+
+<List isPlain>
+  <ListItem>
+    <Checkbox id="alertGroup-a11y-checkbox-1" label="Users should not be able to interact with or navigate to the background image with the keyboard, as it is not an interactive element." />
+  </ListItem>
+  <ListItem>
+    <Checkbox id="alertGroup-a11y-checkbox-2" label="Users do not need to be able to interact with the background image via a screen reader." />
```
thatblindgeye, line 24: For the start of the label, "Users should not be able to..." rather than "Users do not need to be able to..." (this might actually depend on what we end up doing regarding the alt text).

Additional nit for the ID's for both items: `backgroundImage-a11y-checkbox-X`

thatblindgeye (commented):
Couple of additional comments in addition to the above

thatblindgeye (approved):

nicolethoen (approved):

PR #3256, opened 2022-10-25
docs(Menu): add accessibility documentation

jessiehuff:
Closes #https://github.com/patternfly/patternfly-org/issues/3255

This feels like a start, but I'm not sure how much detail to include and how much I should go into specific examples

### DIFF
--- a/packages/v4/patternfly-docs/content/accessibility/menu/menu.md
+++ b/packages/v4/patternfly-docs/content/accessibility/menu/menu.md
@@ -1,0 +1,53 @@
+---
+id: Menu
+section: components
+---
+
+import { Checkbox, List, ListItem } from '@patternfly/react-core';
+
+
+## Accessibility
+
+To implement an accessible PatternFly **menu** component:
+- Due to the composable nature of the menu component, keyboard interaction will need to be added manually between the menu toggle and the menu. See the [composable menu demos](https://www.patternfly.org/v4/demos/composable-menu) to see how this is done.
+- Any actions placed within a menu item will need an `aria-label` on the `MenuItemAction` to communicate to users what the action is.
+
+
+## Testing
+
+At a minimum, a menu component should meet the following criteria:
+
+<List isPlain>
+  <ListItem>
+    <Checkbox id="menu-a11y-checkbox-1" label={<span>Users should be able to use keyboard navigation to enter the menu, navigate between menu items, select an option, or escape the menu.</span>} description="<kbd>Tab</kbd> navigates to the menu, <kbd>Shift</kbd> + <kbd>Tab</kbd> navigates to the previous focusable element, up and down arrow keys should be used to navigate between menu times, and <kbd>Escape</kbd> should close the menu."/>
+  </ListItem>
+  <ListItem>
+    <Checkbox id="menu-a11y-checkbox-2" label={<span>If using menu item actions, the actions have an accessible name via the <code class="ws-code">aria-label</code></span>} description="This will differentiate each action when announced to users of assistive technologies when navigating through a page or a screen reader's rotor menu." />
+  </ListItem>
+</List>
+
+
+## React customization
+
+Various React props have been provided for more fine-tuned control over accessibility.
+
+| Prop | Applied to | Reason | 
+|---|---|---|
+| `aria-label` | `Menu` | Adds an accessible name to the menu |
+| `aria-label` | `MenuItem` | Adds an accessible name to the menu item |
+| `aria-label` | `MenuItemAction` | Adds an accessible name to the action item on the menu |
+
+
+
+## HTML/CSS customization
+| Attribute | Applied | Outcome |
+| -- | -- | -- |
+| `role="menu"` | `.pf-c-menu__list` | Declares `.pf-c-menu__list` as a menu. |
+| `disabled` | `button.pf-c-menu__item` | When the menu item uses a button element, indicates that it is unavailable and removes it from keyboard focus. |
+| `role="menuitem"` | `.pf-c-menu__item`, `.pf-c-menu__list-item` (checkbox) | Assigns `.pf-c-menu__item` as an option in a set of choices contained by a menu. |
+| `role="none"` | `.pf-c-menu__list-item` | Removes semantic meaning from `.pf-c-menu__list-item`. |
+| `aria-disabled="true"` | `a.pf-c-menu__item` | When the menu item uses a link element, removes it from keyboard focus. |
+| `tabindex="-1"` | `a.pf-c-menu__item` | When the menu item uses a link element, removes it from keyboard focus. |
+| `aria-hidden="true"` | `.pf-c-menu__item-icon`, `.pf-c-menu__item-action-icon`, `.pf-c-menu__item-external-icon`, `.pf-c-menu__item-toggle-icon`, `.pf-c-menu__item-select-icon` | Hides the icon from assistive technologies. |
+| `aria-label="Not starred"` | `.pf-c-menu__item-action-icon.pf-m-favorite` | Provides an accessible label indicating that the favorite action is not selected. |
+| `aria-label="Starred"` | `.pf-c-menu__item-action-icon.pf-m-favorite.pf-m-favorited` | Provides an accessible label indicating that the favorite action is selected. |
--- a/packages/v4/patternfly-docs/content/accessibility/menu/menu.md
+++ b/packages/v4/patternfly-docs/content/accessibility/menu/menu.md
@@ -19,7 +19,7 @@ At a minimum, a menu component should meet the following criteria:
 
 <List isPlain>
   <ListItem>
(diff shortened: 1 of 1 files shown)


packages/v4/patternfly-docs/content/accessibility/menu/menu.md
```diff
@@ -0,0 +1,53 @@
...
+- Any actions placed within a menu item will need an `aria-label` on the `MenuItemAction` to communicate to users what the action is.
+
+
+## Testing
+
+At a minimum, a menu component should meet the following criteria:
+
+<List isPlain>
+  <ListItem>
+    <Checkbox id="menu-a11y-checkbox-1" label={<span>Users should be able to use keyboard navigation to enter the menu, navigate between menu items, select an option, or escape the menu.</span>} description="<kbd>Tab</kbd> navigates to the menu, <kbd>Shift</kbd> + <kbd>Tab</kbd> navigates to the previous focusable element, up and down arrow keys should be used to navigate between menu times, and <kbd>Escape</kbd> should close the menu."/>
```
srambach, line 22: You have to put `{<span> ... </span>}` around the `<kbd>` tags to make them display correctly.

packages/v4/patternfly-docs/content/accessibility/menu/menu.md
```diff
@@ -0,0 +1,53 @@
...
+
+## React customization
+
+Various React props have been provided for more fine-tuned control over accessibility.
+
+| Prop | Applied to | Reason | 
+|---|---|---|
+| `aria-label` | `Menu` | Adds an accessible name to the menu |
+| `aria-label` | `MenuItem` | Adds an accessible name to the menu item |
+| `aria-label` | `MenuItemAction` | Adds an accessible name to the action item on the menu |
```
srambach, line 38: We are putting periods on these (I think?)

packages/v4/patternfly-docs/content/accessibility/menu/menu.md
```diff
@@ -0,0 +1,53 @@
...
+## Testing
+
+At a minimum, a menu component should meet the following criteria:
+
+<List isPlain>
+  <ListItem>
+    <Checkbox id="menu-a11y-checkbox-1" label={<span>Users should be able to use keyboard navigation to enter the menu, navigate between menu items, select an option, or escape the menu.</span>} description="<kbd>Tab</kbd> navigates to the menu, <kbd>Shift</kbd> + <kbd>Tab</kbd> navigates to the previous focusable element, up and down arrow keys should be used to navigate between menu times, and <kbd>Escape</kbd> should close the menu."/>
+  </ListItem>
+  <ListItem>
+    <Checkbox id="menu-a11y-checkbox-2" label={<span>If using menu item actions, the actions have an accessible name via the <code class="ws-code">aria-label</code></span>} description="This will differentiate each action when announced to users of assistive technologies when navigating through a page or a screen reader's rotor menu." />
```
nicolethoen, line 25: the `<code>` snippet should use `className` rather than `class`

nicolethoen (commented):
I agree this is a good foundation. There might be a day where the a11y docs need to outline the expected keyboard interactions between the menu and menu toggle, since I'm not sure that's itemized anywhere. It possible it could be in an a11y tab under the composable menu demos. The only reason I'd hesitate to do that now, is because we are in the middle of implementing the next versions of our dropdown components, and I'm not sure exactly what the future of the composable demos will be. They may be split up and each moved to the most relevant component demo pages.

srambach:
Just checking that we are being consistent - the first part of these guidelines say `MenuItemAction` must have `aria-label`.  React customization lists `aria-label` props for `Menu` and `MenuItem` as well, but in the React tab the props don't list even the `MenuItemAction aria-label` as required. On the HTML tab, we don't list anything as required either and only list `aria-label` for favorites. 
I'm guessing the a11y tab is the most correct and the other tabs might need to be updated?

jessiehuff:
I noticed that we add the `aria-label` in all of the examples with an action (which makes sense), but we don't document the prop in any of the documentation. I imagine it could make sense to add this to both the HTML and React docs.

nicolethoen:
opened good first issues in [react](https://github.com/patternfly/patternfly-react/issues/8305) and [core](https://github.com/patternfly/patternfly/issues/5198) to update the docs around aria-labels on actions

nicolethoen (dismissed):

packages/v4/patternfly-docs/content/accessibility/menu/menu.md
```diff
@@ -0,0 +1,53 @@
...
+## React customization
+
+Various React props have been provided for more fine-tuned control over accessibility.
+
+| Prop | Applied to | Reason | 
+|---|---|---|
+| `aria-label` | `Menu` | Adds an accessible name to the menu. |
+| `aria-label` | `MenuItem` | Adds an accessible name to the menu item. |
+| `aria-label` | `MenuItemAction` | Adds an accessible name to the action item on the menu. |
+
```
thatblindgeye, line 39: MenuGroup: `labelHeadingLevel` (to mention that heading levels shouldnt' be skipped in the context of the rest of the page)

MenuItem: `icon` and `isExternalLink` (similar to how they've been described elsewhere; the [AppLauncher PR](https://github.com/patternfly/patternfly-org/pull/3268/files) had one for `isExternal` on line 44 of that file)

MenuItemAction: `icon` similar to above

Additionally, just adding in some placeholder valuers similar to the other PRs

packages/v4/patternfly-docs/content/accessibility/menu/menu.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+## HTML/CSS customization
+| Attribute | Applied | Reason |
```
thatblindgeye, line 43:
```suggestion
## HTML/CSS customization

The following HTML attributes and PatternFly classes can be used for more fine-tuned control over accessibility.

| Attribute or class | Applied | Reason |
```

thatblindgeye (changes requested):

packages/v4/patternfly-docs/content/accessibility/menu/menu.md
```diff
@@ -0,0 +1,53 @@
...
+</List>
+
+
+## React customization
+
+Various React props have been provided for more fine-tuned control over accessibility.
+
+| Prop | Applied to | Reason | 
+|---|---|---|
+| `aria-label` | `Menu` | Adds an accessible name to the menu. |
```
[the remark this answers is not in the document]
jessiehuff: Adding an `aria-label` to the Menu could make sense in some situations, but I'm not sure I'd consider it required? We don't have it on any of our examples/demos. Do you think it's something we should be adding?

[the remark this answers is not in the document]
thatblindgeye: Looking into it some more, an `aria-label` on the Menu probably wouldn't be required. An `aria-label` on the MenuList sub-component might be worth requiring since placing it there would help differentiate the menu in a screen reader rotor.

That would mean updating some examples + React code, possibly as a breaking change since it might make more sense to remove the aria-label from the Menu and just move it to MenuList. I'd be good with keeping this as-is for now, and it can be revisited if we want to make these updates in React and examples.

packages/v4/patternfly-docs/content/accessibility/menu/menu.md
```diff
@@ -0,0 +1,61 @@
...
+| `aria-label=[text labeling the menu item]` | `MenuItem` | Adds an accessible name to the menu item. |
+| `icon={<Icon aria-hidden />}`  | `MenuItem` | Renders an icon to a menu item. When icon is decorative, pass an `aria-hidden` to the icon to hide it from screen reader users. |
+| `isExternalLink` | `MenuItem` | Renders an icon on the item when it receives focus or is hovered, as well as visually hidden text, to let users of certain assistive technologies know that the item will open in a new tab/window. **Required if the item opens a new tab/window**. |
+| `aria-label=[text labeling the menu item's action]` | `MenuItemAction` | Adds an accessible name to the action item on the menu. |
+|`icon={<Icon aria-hidden />}` | `MenuItemAction` | Renders an icon to a menu item action. When icon is decorative, pass an `aria-hidden` to the icon to hide it from screen reader users. |
+| `labelHeadingLevel=[a heading tag]` | `MenuGroup` | Adds a group label's heading level. Default is `h1`. Be sure to not skip heading levels when passing this prop in. For example, if an accordion is within a page section that has an `h2` heading, you should not pass `headingLevel="h4"`. |
```
thatblindgeye, line 42:
```suggestion
| `labelHeadingLevel=[a heading tag]` | `MenuGroup` | Adds a group label's heading level. Default is `h1`. Be sure to not skip heading levels when passing this prop in. For example, if a menu group is within a page section that has an `h2` heading, you should not pass `headingLevel="h4"`. |
```

thatblindgeye (changes requested):
One last quick-fix, but this is all looking great!

thatblindgeye (approved):

srambach (approved):
LGTM 👍

PR #3265, opened 2022-10-26
docs(MenuToggle): add accessibility documentation

### LINKED ISSUE #3264: Add a11y documentation for Menu Toggle

jessiehuff:
Closes #3264

### DIFF
--- a/packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
+++ b/packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
@@ -50,6 +50,6 @@ The following HTML attributes and PatternFly classes can be used for more fine-t
 
 | Attribute or Class | Applied to | Reason |
 | -- | -- | -- |
-| `aria-expanded="true|false"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the expanded or collapsed state. |
+| `aria-expanded="true or false"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the expanded or collapsed state. |
 | `aria-label="[text describing the menu toggle]"` | `.pf-c-menu-toggle` | An accessible label for the menu toggle when there is no visible text within the toggle. Required when using a `pf-m-plain` with no text. |
 | `aria-hidden="true"` | `.pf-c-menu-toggle__toggle-icon` | Makes the decorative icon on the toggle hidden from screen reader users. **Required when the toggle contains only an icon**| 
--- a/packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
+++ b/packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
@@ -50,7 +50,6 @@ The following HTML attributes and PatternFly classes can be used for more fine-t
 
 | Attribute or Class | Applied to | Reason |
 | -- | -- | -- |
-| `aria-expanded="true"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the expanded state. |
-| `aria-expanded="false"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the collapsed state. |
+| `aria-expanded="true|false"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the expanded or collapsed state. |
 | `aria-label="[text describing the menu toggle]"` | `.pf-c-menu-toggle` | An accessible label for the menu toggle when there is no visible text within the toggle. Required when using a `pf-m-plain` with no text. |
 | `aria-hidden="true"` | `.pf-c-menu-toggle__toggle-icon` | Makes the decorative icon on the toggle hidden from screen reader users. **Required when the toggle contains only an icon**| 
--- a/packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
+++ b/packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
@@ -37,10 +37,10 @@ The following React props have been provided for more fine-tuned control over ac
 
 | Prop | Applied to | Reason | 
 |---|---|---|
-| `aria-label="[text describing the menu toggle]"` | `MenuToggle` | Adds an accessible name to the menu toggle. |
+| `aria-label="[text describing the menu toggle]"` | `MenuToggle` | Adds an accessible name to the menu toggle. Required when there is no text displayed in the menu toggle. |
 | `icon={[custom icon]}` | `MenuToggle` | Adds an optional icon rendered inside the toggle, before the children content. Add `aria-hidden` to the icon to hide the decorative item from screen reader elements. |
-| `aria-label="[text describing the action]"` | `MenuToggleAction` | Adds an accessible name to the menu toggle action. |
-| `aria-label="[text labeling the checkbox]"` | `MenuToggleCheckbox` | Adds an accessible name to the checkbox item on the menu toggle. |
+| `aria-label="[text describing the action]"` | `MenuToggleAction` | Adds an accessible name to the menu toggle action when the action is represented by an icon and no text. |
+| `aria-label="[text labeling the checkbox]"` | `MenuToggleCheckbox` | Adds an accessible name to the checkbox item on the menu toggle when there is no associated visible label text. |
 
 
 
@@ -52,5 +52,5 @@ The following HTML attributes and PatternFly classes can be used for more fine-t
 | -- | -- | -- |
 | `aria-expanded="true"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the expanded state. |
 | `aria-expanded="false"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the collapsed state. |
-| `aria-label="[text describing the menu toggle]"` | `.pf-c-menu-toggle.pf-m-plain` | Gives the plain menu toggle component an accessible label. |
+| `aria-label="[text describing the menu toggle]"` | `.pf-c-menu-toggle` | An accessible label for the menu toggle when there is no visible text within the toggle. Required when using a `pf-m-plain` with no text. |
 | `aria-hidden="true"` | `.pf-c-menu-toggle__toggle-icon` | Makes the decorative icon on the toggle hidden from screen reader users. **Required when the toggle contains only an icon**| 
--- a/packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
+++ b/packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
@@ -20,7 +20,7 @@ At a minimum, a menu toggle component should meet the following criteria:
 
 <List isPlain>
   <ListItem>
-    <Checkbox id="menutoggle-a11y-checkbox-1" label={<span>Users should be able to use keyboard navigation to activate the menu toggle with <kbd>Enter</kbd> or <kbd>Space</kbd> to enter a menu.</span>} description={<span><kbd>Tab</kbd> navigates to the menu and <kbd>Shift</kbd> + <kbd>Tab</kbd> navigates to the previous focusable element. Refer to our [menu accessibility documentation](/components/menu-toggle/accessibility) for more implementation details when using these components together. </span>}/>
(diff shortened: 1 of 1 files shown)


packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,52 @@
...
+section: components
+---
+
+import { Checkbox, List, ListItem } from '@patternfly/react-core';
+
+
+## Accessibility
+
+To implement an accessible PatternFly **menu toggle** component:
+- Due to the composable nature of the menu and menu toggle components, keyboard interaction will need to be added manually between the menu toggle and the menu. See the [composable menu demos](https://www.patternfly.org/v4/demos/composable-menu) to see how this is done.
```
thatblindgeye, line 12: For the URL, we should be able to use `/demos/composable-menu` instead

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,52 @@
...
+- When adding any additional elements to the menu toggle, remember to follow that element's specific considerations. For example, if you are adding an avatar to the menu toggle, you will need to provide an `alt` attribute with a description of the image.
+
+
+## Testing
+
+At a minimum, a menu component should meet the following criteria:
+
+<List isPlain>
+  <ListItem>
+    <Checkbox id="menutoggle-a11y-checkbox-1" label={<span>Users should be able to use keyboard navigation to activate the menu toggle with <kbd>Enter</kbd> or <kbd>Space</kbd> to enter a menu, navigate between menu items, select an option, or escape the menu.</span>} description={<span><kbd>Tab</kbd> navigates to the menu, <kbd>Shift</kbd> + <kbd>Tab</kbd> navigates to the previous focusable element, up and down arrow keys should be used to navigate between menu times, and <kbd>Escape</kbd> should close the menu."</span>}/>
```
thatblindgeye, line 23: I think we could omit the verbiage related to menu/menu items and just keep the Tab/Shift + Tab, and the Enter/Space verbiage. Maybe instead placing a link to the Menu a11y docs? If we got both this and the Menu PR in at the same time, there wouldn't be any issue of the link not working on the site itself.

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,52 @@
...
+  </ListItem>
+  <ListItem>
+    <Checkbox id="menutoggle-a11y-checkbox-3" label={<span>Any action used within a menu toggle has an <code className="ws-code">aria-label</code>.</span>} description="This will differentiate each action when announced to users of assistive technologies when navigating through a page or a screen reader's rotor menu." />
+  </ListItem>
+</List>
+
+
+## React customization
+
+Various React props have been provided for more fine-tuned control over accessibility.
```
thatblindgeye, line 36: Calling this out since @srambach mentioned it on Slack. Do we want to see if content has a recommendation of "Various such and such..." vs "The following such and such..."? (Personally I'd be fine with switching to "The following")

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,52 @@
...
+
+| Prop | Applied to | Reason | 
+|---|---|---|
+| `aria-label` | `MenuToggle` | Adds an accessible name to the menu toggle. |
+| `aria-label` | `MenuToggleAction` | Adds an accessible name to the menu toggle action. |
+| `aria-label` | `MenuToggleCheckbox` | Adds an accessible name to the checkbox item on the menu toggle. |
+
```
thatblindgeye, line 43: The `icon` prop could possibly be mentioned here, mainly to mention that the icon should have aria-hidden (similar table item would need to be added to the HTML/CSS table)

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+## HTML/CSS customization
+| Class | Applied to | Reason |
```
thatblindgeye, line 47: `| Attribute or class | Applied to | Reason |`

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,52 @@
...
+| -- | -- | -- |
+| `aria-expanded="true"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the expanded state. |
+| `aria-expanded="false"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the collapsed state. |
+| `aria-label` | `.pf-c-menu-toggle.pf-m-plain` | Gives the plain menu toggle component an accessible label. |
+| `disabled` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is disabled. |
```
thatblindgeye, line 52: Wondering if this should be kept in... I know I had included this in the Tabs docs, though that was mainly due differentiating it from aria-disabled/when to use one over the other.

thatblindgeye (changes requested):

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,53 @@
...
+- When adding any additional elements to the menu toggle, remember to follow that element's specific considerations. For example, if you are adding an avatar to the menu toggle, you will need to provide an `alt` attribute with a description of the image.
+
+
+## Testing
+
+At a minimum, a menu toggle component should meet the following criteria:
+
+<List isPlain>
+  <ListItem>
+    <Checkbox id="menutoggle-a11y-checkbox-1" label={<span>Users should be able to use keyboard navigation to activate the menu toggle with <kbd>Enter</kbd> or <kbd>Space</kbd> to enter a menu.</span>} description={<span><kbd>Tab</kbd> navigates to the menu and <kbd>Shift</kbd> + <kbd>Tab</kbd> navigates to the previous focusable element. Refer to our [menu accessibility documentation](/components/menu-toggle/accessibility) for more implementation details when using these components together. </span>}/>
```
srambach, line 23: I think this link needs to be an href

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,53 @@
...
+| `aria-label` | `MenuToggleCheckbox` | Adds an accessible name to the checkbox item on the menu toggle. |
+
+
+
+## HTML/CSS customization
+| Attribute or Class | Applied to | Reason |
+| -- | -- | -- |
+| `aria-expanded="true"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the expanded state. |
+| `aria-expanded="false"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the collapsed state. |
+| `aria-label` | `.pf-c-menu-toggle.pf-m-plain` | Gives the plain menu toggle component an accessible label. |
```
srambach, line 52: Same kind of comment as on the Menu docs - the aria-label guidelines don't seem to be the same between the React props listed and adding attributes to the HTML.

thatblindgeye (approved):
🎉 Only other thing is to rebase, but this looks great!

nicolethoen (dismissed):

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,56 @@
...
+</List>
+
+
+## React customization
+
+The following React props have been provided for more fine-tuned control over accessibility.
+
+| Prop | Applied to | Reason | 
+|---|---|---|
+| `aria-label="[text describing the menu toggle]"` | `MenuToggle` | Adds an accessible name to the menu toggle. |
```
nicolethoen, line 40: Can we update this to say something like "Adds an accessible name to the menu toggle. Required when there is no text displayed in the menu toggle."

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,56 @@
...
+| `icon={[custom icon]}` | `MenuToggle` | Adds an optional icon rendered inside the toggle, before the children content. Add `aria-hidden` to the icon to hide the decorative item from screen reader elements. |
+| `aria-label="[text describing the action]"` | `MenuToggleAction` | Adds an accessible name to the menu toggle action. |
```
nicolethoen, line 42: Can we update this to say something like "Adds an accessible name to the menu toggle action when the action is represented by an icon with no text"

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,56 @@
...
+| `aria-label="[text labeling the checkbox]"` | `MenuToggleCheckbox` | Adds an accessible name to the checkbox item on the menu toggle. |
```
nicolethoen, line 43: can we add - "when there is no associated visible label text"

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,56 @@
...
+
+## HTML/CSS customization
+
+The following HTML attributes and PatternFly classes can be used for more fine-tuned control over accessibility.
+
+| Attribute or Class | Applied to | Reason |
+| -- | -- | -- |
+| `aria-expanded="true"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the expanded state. |
+| `aria-expanded="false"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the collapsed state. |
+| `aria-label="[text describing the menu toggle]"` | `.pf-c-menu-toggle.pf-m-plain` | Gives the plain menu toggle component an accessible label. |
```
nicolethoen, line 55: I don't think the aria-label here should only be used when the `pf-m-plain` modifier is used. It'd be more accurate to say "An accessible label for the menu toggle when there is no visible text within the toggle. Required when using a `pf-m-plain` with no text.

nicolethoen (commented):
I updated a related issue to include cleaning up the aria-label guidance in the MenuToggle docs: https://github.com/patternfly/patternfly/issues/5198

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
jessiehuff: Agreed, another one that I copied from core

nicolethoen (approved):

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+## HTML/CSS customization
+
+The following HTML attributes and PatternFly classes can be used for more fine-tuned control over accessibility.
+
+| Attribute or Class | Applied to | Reason |
+| -- | -- | -- |
+| `aria-expanded="true"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the expanded state. |
+| `aria-expanded="false"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the collapsed state. |
```
srambach, line 54: Sorry to add one more thing! But if we have true/false on the same component can we combine it into one line to simplify?

nicolethoen (dismissed):

packages/v4/patternfly-docs/content/accessibility/menu-toggle/menu-toggle.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+
+## HTML/CSS customization
+
+The following HTML attributes and PatternFly classes can be used for more fine-tuned control over accessibility.
+
+| Attribute or Class | Applied to | Reason |
+| -- | -- | -- |
+| `aria-expanded="true|false"` | `.pf-c-menu-toggle`, `.pf-c-menu-toggle__button` | Indicates that the menu toggle component is in the expanded or collapsed state. |
```
thatblindgeye, line 53: We'd probably need to use `true or false` rather than a vertical pipe (or `\|` to try escaping, but I think I've not had better results with that) otherwise we end up with something like this:

menu toggle table with incorrect cells

thatblindgeye (approved):

srambach (approved):
🌟

PR #3277, opened 2022-11-08
chore(Badge): update a11y docs

### LINKED ISSUE #3276: Update a11y docs for Badge

thatblindgeye:
Closes #3276

The a11y docs already had some verbiage about not using only color which I mostly kept in tact, but I also added some verbiage about adding a `pf-screen-reader` class to add additional context for users navigating via AT. Depending on whether we'd want to keep that, it might be worth either updating the Badge examples and/or maybe adding a prop in React for adding in `screenReaderText`.

### DIFF
--- a/packages/v4/patternfly-docs/content/accessibility/badge/badge.md
+++ b/packages/v4/patternfly-docs/content/accessibility/badge/badge.md
@@ -9,8 +9,8 @@ import { Checkbox, List, ListItem } from '@patternfly/react-core';
 
 To implement an accessible PatternFly **badge**:
 
-- Ensure the badge itself is not interactive in anyway. 
-- Ensure that the surrounding context can convey to users via screen reader or similar assistive technologies the purpose of the badge.
+- Ensure the badge itself is not interactive in any way. 
+- Ensure that the surrounding context can convey the purpose of the badge to users via screen reader or similar assistive technologies.
 - Ensure badge information is not conveyed by only color, such as a read or unread badge.
 
 ## Testing
@@ -39,7 +39,7 @@ A badge does not have any further HTML/CSS attributes or classes for accessibili
 
 ## Additional considerations
 
-Consumers must ensure they take any additional considerations when customizing a badge, using it in a way not described or recommended by PatternFly, or in various other specific use-cases not outlined elsewhere on this page.
+Consumers must ensure they take any additional considerations when customizing a badge, using it in a way not described or recommended by PatternFly, or in various other specific use cases not outlined elsewhere on this page.
 
 ### Color
 
@@ -47,12 +47,12 @@ Generally you should avoid using only color to convey information about a badge.
 
 If a badge will always be rendered whether it's unread or read, or different badges will convey different types of information, it may be difficult for users to perceive or understand the difference between the two badges.
 
-There are two ways in which to supplement color in order to more accessibly convey badge information:
+You should do the following to supplement color to convey badge information:
 
-- Adding an icon to visually convey information
-- Adding a visually-hidden element via the `pf-screen-reader` class to convey information to users of screen readers and similar assistive technologies
+- Add an icon to visually convey information
+- Add a visually-hidden element via the `pf-screen-reader` class to convey information to users of screen readers and similar assistive technologies
 
-Note how difficult it is to determine that there are unread or actionable items associated with either of the badges in Figure 1 with its lack of visual cues. Compare Figure 1 to Figure 2 where there is an icon used to indicate that the second badge is associated with actionable or unread content.
+In Figure 1, note that the lack of visual cues make it difficult to differentiate unread or actionable items associated with the badges. Compare this to Figure 2, where an icon indicates that the second badge is associated with actionable or unread content.
 
 Figure 1.
 
--- a/packages/v4/patternfly-docs/content/accessibility/badge/badge.md
+++ b/packages/v4/patternfly-docs/content/accessibility/badge/badge.md
@@ -10,8 +10,15 @@ import { Checkbox, List, ListItem } from '@patternfly/react-core';
 To implement an accessible PatternFly **badge**:
 
 - Ensure the badge itself is not interactive in any way. 
-- Ensure that the surrounding context can convey the purpose of the badge to users via screen reader or similar assistive technologies.
 - Ensure badge information is not conveyed by only color, such as a read or unread badge.
+- Ensure that the surrounding context can convey the purpose of the badge to users via screen reader or similar assistive technologies.
+  - A heading or other text element prefacing the badge or visually hidden text content accessible only to assistive technologies are some examples of a surrounding context.
+
+  ```
+  <h2>Notificaitons <Badge>5</Badge></h2>
+
+  <Badge>5 <span class="pf-screen-reader">unread notifications</span></Badge>
+  ```
 
 ## Testing
 
@@ -22,10 +29,10 @@ To implement an accessible PatternFly **badge**:
     <Checkbox id="badge-a11y-checkbox-1" label="Users should not be able to interact with or navigate to the badge with the keyboard, as it is not an interactive element." />
   </ListItem>
   <ListItem>
-    <Checkbox id="badge-a11y-checkbox-2" label="Users navigating via screen reader or similar assistive technology should understand the purpose of the badge by its surrounding context." />
+    <Checkbox id="badge-a11y-checkbox-2" label="Information about a badge is not conveyed by only color."/>
   </ListItem>
   <ListItem>
-    <Checkbox id="badge-a11y-checkbox-3" label="Information about a badge is not conveyed by only color."/>
+    <Checkbox id="badge-a11y-checkbox-3" label="Users navigating via screen reader or similar assistive technology should understand the purpose of the badge by its surrounding context." />
   </ListItem>
 </List>
 
@@ -43,7 +50,7 @@ Consumers must ensure they take any additional considerations when customizing a
 
 ### Color
 
-Generally you should avoid using only color to convey information about a badge. One possible exception to this rule may be if a badge is only ever rendered in a specific scenario, such as only when there are unread notifications, and will otherwise not be rendered for any other purpose. Caution should still be used when using color in this way, though.
+Generally you should avoid using only color to convey information about different types of badges, such as between read and unread notifications.
 
 If a badge will always be rendered whether it's unread or read, or different badges will convey different types of information, it may be difficult for users to perceive or understand the difference between the two badges.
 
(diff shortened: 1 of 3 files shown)


nicolethoen:
> it might be worth either updating the Badge examples and/or maybe adding a prop in React for adding in `screenReaderText`.

This seems reasonable to me. Maybe that's a great issue to label with 'good first issue' or assign to JHU students.

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
```diff
@@ -3,30 +3,61 @@ id: Badge
...
+import { Checkbox, List, ListItem } from '@patternfly/react-core';
 
-**Keyboard users** should not be able to focus on a badge.
+## Accessibility
 
-**Screen reader users** should be able to have a screen reader describe the contents of and purpose of a badge when it pops up, but not focus on it otherwise.
+To implement an accessible PatternFly **badge**:
 
-No props need to be added or modified for badge accessibility.
+- Ensure the badge itself is not interactive in anyway.
```
srambach, line 12: "in any way." (space)

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
```diff
@@ -3,30 +3,61 @@ id: Badge
...
 
-**Keyboard users** should not be able to focus on a badge.
+## Accessibility
 
-**Screen reader users** should be able to have a screen reader describe the contents of and purpose of a badge when it pops up, but not focus on it otherwise.
+To implement an accessible PatternFly **badge**:
 
-No props need to be added or modified for badge accessibility.
+- Ensure the badge itself is not interactive in anyway. 
+- Ensure that the surrounding context can convey to users via screen reader or similar assistive technologies the purpose of the badge.
```
srambach, line 13: Rearrange to make the sentence easier to read (bold just showing the change)- Ensure that the surrounding context can convey **the purpose of the badge** to users via screen reader or similar assistive technologies.

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
```diff
@@ -3,30 +3,61 @@ id: Badge
...
+
+A badge does not have any further React props for accessibility.
+
+## HTML/CSS customization
+
+A badge does not have any further HTML/CSS attributes or classes for accessibility.
+
+## Additional considerations
+
+Consumers must ensure they take any additional considerations when customizing a badge, using it in a way not described or recommended by PatternFly, or in various other specific use-cases not outlined elsewhere on this page.
```
srambach, line 42: I don't think "use-case" should be hypenated.

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
```diff
@@ -3,30 +3,61 @@ id: Badge
...
+### Color
+
+Generally you should avoid using only color to convey information about a badge. One possible exception to this rule may be if a badge is only ever rendered in a specific scenario, such as only when there are unread notifications, and will otherwise not be rendered for any other purpose. Caution should still be used when using color in this way, though.
+
+If a badge will always be rendered whether it's unread or read, or different badges will convey different types of information, it may be difficult for users to perceive or understand the difference between the two badges.
+
+There are two ways in which to supplement color in order to more accessibly convey badge information:
+
+- Adding an icon to visually convey information
+- Adding a visually-hidden element via the `pf-screen-reader` class to convey information to users of screen readers and similar assistive technologies
```
srambach, line 53: It's not clear to me if these both should be done or two ways of handling it.
Also, I'd use "Add" rather than "Adding" and then put a period at the end of each bullet. But that might be subjective.

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
```diff
@@ -3,30 +3,61 @@ id: Badge
...
+
+Note how difficult it is to determine that there are unread or actionable items associated with either of the badges in Figure 1 with its lack of visual cues. Compare Figure 1 to Figure 2 where there is an icon used to indicate that the second badge is associated with actionable or unread content.
```
srambach, line 55: Suggested rewording: 
In Figure 1, note that the lack of visual cues make it difficult to differentiate unread or actionable items associated with the badges. Compare this to Figure 2, where an icon indicates that the second badge is associated with actionable or unread content.

srambach (approved):
The content is good ⭐ but I had some wording nits.

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
thatblindgeye: Good catch! Updating a couple of other files that hyphenated it ("use case" is used a lot more in other files in org)

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
thatblindgeye: Ah that's fair. I could update the prefacing text to something along the lines of, "To supplement color in order to more accessibly convey badge information you should:" wdyt?

(and updating to "Add" is 👍🏼 to me)

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
```diff
@@ -3,30 +3,61 @@ id: Badge
...
+
+A badge does not have any further HTML/CSS attributes or classes for accessibility.
+
+## Additional considerations
+
+Consumers must ensure they take any additional considerations when customizing a badge, using it in a way not described or recommended by PatternFly, or in various other specific use cases not outlined elsewhere on this page.
+
+### Color
+
+Generally you should avoid using only color to convey information about a badge. One possible exception to this rule may be if a badge is only ever rendered in a specific scenario, such as only when there are unread notifications, and will otherwise not be rendered for any other purpose. Caution should still be used when using color in this way, though.
```
jessiehuff, line 46: So I'm not completely sure what you mean here. Are you trying to say in a scenario in which the color of the badge doesn't genuinely matter for understanding?

thatblindgeye: More or less, yeah. This could probably just be condensed to something like, "Generally you should avoid using only color to convey information about different types of badges, such as between read and unread notifications."

Do you have a suggestion to update this verbiage?

srambach: Although color is generally an accessibility issue, I think the high contrast of the two types of badge plus accessible labeling addresses a11y issues. The understandability of which badge means what seems to be more of a _usability_ issue, so perhaps that part could be moved over to the design tab?

thatblindgeye: That's true... I could see the info technically working in either place. The Badge now supporting a `screenReaderText` prop in React and examples showing that visually hidden text in action definitely help, and I can update both the HTMl and React tables to mention the appropriate attribute/class/prop. 

I could also update the testing item about color to add a description mentioning how icons and visually hidden text can/should be used. @jessiehuff @srambach wdyt?

nicolethoen (dismissed):

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
```diff
@@ -3,30 +3,61 @@ id: Badge
...
 
-Note how difficult it is to determine that there are unread or actionable items associated with either of the badges
-in Figure 1 with its lack of visual cues. Compare Figure 1 to Figure 2 where there is an icon
-used to indicate that the second badge is associated with actionable or unread content.
+<List isPlain>
+  <ListItem>
+    <Checkbox id="badge-a11y-checkbox-1" label="Users should not be able to interact with or navigate to the badge with the keyboard, as it is not an interactive element." />
+  </ListItem>
+  <ListItem>
+    <Checkbox id="badge-a11y-checkbox-2" label="Users navigating via screen reader or similar assistive technology should understand the purpose of the badge by its surrounding context." />
```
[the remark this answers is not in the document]
jessiehuff: What do you think could be a good example of context? Perhaps if there is visible text around the badge or a label on the badge to explain? Like "7 notifications unread" or an example like that? I still think that a lot of people won't understand what we mean by the context here, and if we don't explain, they're likely to just skip over it. Wdyt?

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
jessiehuff: Also, I do think that it could make sense in both sections, but I've always found it important to remind users not to use colors to communicate information in different use cases. I think a lot of designers and developers forget about how it affects a11y so I think it's important that they're not forgetting to add contextual information, either in the form of surrounding UI or additional hidden text with labels or screen reader text.

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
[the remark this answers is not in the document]
thatblindgeye: Definitely agree. Maybe for the description we can add, "Some examples of surrounding context can include the badge's aria-label or a heading or other text content prefacing the badge." or "For example, an aria-label on the bage with a value of "unread notifications", or a text element immediately before the badge stating "Notifications"", or some variation of either?

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
thatblindgeye: @jessiehuff I'd be for omitting the second bit and just keeping the "Generally you should avoid using only color to convey information about different types of badges, such as between read and unread notifications" portion. Do you think the paragraph that follows ("If a badge will always be rendered whether it's unread or read...") is fine as-is, or could more be tweaked in this section?

jessiehuff: Yeah, I think that makes sense. I'm open to suggestions though if you or others think differently. :)

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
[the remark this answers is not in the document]
jessiehuff: Sounds perfect! We probably don't need to add it to both sections. If we add the examples in the top section, they'll probably understand the testing checklist item.

packages/v4/patternfly-docs/content/accessibility/badge/badge.md
thatblindgeye: @jessiehuff let me know what you think of the latest updates! I added some verbiage + a couple of quick examples re: context to the first section, and updated the Colors section to begin with the verbiage you commented with here

nicolethoen (approved):

jessiehuff (approved):
⭐

PR #3287, opened 2022-11-14
chore(Button): updated a11y docs

### LINKED ISSUE #3286: Update a11y docs for Button

thatblindgeye:
Closes #3286 

The HTML a11y section mentions `aria-pressed`. Would it be worth keeping this verbiage and adding a prop to React? Currently the attribute isn't used anywhere for the component, but React does have examples of native button elements being used with `aria-pressed` passed in, so if we ever plan to update those examples to use our component instead it'd make sense.


srambach (approved):
Looks good to me 👍

thatblindgeye:
@srambach I just updated the first comment too 😆 If you have any thoughts about what I edited in that'd be great, but if not no worries!

nicolethoen (approved):

PR #3291, opened 2022-11-14
feat: add deprecated subpaths

### LINKED ISSUE #3289: add 'deprecated' sub-paths for deprecated tabs
add 'deprecated' sub-paths for deprecated tabs in the documentation-framework

nicolethoen:
Closes #3289


packages/documentation-framework/components/example/example.js
```diff
@@ -105,7 +105,8 @@ export const Example = ({
     // These 2 are in the bundle anyways for the site since we dogfood
     ...reactCoreModule,
     ...reactTableModule,
-    ...(source === 'react-next' ? reactCoreNextModule : {})
+    ...(source === 'react-next' ? reactCoreNextModule : {}),
+    ...(source === 'react-deprecated' ? reactCoreDeprecatedModule : {})
```
jeffpuzzo, line 109: It doesn't look like `reactCoreDeprecatedModule` is defined, and I think we want to extend upon `react-table`'s module for this scope and not `react-core` (I assume we'll want both eventually, but for [this change](https://github.com/patternfly/patternfly-react/pull/8362), at least `react-table`).

nicolethoen: truee

packages/documentation-framework/routes.js
```diff
@@ -50,7 +50,8 @@ const sourceOrder = {
   react: 1,
   'react-next': 1.1,
   'react-composable': 1.2,
-  'react-legacy': 1.3,
+  'react-deprecated': 1.3,
+  'react-legacy': 1.4,
```
jeffpuzzo, line 54: Do we still have plans to use `react-legacy` after this change and introduction of `react-deprecated`?

packages/documentation-framework/scripts/md/parseMD.js
```diff
@@ -281,9 +281,14 @@ function writeIndex() {
 }
 
 // Build unique names for components if they are a part of the "next" module.
```
jeffpuzzo, line 283: If we keep this function as you've expanded upon it, we should update this comment.

packages/documentation-framework/scripts/md/parseMD.js
```diff
@@ -281,9 +281,14 @@ function writeIndex() {
...
-function getTsDocName(name, isNextComponent) {
-  return `${name}${isNextComponent ? '-next' : ''}`;
-};
+function getTsDocName(name, isNextComponent, isDeprecatedComponent) {
```
jeffpuzzo, line 284: So as to not have to add new parameters to this function whenever a new condition arises, we might want to consider reworking this a bit. This is what I had in mind roughly, where if there's another "variant" at any point, we just have to update the 1 location in `getTsDocNameVariant`:

```
// Build unique names for components with a "variant" extension
function getTsDocName(name, variant) {
  return `${name}${variant ? `-${variant}` : ''}`;
}

function getTsDocNameVariant(source) {
  if (source.includes('next')) {
    return 'next';
  }

  if (source.includes('deprecated')) {
    return 'deprecated';
  }
}
```

and the 2 usages above would then look like the following:
`tsDocs[getTsDocName(name, getTsDocNameVariant(file))] = { name, description, props };`
and
`const name = getTsDocName(componentName, getTsDocNameVariant(source));`

packages/documentation-framework/routes.js
nicolethoen: ultimately react-deprecated will replace the legacy here so I can remove it, but probably not until it's been removed after the breaking change. I don't want to remove it before then.
The deprecated work is going in the v5 branch

jeffpuzzo: Got it, so maybe we make a follow-up issue to remove this when we're ready?

jeffpuzzo (approved):
Looks good. Once Jenny's PR is merged; https://github.com/patternfly/patternfly-react/pull/8362, this build should pass.

packages/v4/patternfly-docs/patternfly-docs.source.js
```diff
@@ -56,9 +56,10 @@ module.exports = (sourceMD, sourceProps) => {
   sourceMD(path.join(reactCorePath, '/**/demos/**/*.md'), 'react-demos');
 
   // React-table MD
-  sourceMD(path.join(reactTablePath, '/**/TableComposable/examples/*.md'), 'react-composable');
-  sourceMD(path.join(reactTablePath, '/**/demos/*.md'), 'react-demos');
+  sourceMD(path.join(reactTablePath, '/**/TableComposable/examples/*.md'), 'react');
```
jeffpuzzo, line 59: Was this tested locally by any chance? This regex might conflict with the deprecated path regex on line 61. To prevent the potential for that we can remove the wildcard at the beginning of this line and replace it with `components` (e.g. `/components/TableComposable/...`).

nicolethoen: not recently but i'll do that

nicolethoen: @jeffpuzzo so i'm realizing now that this line will be better tested in the react-docs package where we have our own patternfly-docs.source.js file defined. patternfly-react does not consume org's v4 package, it's actually the other way around, so if these lines dont work in the react workspace when the change is made there, then we can update it here to match whatever worked in the workspace. If this is good to be merged, hopefully this will unblock the work to deprecate table

jeffpuzzo: Ah right, that makes sense.

packages/documentation-framework/scripts/webpack/webpack.base.config.js
```diff
@@ -119,7 +119,9 @@ module.exports = (_env, argv) => {
         'client-styles': path.resolve(process.cwd(), 'patternfly-docs/patternfly-docs.css.js'),
         './routes-client': path.resolve(process.cwd(), 'patternfly-docs/patternfly-docs.routes.js'),
         './routes-generated': path.resolve(process.cwd(), 'patternfly-docs/generated/index.js'),
-        '@patternfly/react-core/next': '@patternfly/react-core/dist/esm/next/index.js' // Can remove when webpack is updated to v5
+        '@patternfly/react-core/next': '@patternfly/react-core/dist/esm/next/index.js', // Can remove when webpack is updated to v5
+        '@patternfly/react-core/deprecated': '@patternfly/react-core/dist/esm/deprecated/index.js', // Can remove when webpack is updated to v5
+        '@patternfly/react-table/deprecated': '@patternfly/react-table/dist/esm/deprecated/index.js' // Can remove when webpack is updated to v5
```
jeffpuzzo, line 124: I think all of these path aliases can be removed now since we're no longer using package exports in react-core/table.

packages/v4/patternfly-docs/patternfly-docs.source.js
```diff
@@ -54,24 +56,30 @@ module.exports = (sourceMD, sourceProps) => {
   sourceProps(path.join(reactCodeEditorPath, '/**/*.tsx'),reactPropsIgnore);
   sourceProps(path.join(reactChartsPath, '/**/*.tsx'),reactPropsIgnore);
   sourceProps(path.join(reactLogViewerPath, '/**/*.tsx'), reactPropsIgnore);
+  //sourceProps(path.join(reactDragDropPath, '/**/*.tsx'), reactPropsIgnore);
```
jeffpuzzo, line 59: Maybe for now, unless we want to add TODO comments explaining when we'd add this back in, we can remove this commented out line along with lines 43-44?

packages/documentation-framework/components/example/example.js
```diff
@@ -109,7 +111,9 @@ export const Example = ({
     // These 2 are in the bundle anyways for the site since we dogfood
     ...reactCoreModule,
     ...reactTableModule,
-    ...(source === 'react-next' ? reactCoreNextModule : {})
+    ...(source === 'react-next' ? reactCoreNextModule : {}),
+    ...(source === 'react-deprecated' ? reactCoreDeprecatedModule : {}),
+    ...(source === 'react-deprecated' ? reactTableDeprecatedModule : {})
```
jeffpuzzo, line 116: I think we can use the condition `source === 'react-deprecated'` just once for the inclusion of both of these deprecated modules.

jeffpuzzo (approved):
Tested this locally against patternfly-react successfully. Code changes LGTM.

jenny-s51 (approved):
Looks great @nicolethoen !! 🚀

Also tested locally and I am able to now add demos under Table -> react-deprecated .

PR #3293, opened 2022-11-15
chore(Checkbox): updated a11y docs

### LINKED ISSUE #3292: Update a11y docs for Checkbox

thatblindgeye:
Closes #3292 

I included verbiage about the checkbox input size (cc @mcoker since there is that core issue about checkbox styling, do you think we'd update our checkbox component to use a more custom styling that could allow the checkbox to be larger?) and placing related checkboxes in list elements. Currently our examples/demos don't place multiple checkboxes in list elements which would require a followup wherever necessary if we want to keep that verbiage.


packages/v4/patternfly-docs/content/accessibility/checkbox/checkbox.md
```diff
@@ -3,17 +3,78 @@ id: Checkbox
...
+## Accessibility
 
-In general, the checkbox component already has accessibility built in. The **standalone input variant** will need special consideration.
+To implement an accessible PatternFly **checkbox** component:
 
-To make the checkbox **standalone input variant** accessible, add an `aria-label` with a label for the input. You can also use `aria-labelledby` to associate the checkbox with visible text on the page to explain the checkbox. 
+- Ensure the checkbox can be navigated to and interacted with via keyboard and other assistive technologies such as a screen reader
+- Ensure the checkbox can be toggled by clicking its visible text label, if it has one
+- Ensure the checkbox label has hover styles to indicate that it can be clicked
+- Ensure the checkbox input provides enough clickable space on its own
```
mcoker, line 15: I think it would be good to include some kind of standard/criteria for this size, and to make sure we're implementing it the same way, if we're going to include this. The issue we have to bump the size is pretty minor.

packages/v4/patternfly-docs/content/accessibility/checkbox/checkbox.md
```diff
@@ -3,17 +3,78 @@ id: Checkbox
...
 
-| React component| React prop | Which HTML element it appears on in markup | Explanation | 
-|---|---|---|---|
-| Checkbox | aria-label | .pf-c-check__input | Aria-label or accessible text for the checkbox. |
\ No newline at end of file
+- Include the `isRequired` prop if the checkbox must be checked
+
+For the HTML/CSS library:
+
+- Include the `required` attribute if the checkbox must be checked
```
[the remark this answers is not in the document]
nicolethoen: I think I agree that the required label seems unnecessary in the top section. I think it falls outside of the TLDR of a11y criteria wdyt?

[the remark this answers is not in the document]
thatblindgeye: Yeah I wasn't sure how often a single, required checkbox would be used but put it there in case. I can definitely remove it, though, and save it for the tables

packages/v4/patternfly-docs/content/accessibility/checkbox/checkbox.md
thatblindgeye: Yeah that's a good point. I thought I had seen some mention about a preferred size in terms of a checkbox so will see if I can find it (though it may have been an unrelated component).

If not providing a standard for the size, we could possibly replace the verbiage with something along the lines of ensuring a checkbox isn't placed too closely to other interactive elements (though we may also want to include a standard spacing)

thatblindgeye: @mcoker The PF site actually mentions [Success Criterion 2.5.5: Target Size](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html) in the [Patternfly's accessibility](https://www.patternfly.org/v4/accessibility/patternflys-accessibility) page. Since there's the asterisk of, "WCAG 2.5.5 is included for reference only. This guideline suggests a size that is larger than what PatternFly requires", we could just omit verbiage about the input itself having enough clickable space for now. 

An answer for [Checkbox size of 'clickable' area on desktop](https://ux.stackexchange.com/a/124093) has some interesting examples regarding target size from other sources.

packages/v4/patternfly-docs/content/accessibility/checkbox/checkbox.md
```diff
@@ -3,17 +3,63 @@ id: Checkbox
...
 
-| React component| React prop | Which HTML element it appears on in markup | Explanation | 
-|---|---|---|---|
-| Checkbox | aria-label | .pf-c-check__input | Aria-label or accessible text for the checkbox. |
\ No newline at end of file
+At a minimum, a checkbox should meet the following criteria:
+
+<List isPlain>
+  <ListItem>
+    <Checkbox id="checkbox-a11y-checkbox-1" label="Standard keyboard navigation can be used to navigate between checkboxes or other focusable elements." description={<span><kbd>Tab</kbd> navigates to the next checkbox or focusable element, and <kbd>Shift</kbd> + <kbd>Tab</kbd> navigates to the previous checkbox or focusable element. If a checkbox is inside of a menu, follow the [menu accessibility documentation](/components/menu/accessibility) instead.</span>} />
```
srambach, line 24: This link needs to be an <a href> rather than markdown.

srambach (approved):
Looks great! ⭐

nicolethoen (approved):

PR #3297, opened 2022-11-16
chore(Chip, ChipGroup): updated a11y docs

### LINKED ISSUE #3294: Update a11y docs for Chip and ChipGroup

thatblindgeye:
Closes #3294 

- The Core a11y section mentions placing `role="list"` on the `.pf-c-chip-group__list` element which is already a `<ul>`. It doesn't seem like that was necessary when testing with VO, but can be added back if there are still issues without it.
- Core examples will need to be updated as a followup to match the React markup, or vice versa. Currently React has `.pf-c-chip-group` with a `role="group"` so that an aria-label can be applied for groups without a visible name, and the category names have `aria-hidden="true"` which prevents the close button from using the category name for its accessible name.


nicolethoen:
Can you open follow up issues to track these discrepancies?

thatblindgeye:
Added back verbiage about the list role, as there is a bug still with Safari when a a list doesn't have any list markers and it ends up losing that semantic role. 

Shoutout to @mcoker for jogging my memory on this since it's been a long time since I've dealt with it:
https://web.dev/creative-list-styling/#styling-lists-that-dont-look-like-lists

packages/v4/patternfly-docs/content/accessibility/chip-group/chip-group.md
```diff
@@ -3,17 +3,58 @@ id: Chip group
...
-Bear in mind that the `aria-labelledby` attribute on the close button is based on the ID of the chip itself. So, if you are **changing the id of the chip**, you will need to update the `aria-labelledby` on the chip.
+To implement an accessible PatternFly **chip group** component:
 
-<br/>
-In general, the chip group component already has accessibility built in. However, if you’d like to customize it, edit the following:
+- Ensure any interactive elements - such as close buttons or overflow chips - within a chip group can be navigated to and interacted with via keyboard and other assistive technologies such as a screen reader
+- Ensure the entire chip group category name can receive focus, not just the chip group's close button, when the category name is truncated
+- Ensure any truncated text within a chip group category name is displayed as a tooltip when the category name is hovered via mouse or has focus
+- Provide an `aria-label` if the chip group does not have a visible text category name
+- Follow the [tooltip accessibility documentation](/components/tooltip/accessibility) for truncated content
```
nicolethoen, line 16: I think this note may be overkill here since we implement this on behalf of the consumer, right?
It's worth leaving a note about it in the testing section I think, but I like keeping these bullets at the top limited to the considerations the consumer is expected to make when implmeneting a Chip

packages/v4/patternfly-docs/content/accessibility/chip/chip.md
```diff
@@ -3,17 +3,58 @@ id: Chip
...
+## Accessibility
 
-**Screen reader users** should be able to navigate to the chip and have a screen reader read the chip’s text, as well as remove the chip from a selection if it has a button.
+To implement an accessible PatternFly **chip** component:
 
-The following attributes have been added for you or are customizable in PatternFly:
+- Ensure any close button within a chip can be navigated to and interacted with via keyboard and other assistive technologies such as a screen reader
+- Ensure the entire chip can receive focus, not just the chip's close button, when the chip content is truncated
+- Ensure any truncated text within a chip is displayed as a tooltip when the chip is hovered via mouse or has focus
+- Follow the [tooltip accessibility documentation](/components/tooltip/accessibility) for truncated chip content
```
nicolethoen, line 15: Same as note in chipgroup docs. 
It's worth leaving a note about the truncated chip being focusable in the testing section I think, but I like keeping these bullets at the top limited to the considerations the consumer is expected to make when implmeneting a Chip

packages/v4/patternfly-docs/content/accessibility/chip-group/chip-group.md
thatblindgeye: That's a good point. Do you think all 4 of these items could be removed from the first section? The aria-label one might have a case to be left in, but since there is a testing item for it it could be a bit redundant

nicolethoen: I think we can leave the aria-label note

nicolethoen (approved):

jessiehuff (approved):

PR #3327, opened 2022-12-21
fix(lodash): remove lodash alias

### LINKED ISSUE #3325: Remove global._ manipulation
Remove change introduced in #3324 once we determine the root cause of the topology/docs issue.

wise-king-sullyman:
Followup to #3324 as that fix does not seem to be working reliably.

Closes #3325


evwilkin (approved):
LGTM

dlabaj (approved):
LGTM

PR #3328, opened 2023-01-03
fixed version switcher latest version to point home

### LINKED ISSUE #2904: Version switcher - "latest" version link doesn't seem to work
If you switch to a previous release version, then switch back to the "latest" release, the link for "latest" takes you to whatever previous version you chose/are currently on instead of the latest release version.

evwilkin:
Closes #2904 

Having trouble testing due to older versions don't work locally or on Surge.  This PR updates the "latest" version to always point to `/`.


jenny-s51 (commented):
@evwilkin This looks like a reasonable workaround and the version switcher seems to function as expected now!

Though I'm wondering if this change will cause any link including `/v4` to no longer be functional link on the website, and how important it is that we keep it there to differentiate between v3, v4, v5. etc. 

FWIW, [Material UI prepends their domain names with any previous version number, which is something we could consider doing in order to keep those version numbers consumer-facing!](https://mui.com/versions/). What are your thoughts @evwilkin ?

evwilkin:
@jenny-s51 thanks for these questions!  This only affects the top "Latest" link, which currently lives at `patternfly.org/v4/` - this is redirected to from `patternfly.org` so this shouldn't change the visibility of the older version numbers in the URL (ex: `https://www.patternfly.org/2022.15/`) and should still work even if we end up changing the website from `/v4` to `/v5` or another redirect - this should point back to `patternfly.org` for the latest version which should always point to the latest version of the site.

We are looking at the possibility of having different sites for `v4` and `v5` altogether and changing the version switcher to link out to each site rather than individual versions, so this may change in the coming months but hopefully this update should fix the existing but for now!

Screenshot 2023-01-17 at 3 35 52 PM

jenny-s51 (approved):
Awesome, thanks for clarifying @evwilkin ! In that case this LGTM 🙂 A version switcher that links out to different sites sounds like a cool idea.

PR #3347, opened 2023-01-13
add action guidance

### LINKED ISSUE #3267: Add guidance around "Actions"
Add guidance to dropdown actions - around the different stylings and when to use which
Add guidance around the "Pattern" of where to place actions depending on what they affect

Doc: https://docs.google.com/document/d/1TA6uBuNIyA5F1X2rmz0pSUs0vkc7TYT1Rp-_fkQg-xo/edit

mmenestr:
closes #3267


mcarrano (commented):
This looks good @mmenestr . My only comment is that in the example at the end of the doc, it might be nice to show this with the split action button open to illustrate that there are additional actions hidden behind the toggle.

The other thing that stood out about this example is that it uses the toggle group like tabs which is something we said earlier we wanted to avoid.

mcarrano (approved):
Looks good to me!

Issue #3336, opened 2023-01-11
Alert - update a11y docs


nicolethoen:
Update the alert a11y docs to reflect the changes made in https://github.com/patternfly/patternfly-react/pull/8518

essentially, the Alert title might not need be a heading element if there is not description content below for it to be heading. A non-heading element like a span or a div could be used in that case. Also the titleHeadingLevel prop has been renamed to component prop.

PR #3348, opened 2023-01-16
fix(Alert): update alert a11y to address title elements

nicolethoen:
Closes: #3336 

Update the alert a11y docs to reflect the changes made in https://github.com/patternfly/patternfly-react/pull/8518


packages/v4/patternfly-docs/content/accessibility/alert/alert.md
```diff
@@ -19,21 +21,27 @@ At a minimum, an alert should meet the following criteria:
   <ListItem>
     <Checkbox id="alert-a11y-checkbox-1" label="If an alert will dynamically appear or update, it is rendered inside of an alert group component." />
   </ListItem>
+  <ListItem>
+    <Checkbox id="alert-a11y-checkbox-2" label="If an alert is using a heading level for its title, the heading level should fit the page heading level order so that heading levels are not skipped." />
```
thatblindgeye, line 25:
```suggestion
    <Checkbox id="alert-a11y-checkbox-2" label="If an alert is using a heading element for its title, the heading level fits the page heading level order so that heading levels are not skipped." />
```

packages/v4/patternfly-docs/content/accessibility/alert/alert.md
```diff
@@ -19,21 +21,27 @@ At a minimum, an alert should meet the following criteria:
...
+  </ListItem>
+  <ListItem>
+    <Checkbox id="alert-a11y-checkbox-2" label="If an alert's title is not describing any content in a description below it, the title should be wrapped in a non-heading element such as a 'span' or a 'div'." />
```
thatblindgeye, line 28:
```suggestion
    <Checkbox id="alert-a11y-checkbox-2" label="If an alert does not have any description below its title, the title is wrapped in a non-heading element such as a 'span' or a 'div'." />
```

I can't think of a good way to word one at the moment, but a description here quickly summarizing the "why" would be great. Best I can come up with is, "Without a description, a heading element in the title may cause other, unrelated content that follows seem related when navigating the page via rotor menus." Not a blocker if we don't get a description in, though.

thatblindgeye (changes requested):
Couple of verbiage tweaks below. Also it looks like the formatting of the file got tinkered with when saving.

thatblindgeye (commented):
Looks good! It'd be worth reverting the formatting in the "React customization" table just to keep it consistent

thatblindgeye (approved):

jessiehuff (approved):
Looks good! :)

PR #3363, opened 2023-01-18
added read only variations (documentation)

### LINKED ISSUE #3105: Add to form control guidelines about when to use which read-only variant
A new read only variant is going to be added. Want to add more specific guidance around when to use which read only variant, and when to use disabled vs. read only.

Original design request: https://github.com/patternfly/patternfly/issues/4935#issuecomment-1189607910
React implementation: https://github.com/patternfly/patternfly-react/issues/7930

**What to say:**
In form control guidelines, under "Read-only" section introduce both version of read only, the plain version and the boxed version.

**Guidance:**
use the plain version when you don't have the need to copy to clipboard
use the boxed version when you do have the need to copy to clipboard
Then mention that read-only differs from disabled state in that read-only will never be changeable, but disabled state implies that there is a way for it to become enabled.

Update the image to remove the disabled state and just have the plain read only, and copy to clipboard read only.

JustinXHale:
closes #3105


mmenestr (changes requested):
I would restructure this - I think we just need one header/section for disabled controls, and one for read only - right now it seems there are two for each.. I would also add a picture for the Plain and Boxed variations so it's clear what they are!

mmenestr (approved):

PR #3377, opened 2023-01-24
chore(Modal): updated a11y docs

### LINKED ISSUE #2500: Add guidelines to forms and modals about focusing on the first field
We don't currently have any guidance on when we should (or if we should) focus on the first field within a form or form within a modal. There are also circumstances where maybe we should be focusing on the cancel button in a destructive action (as [the guidelines in cloud.redhat.com say to do](https://docs.google.com/document/d/1Q58potYqoKfHSjzKY4yW-oUv_ESckI-vsbFnKp0cz3s/edit#)). I _think_ this type of documentation belongs at the PF level rather than leaving it up to products.

Add this guidance to the **modal guidelines**, about which buttons to focus on, and when.

### LINKED ISSUE #3346: Update a11y docs for Modal
Related: https://github.com/patternfly/patternfly-org/issues/2500

thatblindgeye:
Closes #2500, closes #3346 

This would require updates in React, possibly Core. For Core it'd probably just be updating/adding an example to show modal content overflowing, while React would need to add in functionality to set things under the hood for when modal content overflows. Would most likely be similar to the Accordion PRs that made similar updates (https://github.com/patternfly/patternfly/pull/5130 and https://github.com/patternfly/patternfly-react/pull/8515)

### DIFF
--- a/packages/v4/patternfly-docs/content/accessibility/modal/modal.md
+++ b/packages/v4/patternfly-docs/content/accessibility/modal/modal.md
@@ -3,28 +3,66 @@ id: Modal
 section: components
 ---
 
-A **modal** displays important information to a user without requiring them to navigate to a new page.
+import { Checkbox, List, ListItem } from '@patternfly/react-core';
 
-**Keyboard and mouse users** shouldn’t be able to interact with the rest of the page outside of an open modal. Set initial focus on the first focusable element in the modal. The user should be able to Tab to any interactive elements within the modal and use Tab + Shift to move backward through interactive elements. Keyboard users should be able to leave the modal by pressing Esc and focus should return to the element that invoked the modal. 
+## Accessibility
 
-**Screen reader users** shouldn’t be able to interact with the rest of the page outside an open modal. Add alternative text for any images or non-textual buttons (like icon buttons). Images should use the alt attribute: `aria-label` is most common to label icon buttons. Write a clear headline that describes the modal, and be sure to follow our modal content guidelines to clearly communicate information contained within it. 
+To implement an accessible PatternFly **modal** component:
 
-Bear in mind that any components placed inside the modal should also follow that component’s accessibility requirements.
-<br/>
-<br/>
+- Add an `aria-label` to the modal if it doesn't have a visible text title
+- Ensure any other PatternFly components follow their respective accessibility documentation
 
-**To make modal accessible:**
+## Testing
 
-- **If you don’t use a title in your modal**, add an `aria-label` to give the modal an accessible name. It will look like `aria-label=”[title of modal]” on the AboutModal and will appear directly on the class .pf-c-about-modal-box. 
-- Use `aria-describedby` for any description you add to the modal.
+At a minimum, a modal should meet the following criteria:
 
-<br/>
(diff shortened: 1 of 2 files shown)


srambach (approved):
👍 Looks ok pending the question about the link.
Also, I assume you meant to slip in the typo fixes for alert?

packages/v4/patternfly-docs/content/accessibility/modal/modal.md
```diff
@@ -3,28 +3,93 @@ id: Modal
...
 
-- **If you don’t use a title in your modal**, add an `aria-label` to give the modal an accessible name. It will look like `aria-label=”[title of modal]” on the AboutModal and will appear directly on the class .pf-c-about-modal-box. 
-- Use `aria-describedby` for any description you add to the modal.
+At a minimum, a modal should meet the following criteria:
 
-<br/>
-The following props can be added or are customizable in PatternFly:
+<List isPlain>
+  <ListItem>
+    <Checkbox id="modal-a11y-checkbox-1" label="Any content that is on the page, but outside of the modal, cannot be interacted with or navigated to while the modal is open." description={<span>Learn more about <a href="/accessibility/product-development-guide/#trapping-focus">trapping focus</a> in our product development guide for accessibility.</span>} />
```
srambach, line 22: This link doesn't work from the a11y page, and from the HTML page, the same link works ok but doesn't go to the anchor. Is this because it's the PR preview?

thatblindgeye: Hm, it might be because the "v4" in the URL isn't being taken into account here. There are probably other links in a11y docs that need to be updated to fix that then. I'll try pushing a fix for Modal here and do a followup for other components

srambach (dismissed):
It still doesn't jump to the anchor, but the link is correct. I assume that's a different issue since it exists on the core page as well. 
👍

packages/v4/patternfly-docs/content/accessibility/modal/modal.md
```diff
@@ -3,28 +3,94 @@ id: Modal
...
 
-- **If you don’t use a title in your modal**, add an `aria-label` to give the modal an accessible name. It will look like `aria-label=”[title of modal]” on the AboutModal and will appear directly on the class .pf-c-about-modal-box. 
-- Use `aria-describedby` for any description you add to the modal.
+At a minimum, a modal should meet the following criteria:
 
-<br/>
-The following props can be added or are customizable in PatternFly:
+<List isPlain>
+  <ListItem>
+    <Checkbox id="modal-a11y-checkbox-1" label="Any content that is on the page, but outside of the modal, cannot be interacted with or navigated to while the modal is open." description={<span>Learn more about <Link href="/accessibility/product-development-guide/#trapping-focus">trapping focus</Link> in our product development guide for accessibility.</span>} />
```
mmenestr, line 23: This is kind of picky, but I would link the words "product development guide" instead of trap focus since that's what you're actually linking to! Otherwise this looks great, and the guidelines are super clear!!

srambach (approved):
LGTM! 👍

mmenestr (approved):
LGTM!

PR #3396, opened 2023-02-03
chore(multiple): updated verbiage for arialabel props or attributes

### LINKED ISSUE #3395: Update aria-label description and use of anchors in a11y docs
Existing - multiple

**Describe the enhancement**
Update the description of `aria-label` in tables in the a11y docs to a more accurate description.

Additionally update usage of anchor elements to the Link component from the docs framework to avoid hardcoding in `v4` when linking to other pages.

thatblindgeye:
Closes #3395

For the `aria-label` prop/attribute, basically update the verbiage from "text describing..." to "text that labels..." to differentiate it from `aria-describedby` and more accurately define it. Due to a bug when using anchor elements in the "Testing" sections, updated them to the Link component from the doc framework so that we don't have to hardcode the PF version in the URL.

Also made a couple of tweaks to some table entries in a11y docs.


nicolethoen (approved):

nicolethoen:
@thatblindgeye if you rebase this, i'll merge it

PR #3407, opened 2023-02-10
feat(documentation-framework): Allow type alias definitions to be identified and rendered as prop tables

### LINKED ISSUE #3154: feat(documentation-framework): Allow type alias definitions to be identified and rendered as prop tables
Currently only `interface`s are parsed from files when accumulating propComponents which render PropTable components within the `Props` section of documentation pages. We should also be able to use `type` keywords and have those properties read and rendered in a similar fashion. 

Whether we leverage propComponents or a different markdown keyword is still to be determined. We might want to consider a different keyword called `types` and have those be rendered in their own section in our documentation.

**Example:**

This does not get displayed in our documentation under the `Props` section:
`type SomeProps = { name:  string; }`

This does:
`interface SomeProps { name: string; }`

jeffpuzzo:
**What**: Closes #3154

### DIFF
--- a/packages/documentation-framework/scripts/tsDocgen.js
+++ b/packages/documentation-framework/scripts/tsDocgen.js
@@ -1,28 +1,28 @@
-const fs = require('fs');
-const reactDocgen = require('react-docgen');
-const ts = require('typescript');
+const fs = require("fs");
+const reactDocgen = require("react-docgen");
+const ts = require("typescript");
 
 const annotations = [
   {
     regex: /@deprecated/,
-    name: 'deprecated',
-    type: 'Boolean'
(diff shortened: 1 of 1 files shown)


nicolethoen:
Is there a way I can test this, @jeffpuzzo?

jpuzz0:
> Is there a way I can test this, @jeffpuzzo?

I'm not sure of a way outside of checking out this branch, using the local build of this package in patternfly-react's `react-docs`, then spinning up the PF documentation there with a type alias for an example, where you'd include the type alias name for a given component in the `propComponents` array in that component's examples markdown file.

nicolethoen (approved):

PR #3415, opened 2023-02-20
chore: add v4 version switcher link

### LINKED ISSUE #3272: Version switcher - support PF4 link in v5
Per discussion in this week's Scrum of Scrums, we need to add a PF4 dropdown item to the version switcher, likely similar [how PF3 is currently displayed](https://github.com/patternfly/patternfly-org/blob/main/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js#L139-L147) as an option:
Screen Shot 2022-11-03 at 11 01 09 AM

**Any other information?**
Related: #2904 may be resolved following this work

kmcfaul:
Closes #3272

Adds a basic link to the version switcher. The link itself needs to be updated to point to wherever we decide to move `v4` (and if we choose `v4.patternfly.org` to match `v3`, then we also need to move the `core` site - maybe to `core.patternfly.org`?).

Also, do we want v4 to come before or after v3 in the list as v4 is more recent?

### DIFF
--- a/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
+++ b/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
@@ -151,6 +151,15 @@ const HeaderTools = ({
                     >
                       PatternFly 3
                     </DropdownItem>
+                    <DropdownItem
+                      key="PatternFly 4"
+                      className="ws-patternfly-3"
+                      target="_blank"
+                      href="#"
+                    >
+                      PatternFly 4
+                      <ExternalLinkAltIcon />
+                    </DropdownItem>
                   </DropdownList>
                 </DropdownGroup>
               </Dropdown>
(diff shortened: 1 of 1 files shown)


nicolethoen:
@kmcfaul here is the URL we will use for the v4 archive site
http://v4-archive.patternfly.org/v4/

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -151,6 +151,16 @@ const HeaderTools = ({
...
                     </DropdownItem>
+                    <DropdownItem
+                      key="PatternFly 4"
+                      className="ws-patternfly-3"
+                      isExternalLink
+                      to="http://v4-archive.patternfly.org/v4/"
+                      itemId="patternfly-4"
+                    >
+                      PatternFly 4
+                      <ExternalLinkAltIcon />
```
evwilkin, line 162: Looks like this icon is not longer imported & no longer needed with the `isExternalLink` prop that's used.

```suggestion

```

PR #3424, opened 2023-02-27
Added guidance around disabling menu actions

### LINKED ISSUE #2258: Provide guidance on presenting disabled menu items
Menu items in dropdown or select menus may currently be shown as disabled. However it is unclear how best to reflect the reason why an option is disabled. Some ideas include using descriptive text within the menu item or showing a tooltip or help popover. Would be good to provide designers with guidance about what options are available and when to use different approaches.

This would be added to the guidelines here: https://www.patternfly.org/v4/components/menu (will be released Nov 18th)

kaedward:
closes #2258


packages/v4/patternfly-docs/content/design-guidelines/components/menu/menu.md
```diff
@@ -97,6 +97,35 @@ When using red text for destructive actions, it is still recommended to require
 
 <img src="./img/red-delete-text.png" alt="drilldown menu" width="935"/>
 
+### Disabling menus and menu options
+Menus or menu items can be disabled for multiple reasons. The reason an action is disabled informs which design pattern you should use.
+
+1. **An action can be performed, but is not recommended.**
```
mmenestr, line 103: Thinking instead of having a numbered list, maybe they should just be mini headings -with the text underneath. Because right now it looks quite weird, with the example not being indented with the list: 
image

packages/v4/patternfly-docs/content/design-guidelines/components/menu/menu.md
```diff
@@ -97,6 +97,35 @@ When using red text for destructive actions, it is still recommended to require
...
+When an action can be performed but may result in an undesirable outcome, do not disable it. 
+
+Instead, add a [confirmation model](components/model). When the user clicks on the action, use a modal to explain the potential consequences and ask the user if they are sure they want to proceed.
+
+Example: A user wants to start maintenance on a master node.
```
mmenestr, line 108: Having this first item is confusing because you're talking about a case where something should NOT be disabled. I would probably make this last in the list, and also change the example to something more tangible like "Example: A user wants to delete a system."

packages/v4/patternfly-docs/content/design-guidelines/components/menu/menu.md
```diff
@@ -97,6 +97,35 @@ When using red text for destructive actions, it is still recommended to require
...
+
+2. **The user doesn’t have the correct permissions to perform an action.**
```
mmenestr, line 110: In this scenario, I would also advise adding a tooltip that says something like "You do not have the required permissions to perform this action. Contact your org admin" 

It would also help to have an image of this here.

mmenestr: I feel like this second bullet actually fits into the third - the unmet prerequisite is having the permissions to perform the action... so I would maybe combine them and just add this scenario as an example instead (?)

packages/v4/patternfly-docs/content/design-guidelines/components/menu/menu.md
```diff
@@ -97,6 +97,35 @@ When using red text for destructive actions, it is still recommended to require
...
+Example: A user can’t perform bulk actions until they select resources in the list.
+
+If an action is dependent on an unmet condition or status in the console, disable the action with a tooltip explaining why the action currently is unavailable.
+
+Example: A user cannot view past logs until their container is finished restarting.
+
+4. **An action cannot be performed due to a product constraint or rule.**
+If an action cannot be taken because of a product constraint or rule, hide the action. 
+
+Example: A user can’t delete a Red Hat template or remove an Environment variable from a managed resource.
```
mmenestr, line 127: This one is also confusing to me - why wouldn't a user be able to delete something, or remove something - wouldn't that also be permission related? What would be an example of a product constraint. I would make the examples more descriptive so that they're a bit more applicable/clear

packages/v4/patternfly-docs/content/design-guidelines/components/menu/menu.md
```diff
@@ -97,6 +97,36 @@ When using red text for destructive actions, it is still recommended to require
...
+### **An action is unavailable due to an unmet prerequisite, condition, permission, or status.** 
+If a user needs to complete a prerequisite to enable an action, disable the action and add a [tooltip](components/tooltip). The tooltip should explain what the user needs to do to enable the action. 
+
+<img src="./img/dropdown-disabled-action-tooltip.png" alt="disabled action tooltip" width="935"/>
+
+Example: A user can’t perform bulk actions until they select resources in the list.
+
+Example: A user with view-only privileges looks at a Pod’s details page and the options within the Actions dropdown are disabled.
+
+If an action is dependent on an unmet condition or status in the console, disable the action with a [tooltip](components/tooltip) explaining why the action currently is unavailable.
```
mmenestr, line 112: Feel like this sentence should be deleted/merge with the first sentence in this section because they're basically saying the same thing (?)

packages/v4/patternfly-docs/content/design-guidelines/components/menu/menu.md
```diff
@@ -97,6 +97,36 @@ When using red text for destructive actions, it is still recommended to require
...
+
+Example: A user cannot view past logs until their container is finished restarting.
+
+### **An action cannot be performed due to a product constraint or rule.**
+If an action cannot be taken because of a product constraint or rule, hide the action. 
+
+Example: While a user can delete a resource they own, they cannot delete a template or default resource.
```
mmenestr, line 119: This example makes sense, but the one below about permissions doesn't make sense in this section since you're already mentioning the permission use case in the first mini header? So I would just leave the first example and delete the second

packages/v4/patternfly-docs/content/design-guidelines/components/menu/menu.md
```diff
@@ -97,6 +97,36 @@ When using red text for destructive actions, it is still recommended to require
...
+Example: The user doesn’t have the correct permissions to perform an action.
+
+### **An action can be performed, but is not recommended.**
```
mmenestr, line 122: This feels weird here too.

I honestly wonder if we should have a "When to use disabled menu" "When not to use disabled menu" and separate it that way? Otherwise I'd just assume that every use case mentioned here should lead to using a disabled menu.

What do you think?

packages/v4/patternfly-docs/content/design-guidelines/components/menu/menu.md
```diff
@@ -84,7 +84,7 @@ Favoriting or other optional actions can be associated with any menu item. In th
 Any action that can be represented as an icon button can be placed in one or more menu items. In this case, the favoriting action is used to mark an item as a favorite.
 
 ### Drilldown menu
-<img src="./img/drill-down-menu.png" alt="drilldown menu" width="935"/>
+<img src="./img/drill-down-menu.png" alt="drilldown menu" width="385"/>
```
mmenestr, line 87: this image is way too small haha

PR #3431, opened 2023-03-02
chore: remove hard coded GA ID

### LINKED ISSUE #2944: Update analytics to GA4
Google analytics will need to be updated to GA4 before July 1, 2023:

- https://support.google.com/analytics/answer/11583528

evwilkin:
Closes #2944 

This PR removes the harded coded Google Analytics ID from the documentation framework, instead picking it up where it's passed through by the user as it does in other places.

Additionally, this PR confirms completion of GA4 setup which was completed through the analytics account and does not require additional code changes at this time (see issue for relevant page links).

### DIFF
--- a/packages/documentation-framework/app.js
+++ b/packages/documentation-framework/app.js
@@ -19,7 +19,7 @@ import './layouts/sideNavLayout/sideNavLayout.css';
 const AppRoute = ({ child, katacodaLayout, title, path }) => {
   const pathname = useLocation().pathname;
   if (typeof window !== 'undefined' && window.gtag) {
-    gtag('config', 'UA-47523816-6', {
+    gtag('config', process.env.googleAnalyticsID, {
       'page_path': pathname,
       'page_title': (title || pathname)
     });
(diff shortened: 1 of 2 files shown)


wise-king-sullyman (commented):
LGTM, though do we not need to add the environment variable somewhere? I see that we set a bunch of them in the base webpack config. Not super sure how all that works though.

evwilkin:
@wise-king-sullyman this ID is already passed in for the website [in patternfly-docs.config.js](https://github.com/patternfly/patternfly-org/blob/main/packages/v4/patternfly-docs/patternfly-docs.config.js#L6), and this PR picks that up & [adds the environment variable here](https://github.com/patternfly/patternfly-org/pull/3431/files#diff-d6c49dcd5290045cdf41c7b3f1e338d3063a6c9f73369503b3e84a02abc64792R140) - I think this is what you were thinking of but if not just let me know

wise-king-sullyman (approved):
Ah ok! That piece in the patternfly-docs.config file is what I was missing.

Issue #3432, opened 2023-03-02
Update the docs framework to Webpack 5


wise-king-sullyman:
During the Node 18 upgrade process an OpenSSL version related error was encountered. I implemented a work around in #3433, but that workaround still leaves us open to a security vulnerability that the breaking OpenSSL upgrade was meant to resolve, [as is discussed in this SO post](https://stackoverflow.com/a/73027407).

Because of this we should still upgrade to the most recently released version of Webpack to remove the previous workaround and resolve this potential security vulnerability.

wise-king-sullyman:
CC @evwilkin, I've updated this issue with new information that you should probably be aware of.

nicolethoen:
Closed by #3454

PR #3433, opened 2023-03-02
feat(docs-framework): add legacy ssl cli option

wise-king-sullyman:
Towards #3432

How to test:

0. Checkout branch
1. Set your node version to the latest LTS
2. Attempt to run `yarn build` to build the documentation, observe that it fails
3. `cd` to `packages/v4` and run `yarn build --legacySSL`, observe that it is now able to build

This shouldn't change how the package currently works, but will optionally enable use with node 18 (thus unblocking patternfly/patternfly-react#8303) until we are able to make the more substantial upgrade of webpack.


evwilkin (approved):
LGTM - worked as instructed 👍

PR #3443, opened 2023-03-13
Update to website title PatternFly 5

### LINKED ISSUE #3388: Update site title from 4 to 5
**Describe the enhancement**
The title of the site showing in tabs, etc currently says "PatternFly 4" and should be updated for PF5

**Any other information?**
https://github.com/patternfly/patternfly-org/blob/main/packages/documentation-framework/helpers/getTitle.js

mcarrano:
This PR will update the title text that displays on browser tabs from PatternFly 4 to PatternFly 5.

Closes #3388

### DIFF
--- a/packages/documentation-framework/helpers/getTitle.js
+++ b/packages/documentation-framework/helpers/getTitle.js
@@ -1,5 +1,5 @@
 module.exports = {
   getTitle(title) {
-    return `PatternFly 4${title ? ` • ${title}` : ''}`;
+    return `PatternFly 5${title ? ` • ${title}` : ''}`;
   }
 }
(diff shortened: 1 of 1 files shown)


evwilkin:
@mcarrano this looks good, but I'm wondering what you think about just removing the number altogether.  This seems like a change we'll have to repeat annually with each new major version number, and I'm not sure at what point the number loses a bit of significance at least as it's displayed in the site title.  For reference, here are the tabs of some other design systems:

Screenshot 2023-03-14 at 10 28 10 AM

evwilkin (approved):
LGTM - thank you!

PR #3451, opened 2023-03-22
chore: update to createRoot/hydrateRoot for React 18

kmcfaul:
Should still be backwards compatible with previous React versions.

Until webpack is also updated, using node >16 requires a workaround for building and starting the dev server.

### DIFF
--- a/packages/documentation-framework/app.js
+++ b/packages/documentation-framework/app.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import ReactDOM from 'react-dom';
+import { createRoot, hydrateRoot } from 'react-dom/client';
 import { Router, useLocation } from '@reach/router';
 import 'client-styles'; // Webpack replaces this import: patternfly-docs.css.js
 import { SideNavLayout } from '@patternfly/documentation-framework/layouts';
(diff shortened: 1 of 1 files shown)


kmcfaul:
Running into an issue with React 18's `hydrateRoot` and rehydration on the server-side rendering, where it's coming back with a mismatch and swapping to client-side rendering. I've tried adjusting the rendered content (fixing some DOM validation issues and even locally replacing the app with some dummy test html) but I'm still getting the mismatch so I think I'm missing something about what the server is expecting to be rendered. Has anyone run into an issue like this before?

@jschuler @dlabaj @wise-king-sullyman @dgutride @jonkoops

jonkoops:
I haven't ran into this as we don't have any applications that actually do SSR at the moment. But it is indeed strange that you are getting an issue here.

kmcfaul:
Closing as work wrapped up into #3454

kmcfaul:
For reference, we've found that `@reach/router` and Monaco editors had to be updated as they referenced older react versions, and our server template html had to have a couple newlines removed that were causing an issue with the SSR matching.

Issue #3450, opened 2023-03-20
Update sideNavLayout to use DropdownNext


thatblindgeye:
As part of https://github.com/patternfly/patternfly-react/issues/8072, the components used in the documentation framework's [sideNavLayout](https://github.com/patternfly/patternfly-org/blob/ee806688406004d5e6d947176c96ab159eb08f1b/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js) will need to be updated to use DropdownNext/the new default implementation of Dropdown after react-8072 is closed out.

thatblindgeye:
@nicolethoen this might actually be a candidate to use the Next Select instead

nicolethoen:
Closed by https://github.com/patternfly/patternfly-org/pull/3474

PR #3452, opened 2023-03-24
chore(Version Switcher): updated to Next Dropdown

thatblindgeye:
Draft for #3450 , should probably wait until Dropdown Next is promoted and then this can be updated again to point towards the correct package

Some updates need to be made in react-core:

- Allow focus to go to anchors (currently if the first item is an anchor, focus is not placed in the menu)
- Currently it does not seem like there's a way to make an anchor dropdown item open in a new tab/window

I also tested out adding visible text label above the Patternfly 3 link ("Previous versions"). That could possibly be confusing with the "Previous releases" label, and wouldn't be useful if we only plan to link to the most recent previous version (rather than PF3, PF4, etc).

### DIFF
--- a/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
+++ b/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.css
@@ -12,10 +12,6 @@
   padding: 0 var(--pf-global--spacer--md);
 }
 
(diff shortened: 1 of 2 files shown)


thatblindgeye:
Closing as https://github.com/patternfly/patternfly-org/pull/3474 resolved the issue

PR #3455, opened 2023-03-28
Added Test button guidelines to form validation

### LINKED ISSUE #3418: Add guidelines around test buttons to Form guidelines
Background information: https://github.com/patternfly/patternfly-design/issues/1151

andrew-ronaldson:
closes #3418


mmenestr:
A few comments I messaged you but want to write here so they don't get lost;
- I wonder if we should rename the header something else - Like “Validation with Test button” or “Test button validation”. Don't feel that strongly about that one but the current header reads weird to me, and this scenario a bit different to the other two types of validation. 
- Also wondering how this test button validation plays into the other forms of validation. I think we should add more detail around that. Essentially, want to convey what the value of it is that you don’t get with the other forms of validation, and whether it replaces the other forms vs goes together with them (ie: is the point to test before submitting, but you will still have error validation on submission, or is it either or?)

mmenestr (approved):

PR #3462, opened 2023-03-30
added old dark theme class

evwilkin:
TEST: do not merge.

### DIFF
--- a/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
+++ b/packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
@@ -87,7 +87,7 @@ const HeaderTools = ({
         >
           {hasDarkThemeSwitcher && (
             <ToolbarItem>
-              <Switch id="ws-theme-switch" label="Dark theme" defaultChecked={false} onChange={() =>
+              <Switch id="ws-theme-switch" label="Dark theme" defaultChecked={true} onChange={() =>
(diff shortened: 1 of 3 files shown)


evwilkin:
@lboehling @mcarrano @mceledonia FYI here is the prior dark theme slapped onto the website - https://patternfly-org-pr-3462-v4.surge.sh/v4

mcarrano:
Doing a quick audit of this. Most of the pages look OK with some exceptions. These are pages that would require rework if we were to do this:

Home page
Community page - images
Colors
Icons
Typography

Design guidelines mostly look OK with light images except for full screen images where it can be hard to differentiate image fom the backrgound.

Screenshot 2023-03-31 at 4 46 50 PM

Training pages

Issues on these pages are mostly images that don't display well on a dark background. Also some pages have text that will not pass contrast test with dark styles applied - see Colors and Typography pages. Realize that the are already issues open to replace the Home and Community pages so issues here might be mute, but we should ensure new designs are theme friendly. cc @lboehling 

@mmenestr would also be good for you to take a look. This came out of a conversation that took place about placing a dark theme toggle on the entire website.

mmenestr:
Another issue is that some images don't have backgrounds - so if they have black text it would not show up. I'm kind of worried about including the design guidelines in this. Would it be possible to change everything to dark theme EXCEPT the page background on design guidelines, or is that vey weird?

evwilkin:
Closed in favor of #3519

PR #3474, opened 2023-04-07
feat(tables): update tables, context selector

nicolethoen:


packages/documentation-framework/components/cssVariables/cssSearch.js
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { TextInput } from "@patternfly/react-core";
+import { SearchInput } from "@patternfly/react-core";
```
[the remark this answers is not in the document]
nicolethoen: it shouldn't. The search input will still exist in React. It already is built by simply using the TextInputGroup under the hood.

nicolethoen:
> * Should we open a followup issue to add back in the sorting functionality for the CSS table? Not really a blocker to include it in this PR, and also depends if we want to add the sorting back in

I could try to add it back in. when I was playing with it, it didn't feel like the sorting was adding very much. But i think I have a bit more time to take a second stab at it.

tlabaj:
> > * Should we open a followup issue to add back in the sorting functionality for the CSS table? Not really a blocker to include it in this PR, and also depends if we want to add the sorting back in
> 
> I could try to add it back in. when I was playing with it, it didn't feel like the sorting was adding very much. But i think I have a bit more time to take a second stab at it.

@thatblindgeye @nicolethoen Is it possible to tackle as a follow up since we need this to land to unblock development in React?

evwilkin (approved):
LGTM

packages/v4/patternfly-docs/content/design-guidelines/styles/icons/IconsTable.js
```diff
@@ -78,189 +64,137 @@ export class IconsTable extends React.Component {
...
-      const sortedRows = filteredRows.sort((a, b) => {
-        a = Array.isArray(a) ? a[0] : a;
-        b = Array.isArray(b) ? b[0] : b;
-        return a.cells[index].title < b.cells[index].title
+  const filteredRows = React.useMemo(() => {
+    return iconsData.filter(icon => {
+      return icon.Name.toLowerCase().includes(searchValue.toLowerCase()) ||
+        icon.Contextual_usage?.toLowerCase().includes(searchValue.toLowerCase()) ||
+        icon.Extra_context?.toLowerCase().includes(searchValue.toLowerCase()) ||
+        icon.Label?.toLowerCase().includes(searchValue.toLowerCase()) ||
```
evwilkin, line 77: Suggest including `React_name` in the search

```suggestion
        icon.Label?.toLowerCase().includes(searchValue.toLowerCase()) ||
        icon.React_name.toLowerCase().includes(searchValue.toLowerCase()) ||
```

packages/v4/patternfly-docs/content/design-guidelines/styles/icons/IconRecommendations.js
```diff
@@ -53,154 +51,142 @@ const iconSvgs = {
...
+                        </span>
+                        {newRec.iconName}
+                      </div>
+                    );
+                  } else return '';
+                })}
+              </Td>
+              <Td data-label="Contextual usage">
+                {recommendation.map(rec => {
+                  return rec.iconUsage
```
evwilkin, line 175: We should only display iconUsage for recommended icons (aka 'new' icons), and add block spacing to avoid results with multiple descriptions from running together (see `pficon-dragdrop` row in PR preview for 3 contextual usages running together)

```suggestion
                  if (rec.iconType === 'new') {
                    return (
                      <div key={`${rec.style}-${rec.iconName}-contextual-usage`}>
                        {rec.iconUsage}
                      </div>
                    )
                  }
```

packages/v4/patternfly-docs/content/design-guidelines/styles/icons/IconRecommendations.js
```diff
@@ -53,154 +51,142 @@ const iconSvgs = {
...
+      const newRec = recommendation[1];
+      return oldRec.iconName.toLowerCase().includes(searchValue.toLowerCase()) ||
+        oldRec.iconUsage.toLowerCase().includes(searchValue.toLowerCase()) ||
+        oldRec.type.toLowerCase().includes(searchValue.toLowerCase()) ||
+        newRec.iconName.toLowerCase().includes(searchValue.toLowerCase()) ||
+        newRec.iconUsage.toLowerCase().includes(searchValue.toLowerCase()) ||
+        newRec.type.toLowerCase().includes(searchValue.toLowerCase())
     });
-  }
+  }, [searchValue]);
```
evwilkin, line 83: This is hard coded to change 2 results (oldRec & newRec), but some rows have up to 5 results (4 old icons, 1 new icon) so suggest refactoring for a more dynamic approach to cover all use cases.  Also incorporates a separate recursive function to pull text out from `children` for the case of `Spinner` which otherwise was breaking upon search due to passing nested JSX rather than a searchable string to the `iconUsage` field:

```
const concatChildText = (children, text = '') => {
  React.Children.forEach(children, (child) => {
    if (typeof child === 'string') {
      text += child;
    } else if (React.isValidElement(child)) {
      text = concatChildText(child.props.children, text);
    }
  });
  return text;
}
```

```
const filteredRows = React.useMemo(() => {
    return recommendationsArray.filter(recommendation => (
      recommendation.some(rec => {
        return Object.values(rec).some(val => {
          const cellText = concatChildText(val);
          return cellText.toLowerCase().includes(searchValue.toLowerCase());
        });
      })
    ))
  }, [searchValue]);
```

thatblindgeye (approved):

PR #3484, opened 2023-04-11
chore(Label/LabelGroup): created combined a11y docs

### LINKED ISSUE #3460: Label and Label Group - merge a11y docs
Merge a11y docs for both components into a single, Label page

thatblindgeye:
Closes #3460 (also takes care of updating the Label a11y docs to the new template and creating LabelGroup docs)

[Label accessibility documentation](https://patternfly-org-pr-3484-v4.surge.sh/v4/components/label/accessibility)

### DIFF
--- a/packages/v4/patternfly-docs/content/accessibility/label/label.md
+++ b/packages/v4/patternfly-docs/content/accessibility/label/label.md
@@ -27,7 +27,7 @@ To implement an accessible PatternFly **label** and **label group**:
     <Checkbox id="label-a11y-checkbox-2" label="Standard keyboard interaction can be used to interact with the label group close button, label close button, or other interactive elements inside the chip group." description={<span><kbd>Enter</kbd> and <kbd>Space</kbd> should be able to activate the label group close button, individual label close buttons, or an overflow chip. This can usually be achieved by using "click" events.</span>} />
   </ListItem>
   <ListItem>
-    <Checkbox id="label-a11y-checkbox-3" label="If the label group's category name or an individual label's content is truncated, the entire category name or label can receive focus, not just the close button." description={<span>This can be achieved by ensuring the category name or label component itself has the <code className="ws-code">tabindex="0"</code> attribute. The category name/label should then be included in the standard keyboard navigation.</span>} />
+    <Checkbox id="label-a11y-checkbox-3" label="If the label group's category name or an individual label's content is truncated, the entire category name or label can receive focus, not just the close button." description={<span>This can be achieved by ensuring the category name or label component itself has the <code className="ws-code">tabindex="0"</code> attribute. The category name/label content should then be included in the standard keyboard navigation.</span>} />
   </ListItem>
   <ListItem>
     <Checkbox id="label-a11y-checkbox-4" label="If the label group's category name or an individual label's content is truncated, it has a tooltip that displays on hover or focus." description={<span>Follow the <Link href="/components/tooltip/accessibility">tooltip accessibility documentation</Link>.</span>} />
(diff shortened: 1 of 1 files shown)


packages/v4/patternfly-docs/content/accessibility/label/label.md
```diff
@@ -3,35 +3,113 @@ id: Label
...
+### Accessible label status and severity
+
+While labels can be styled using a variety of colors, relying on color alone to communicate information such as status or severity ("Loading" or "Danger") causes barriers to access. Users that are colorblind or have limited or no vision may not be able to perceive and understand a label's color, and users that rely on assistive technologies such as screen readers won't have those colors announced to them.
+
+When a label is intended or expected to convey information such as status or severity, you must ensure the label includes the following:
+
+- An appropriate color that matches the information being conveyed, if applicable. For example, "red" is commonly used to convey "danger" or "error".
+- An icon that matches the information being conveyed. For example, a lowercase "i" icon is commonly used to convey an informational status.
+- Visually hidden text that is accessible only to assistive technologies that conveys the status or severity. This can best be achieved by using the `pf-screen-reader` class on an element within the label content. For example:
+    ```
```
nicolethoen, line 96: I think we need one more space before this line.

nicolethoen:
I think this is great - I just had one additional thought while we are editing this page.
I have fielded so many requests for putting tooltips on labels. We might implement a variant one day that allows for this, but for now, could we call that out in the accessibility docs? Can we be explicit that labels are not interactable, and so if they have a tooltip attached, they need to be placed in a plain button and that button needs to be the trigger for the tooltip? Maybe reference the tooltip docs regarding rules for triggers?

thatblindgeye:
@nicolethoen Added a section at the end, let me know what you think. Regarding the example code I provided, it was between that and using our button with just plain variant. Here's the difference between the two:

Label with tooltip and two variations of a button

The plain variant has too much padding around it which would look odd when used with normal labels, but the link variant  + `is?Inline` has the text underline on hover which may make it seem like a link. I could also just provide a native button element with custom inline styling instead.

nicolethoen:
What you have might work
I would have formatted the code snippet more like this:
```
   <Tooltip content="Grey label that overflows and truncates">
      <Button isInline variant="plain">
        <Label textMaxWidth="16ch" isCompact>
          Grey label that overflows and truncates
        </Label>
      </Button>
    </Tooltip>
```
Is there a strong case for one over the other?

thatblindgeye:
It looks like that structure causes a double tooltip effect, with the tooltip being rendered both on the button and label. Even using the triggerRef prop rather than wrapping Tooltip around everything causes it. Even the structure I proposed isn't great either; doesn't look like by default you can get that truncation effect with the ellipsis, and even setting `textMaxWidth` on the label has some visual bugs.

We could suggest adding a `tabIndex={0}` prop to Label for now, then down the line introduce a new variant of label that takes tooltips into account (probably by utilizing a button as the wrapping element)?

nicolethoen:
hm - I guess I hadn't tried to use my structure on a label which truncates, so i hadn't noticed the double tooltip. Makes sense.

nicolethoen (approved):

nicolethoen (commented):
okay - I see.
We need to add tabIndex=0 for truncated labels. That makes sense.

But in the case on non-truncated labels, I think that using my code-snippet keeps the label without the link styles (the underline and blue text), so it might be preferred?

I'm thinking your code snippet should be talking about non-truncated labels.

thatblindgeye:
One thing we could do is update the "Additional considerations" section and show two code snippets: one for how you might achieve a tooltip on a truncated label (adding tabindex to the label) and one for a tooltip on a non-truncated label (using a Label, Button, and Tooltip).

Alternatively it looks like the `isOverflowLabel` prop might be usable as well since it makes the wrapping element a `button` already. The only issue is the Label code would need to be updated so that the tooltip ref is placed on that button instead of the label content. Might also require `isOverflowLabel` to be renamed to be less specific, and we'd still need to provide a snippet about tooltips on non-truncating labels here.

nicolethoen:
I still think the codesnippet in the docs is not the ideal solution:
As it is, it renders the label w/ button like this:
Screenshot 2023-05-04 at 11 58 07 AM

even if i update a few modifiers to make the label smaller to match this snippet:
```
<Label isCompact>
  <Tooltip content="Supplementary information about the tooltip">
    <Button size="sm" variant="plain">Label content</Button>
  </Tooltip>
</Label>
```
It looks like a non-compact label:


So can we not do this for non-truncated labels? 
```
<Tooltip content="Supplementary information about the tooltip">
    <Button size="sm" variant="plain">
      <Label >label with tooltip</Label>
    </Button>
</Tooltip>
```

thatblindgeye:
@nicolethoen That code snippet looks good now! Pushed an update to use the one you suggested

nicolethoen:
@thatblindgeye looks like it needs a rebase, then i'll merge it

nicolethoen (approved):

PR #3491, opened 2023-04-14
fix: clean up references to katacoda

nicolethoen:


packages/documentation-framework/app.js
```diff
@@ -31,26 +31,25 @@ const AppRoute = ({ child, katacodaLayout, title, path }) => {
...
-      {!katacodaLayout && process.env.hasFooter && <Footer />}
+      {process.env.hasFooter && <Footer />}
     </React.Fragment>
   );
 }
 
 const SideNavRouter = () => {
   const pathname = useLocation().pathname.replace(process.env.pathPrefix, '');
   const componentsData = process?.env?.componentsData;
-  const navOpen = !routes[pathname] || !routes[pathname].katacodaLayout;
```
mcoker, line 42: FYI I'm not sure if this is intentional, but I noticed the core workspace sidenav is no longer open by default (https://pf5.patternfly.org/, https://pf5.patternfly.org/components/alert/, etc) and looks to be a result of this change since the `.katacodaLayout` condition would evaluate as true on most pages.

Issue #3302, opened 2022-11-23
Accessibility fundamentals potential issues


mcoker:
Looking at https://www.patternfly.org/v4/accessibility/accessibility-fundamentals and I noticed some issues.

* "different user groups" link to https://www.a11yproject.com/posts/myth-accessibility-is-blind-people/ 404's
  * Looks like it changed to https://www.a11yproject.com/posts/accessibility-is-blind-people/
* > Users with poor motor control can use a range of devices to access contents. **
  * I don't see any sort of footnote for what the `**` may refer to
* > These needs mean that interfaces should not rely on color to communicate information
  * Maybe say "rely on color alone..."? We to use color as an indicator to communicate info, but in combination with some other visual indication
* > Often “the Persona Spectrum”
  * just looking around at how this is referred to, it looks like maybe this should be "Often the “Personal Spectrum“"
* > Don’t show interactive elements on hover. Interactive elements that can display in a pop-up must display on click/touch/Enter events.
  * I'm not sure if I have a specific recommendation, but the second part doesn't make much sense to me, but that could just be me!
* > There should be parity between hover and focus events.
  * This seems a little redundant with the first bullet point about parity between input types, but I might not see the difference.

The image:
1) There is no screen reader equivalent. The `alt` text just reads "Accessibility spectrum example". Assuming the image is meaningful, I think this goes against the experience parity principle. "There should be parity between the screen reader contents and visibly rendered contents"
2) The vertical text is hard to read
3) The text along the very bottom is so small, it's barely legible - looks like it's about 10px. I can't read it at all on my phone.
4) The links in the text at the bottom don't do anything since this is just a plain image
5) The native size is 6250x10629... and 5.3mb in size... there are all kinds of problems with that. At a minimum it should be resized to around 832px, which is the max size it will render currently. That said, if we ever widen the content area, the rendered size may change. If it hasn't already been, we should try and optimize its size. And we can deliver different images at different viewport sizes, and lazyload images, so they only load when they're brought into the viewport. I can help out here, just lemme know!
6) I wonder if it should come directly under the paragraph that talks about the personal spectrums represented in the image. Where it is now, it reads like it comes after the content for that section ends, under the last paragraph "There are plenty more examples we could include. Essentially, accessibility takes all users into account."

mcoker:
[...]

PR #3511, opened 2023-05-01
Fix accessibility fundamentals page issues

jenny-s51:
Towards #3302 

Fixed md related issues. Remaining issues are related to the content in the image - if we want to make those changes to the image, it looks like we may need to open a design issue

### DIFF
--- a/packages/v4/patternfly-docs/content/developer-resources/accessibility-guide.md
+++ b/packages/v4/patternfly-docs/content/developer-resources/accessibility-guide.md
@@ -14,57 +14,64 @@ When it comes to product experiences, no one should get left behind. That's why 
 
 Accessibility is best achieved when considered early in the design and development process. So if you contribute to PatternFly or use it for your products, familiarize yourself with accessibility first.
 
-
-
 ## Understanding users’ needs
-Great user experiences don’t just happen. They’re designed, tested, and refined with the user in mind. To develop inclusive products, it’s important to understand the varying needs of a wide range of users and consider the assistive tools and methods they use. 
 
-This section provides information to help you better understand and address the needs of these [different user groups](https://a11yproject.com/posts/myth-accessibility-is-blind-people/).
+Great user experiences don’t just happen. They’re designed, tested, and refined with the user in mind. To develop inclusive products, it’s important to understand the varying needs of a wide range of users and consider the assistive tools and methods they use.
+
+This section provides information to help you better understand and address the needs of these [different user groups](https://a11yproject.com/posts/accessibility-is-blind-people/).
 
 **Note**: Some users might fall into multiple groups, and some might use tools created for a different group.
 <br />
 
 ### Motor
-Users with poor motor control can use a range of devices to access contents. **
-- Users who rely on a keyboard need elements that are keyboard accessible and highly visible when in focus. 
+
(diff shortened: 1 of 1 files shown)


packages/v4/patternfly-docs/content/developer-resources/accessibility-guide.md
```diff
@@ -14,57 +14,64 @@ When it comes to product experiences, no one should get left behind. That's why
...
-- Alternative text (alt text) makes images accessible to users with low bandwidth connections or with older technologies that can’t load the images. 
-- Closed captioning benefits those in crowded areas, those teaching children how to read, or those learning a new language. 
+- Accessible written content is easier to comprehend for people who aren’t fluent in the language and for people with low literacy levels.
+- Alternative text (alt text) makes images accessible to users with low bandwidth connections or with older technologies that can’t load the images.
+- Closed captioning benefits those in crowded areas, those teaching children how to read, or those learning a new language.
 
 There are plenty more examples we could include. Essentially, accessibility takes _all users_ into account.
 
-<img src="./PF-a11y.png" alt="Accessibility spectrum example" />
+The image below shows that accessibility is a spectrum. Different kinds of cognitive and physical impairments impact how users interact with products. It is important to take all these differing disabilities into account when building accessible products.
```
nicolethoen, line 66: A `pf-v5-u-screen-reader` class on this paragraph might make more sense. I don't think it needs to be a visible paragraph. WDYT, @jessiehuff ?

packages/v4/patternfly-docs/content/developer-resources/accessibility-guide.md
```diff
@@ -14,57 +14,64 @@ When it comes to product experiences, no one should get left behind. That's why
...
+
+Users who have difficulty processing information benefit from well-written content.
+
+- Information should be clear, concise, and easy to scan.
 - Consider visual hierarchy, chunk content into short, related sections, and avoid long paragraphs.
 <br />
 <br />
 
-These are some of the main user groups, but it is not an exhaustive list. Often “the Persona Spectrum”, coined by Microsoft, can be used to understand related mismatches and motivations across a spectrum of permanent, temporary, and situational scenarios. 
+These are some of the main user groups, but it is not an exhaustive list. Often “the Persona Spectrum”, coined by Microsoft, can be used to understand related mismatches and motivations across a spectrum of permanent, temporary, and situational scenarios.
```
nicolethoen, line 56: I think, according to the original issue, this line should be changed from `Often "the Persona Spectrum"` to `Often the "Persona Spectrum"`

packages/v4/patternfly-docs/content/developer-resources/accessibility-guide.md
jenny-s51: Great idea @nicolethoen - added `pf-v5-u-screen-reader`

jessiehuff: I'm ok with either approach! When I added that paragraph, I think I was just trying to give context to that image in general, but I'm ok with making it only visible to screen readers if it doesn't make sense to include for visual users/is already obvious.

nicolethoen (approved):

PR #3512, opened 2023-05-01
Add images and summaries for All Components view

mcarrano:


evwilkin:
Thanks for working on this @mcarrano !

- In addition to the illustrations being added to the correct folder, they need to be added (or updated in the case of a name change) in the `index.js` file within that folder as this is what allows them to be shared to the other files that build out the component gallery:
  - https://github.com/patternfly/patternfly-org/blob/v5/packages/v4/patternfly-docs/images/component-illustrations/index.js
- Any items in the nav should show in the component gallery, whether or not they have a corresponding summary text/illustration in the `components-data` file - the gallery is generated the same way the sidenav is, so to remove a component from a gallery would mean removing it from the navigation as well.
- I do see "All components" in the preview link but not when I run the code locally, I'll be interested to see if this resolves itself next time updates are pushed to this issue.

If you need help updating that `index.js` file with the new image exports/updated image names just let me know!

evwilkin:
@mcarrano I was able to push these changes to your PR after testing them locally, so ideally when the preview link updates it should look as expected.  To recap the changes:
- Updated the [illustrations index file](https://github.com/patternfly/patternfly-org/blob/v5/packages/v4/patternfly-docs/images/component-illustrations/index.js):
  - Add exports for the new images you added
  - Remove exports for images no longer needed (ex: alert group)
  - Edit the names of exports for components that have been renamed (ex: file-upload-multiple to multiple-file-upload).
- Updated the illustration images to match changes above (delete unused illustrations, rename illustrations to match updated component names).

mcarrano:
@evwilkin Thanks for all this. The images now show up as I expect. I've updated these instructions to include the index.js update as that was the missing piece: https://github.com/patternfly/patternfly-org/wiki/Adding-new-components-to-the-website  Please review this and let me know if it all looks correct.

Just two remaining things:
* Still seeing "All Components" when I preview this.
* I think that this needs to be rebased against the latest core and React changes to reflect the correct set of components in the nav. Can you help me or tell me how to do that?

I am going to keep this as 'Draft' for now as there are some additional illustrations that we must have missed in the first round. I've requested these from design.

evwilkin:
@mcarrano updated - there were still design guideline markdown files for the removed components, which is why they were still listed in the gallery view (and in the nav).

mcarrano:
@evwilkin thanks for updating to remove the All Components from the gallery. Can you add back the design guideline files you removed? We have open design guideline issues to combine the text from alert/alert group, chip/chip group, and label/label group. @mmenestr will work on that so I don't want to loose the original content. We will definitely get those updates in before creating a final release candidate. Thanks.

evwilkin:
@mcarrano sorry about that - updated 👍

mcarrano:
@evwilkin @nicolethoen This should now be ready for review, except that there are some merge conflicts. Can you help to resolve those? Thanks!

PR #3519, opened 2023-05-08
feat: turn on dark switcher, remove extra css

### LINKED ISSUE #3372: Implement light/dark theme switcher
**Describe the enhancement**

For PF 5, need the ability to switch examples between light and dark theme in component preview pages (core and React).

Include screenshots or links to Marvel or other mockups.

Design issue: https://github.com/patternfly/patternfly-design/issues/1205
Design links: https://www.sketch.com/s/4abdb322-cd19-49af-8dbc-111adf6b071b

**Any other information?**

kmcfaul:
Closes #3372
Includes the dark theme switcher always (was previously dev-env only).
Removes some redundant topNav css that shouldn't be in use (commented out for now to double check preview, can fully remove if a css dev signs off that everything looks correct without it).

Open questions:
- In dark theme, the releases dropdown uses a secondary dark color (lighter than the masthead). Should this be overridden to match the masthead or left as default behavior?
- Primary buttons and inline links with updated grey text are not meeting contrast standard

Condensed issue list -
Core issues:
- Custom alert coloring
- Gold and cyan label coloring
- Menu toggle split-action missing dark theme color adjustment
- Progress stepper help action button needs to remove default browser background
- Text input group missing dark theme

React issues (follow up will be linked):
- Brand images don't reflect dark theme (requires investigation as svgs accounting for css var only works inline)
- Data list cell example uses custom styling not compatible with dark theme
- Table custom row wrapper also uses custom styling not compatible with dark theme


nicolethoen:
This PR would benefit from a rebase to fix a few of the wonky styles (esp on the home page)

nicolethoen (changes requested):
what this PR does is make it so that every consumer of the documentation-framework will have a dark theme switcher by default. That might be fine.

But the value of that environment variable for org is currently set in patternfly-org/packages/v5/patternfly-docs/patternfly-docs.config.js on line 16.

We should either remove that value, and keep the changes you made to the sideNavLayout, or we should just update that environment variable for true and keep sideNavLayout as it is.

kmcfaul:
@nicolethoen Updated to use the env variable so we have more control of it. It was located in the v4 folder, not v5, did I update the correct file? I'm guessing that folder name is part of the "update v4 references" issue.

nicolethoen:
I'm going to wait to merge this until the core/react work for dark theme is all merged and pulled into this PR. We can both keep an eye on the status of all that.
And at that point you should be able to remove the commented out CSS, as well.

mcoker:
@nicolethoen @kmcfaul the core PR is ready for review - https://github.com/patternfly/patternfly/pull/5535

Also the theme toggle switch on full page examples isn't applying the right class. I put a separate PR up for that so that we can merge that and I can pull it into the core PR ASAP so it's easier for folks to test those pages. Currently you have to manually edit the dark theme class on the `<html>` element to trigger dark theme. Here's that PR - https://github.com/patternfly/patternfly-org/pull/3532

kmcfaul:
Merged the fix for the dark theme CSS class and rebased to pull that in.

nicolethoen (commented):
Can we remove the commented out CSS now?

nicolethoen (approved):

evwilkin (approved):
LGTM

PR #3538, opened 2023-05-17
fixed codesandbox relative to absolute imports

### LINKED ISSUE #3525: Bug - full screen table demos don't open in codesandbox
We import our own data to populate rows and columns in full page Table demos. This data is not being imported into codesandboxes correctly. I think it needs an absolute import path rather than relative?

evwilkin:
Closes #3525 

This PR updates the conversion of relative import paths to absolute import paths, to be used in Codesandbox where the relative files aren't present.

Previously:
- `mdx-hast-to-jsx` would identify any relative imports in example code (`import Something from '../Something.js'`), parse out the name of the imported item, and build a map matching this relative imported item to the absolute import path for it
  - `{ Something:  '@patternfly/some-package/Something.js' }`
  - This is referenced in `codesandbox.js` to use simple string replacement to swap out the relative path with the absolute path in the example code before sending to Codesandbox example
  - Code had bugs and added extra complexity by parsing out imported item names from relative import path, when we already have relative & absolute paths which is all we really need

Now:
- `mdx-hast-to-jsx` maps the relative import path (not the name of the imported item) directly to the absolute import path
  - `{ '../../Something.js': '@patternfly/some-package/Something.js' }`
  - This also resolve the issue of destructuring relative imports including multiple items (ex: `import { rows, columns } from '../../Data.js'`) - _what_ is being imported doesn't really matter, we're just concerned with finding the relative file _where_ it's coming from and getting the absolute path to that file.


jenny-s51 (commented):
This is looking awesome @evwilkin!

Looked through a bunch of codesandbox demos that use relative import paths and lots of them have now been resolved with this update... Some icons within certain demos are still not rendering due to some residual image paths that should updated in react; opened https://github.com/patternfly/patternfly-react/issues/9140 to track these and assigned myself to these cleanup tasks

LGTM 👍

jenny-s51 (approved):

nicolethoen (approved):

PR #3559, opened 2023-05-27
Update numerics.md

anyuruf:
Swapped Euro n Pound signs


nicolethoen:
I think the labels were correct before? I don't think this is a change we can accept. 
@evwilkin can you sanity check me here?

anyuruf:
My bad!

evwilkin:
@anyuruf thanks for catching this - it does still need a fix, it's not quite correct currently and your update fixed part of it but not quite all the way.  Would you like to reopen this and update the PR?

Here's what I see needs changing:
Before
```
- EUR 1,500 or £1,500 (British pound)
- GPB 1,500 or €1,500 (Euro)
```
After
```
- GPB 1,500 or £1,500 (British pound)
- EUR 1,500 or €1,500 (Euro)
```

anyuruf:
If you could help. I don't think it matters who pulls all the matters is,
it gets done
🙏🙏🙏🙏🙏

PR #3570, opened 2023-06-02
enhancements & bug fixes to sectiongallery

### LINKED ISSUE #3568: Enhancements & bug fixes for sectionGallery
Add enhancements as initially tested in #3558 to allow more flexibility when using `sectionGallery` component within different navigation sections.

```[tasklist]
### Tasks
- [x] Only render `DataListCell` in list view if item has an illustration (rather than an empty cell that conditionally renders the image)
- [x] Add `initialLayout` prop allowing user to set the initial layout (grid or list)
- [x] Exclude gallery page from gallery results by checking location.pathname does not end with slug (rather than does not equal slug) - accounts for /v4 or similar prefix values
- [x] Handle case where no illustrations are provided
- [x] Handle subsections don't have title or id, need to provide fallback value
- [x] Toggle summary text within cards in grid view
```

evwilkin:
Closes #3568 

This PR:
- cleans up bugs that occurred when using `sectionGallery` without illustrations
- fixes bug with gallery page being listed in gallery due to pathPrefix interfering when looking for exact match to slug
- pulls the `SummaryComponent` out into a separate file renamed `TextSummary` and adds it to the grid view in addition to to the existing list view
- add new props to enable/disable both text & images in grid view and list view
- add new prop to set initial layout view (still defaults to grid)

### DIFF
--- a/packages/documentation-framework/components/sectionGallery/TextSummary.js
+++ b/packages/documentation-framework/components/sectionGallery/TextSummary.js
@@ -24,7 +24,7 @@ export const TextSummary = ({ id, itemsData }) => {
   return (
     <TextContent>
       <Text>
-        { id ? <SummaryComponent id={id} itemsData={itemsData} /> : null }
+        <SummaryComponent id={id} itemsData={itemsData} />
       </Text>
     </TextContent>
   )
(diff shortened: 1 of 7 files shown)


nicolethoen (approved):
Not quite sure how to test this, but logic seems good and I imagine we can debug as we need to as we build landing pages.

Issue #3565, opened 2023-05-31
V4 archive - add banner directing to current docs site


kmcfaul:
The v4 docs archive should link to our current documentation site.

PR #3572, opened 2023-06-05
chore: added banner to homepage with a link to new v5 page

ashley-o0o:
Added a banner to top of home page with a link to the new v5 url.

Closes https://github.com/patternfly/patternfly-org/issues/3565

### DIFF
--- a/packages/documentation-framework/templates/mdx.css
+++ b/packages/documentation-framework/templates/mdx.css
@@ -1,7 +1,7 @@
 #ws-page-content-router {
   display: flex;
   flex-direction: column;
-  height: 100%;
+  flex-shrink: 0;
 }
 
 .ws-release-notes-toc {
--- a/packages/v4/patternfly-docs/pages/home.css
+++ b/packages/v4/patternfly-docs/pages/home.css
@@ -223,34 +223,5 @@
 }
 
 .pf4-banner-link{
-  font-size: 18px;
-  margin-left: 8px;
-  color: #06c;
-}
-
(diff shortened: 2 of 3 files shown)


ashley-o0o:
Banner preview:
Screenshot 2023-06-05 at 3 28 36 PM

ashley-o0o:
**What:** Closes [#3565](https://github.com/patternfly/patternfly-org/issues/3565)

thatblindgeye (commented):
This looks nice! Had some comments/questions that aren't necessarily blockers, let me know what your thoughts:

- I wonder if the banner should either be sticky in its current position, or if it should be placed above the page content (similar to the [top and bottom Banner demo](https://pf5.patternfly.org/components/banner/html-demos/topbottom/), just without the bottom banner). Maybe a question more for design to chime in on cc @mcarrano
- Rather than defining new classes and adding onto the CSS for them, could we use PF components/layouts? `pf-c-banner` instead of `pf4-banner`, using Flex layouts, etc. We *are* already adding custom classes onto some PF components/native elements throughout, though, and we may need a new class or two to get the intended styling (the link styling inside a Banner may not be what we want here), so not too huge a deal


One possible issue I'm noticing, though, is when the viewport is shrunk down the spacing/placement of the banner content ends up like the following screenshot:

New banner at small viewport

nicolethoen (approved):

thatblindgeye (approved):
🎉 nice job on this!

PR #3603, opened 2023-06-15
Update favicon

### LINKED ISSUE #3487: Update the PatternFly favicon for the website to the new PF5 brand
The favicon that displays in browser tabs for the PattenFly website needs to be updated to the new PF 5 brand. I'm talking about the icon shown below.

Screenshot 2023-04-11 at 5 01 12 PM

@mceledonia can you provide the graphic? Not sure what the specs for this are.

evwilkin:
Closes #3487 

This PR updates the PF favicon to the new logo.

Note: We opted not to toggle dark/light favicons due to inconsistencies between browser tabs and how they react to preferred color schemes.


mcarrano:
Hmmm... This looks good in Chrome, but in Firefox it makes the favicon illegible as Firefox does not seem to change the color of the tab area based on OS theme.

Screenshot 2023-06-22 at 9 22 17 AM

evwilkin:
@mcarrano thanks for that callout - I see the same thing you mentioned in Firefox, though my tabs are black and stay black regardless of the light/dark theme of the rest of the browser.  Looking at the color of your tabs raises an interesting question, as this can be customized and some browsers update based on system theme (Chrome in my example) whereas others don't (Firefox in your example), and the user may have an otherwise customized colors scheme (for instance, your screenshot shows a blue background for non-active tabs that may be low contrast with the blue PF logo, but a white background for the active tab which would have low contrast with the white PF logo).

Based on this it may be easiest to just avoid toggling color altogether as it's inconsistent, and proceed with the blue logo which seems will have best contrast across standard dark/light backgrounds.  @lboehling @mceledonia does this make sense?

mcarrano:
I think so. Here's what the current icon looks like now without toggling it:

Tab deselected

Screenshot 2023-06-22 at 1 35 54 PM

Tab selected

Screenshot 2023-06-22 at 1 35 38 PM

The contrast is poor when the tab is selected but I think that's OK because as you say, it's hard to predict what color scheme a user might choose.

evwilkin:
@mcarrano  @mceledonia thanks for the feedback - updated to simply replace existing favicon with new logo.

mcarrano (approved):
Looks good. Thanks @evwilkin !

PR #3607, opened 2023-06-16
update font, background colors, replace hardcoded values

### LINKED ISSUE #3598: Bug - Some pages have unreadable text in dark mode
**Describe the problem**
There is some inline CSS that is causing text to be unreadable when the website is switched to dark mode.

Screenshot 2023-06-14 at 9 28 43 AM

Pages exhibiting this problem include:

https://patternfly-org-pr-3596-v4.surge.sh/v4/design-foundations/colors
https://patternfly-org-pr-3596-v4.surge.sh/v4/design-foundations/spacers
https://patternfly-org-pr-3596-v4.surge.sh/v4/charts/colors-for-charts

evwilkin:
Closes #3598 

This PR addresses bugs with dark theme text & background colors across Colors for charts, Spacers, and Colors pages.


evwilkin:
@mcoker - tagged you for review because I'm updating the CSS variable names displayed on the Colors for charts page to the new `pf-v5-` values, and also using those values for each color circle example.  Want to make sure that looks correct, and if anything around the charts colors needs to be adjusted for dark mode.

evwilkin:
@mcarrano thanks for those notes
- I've fixed the text color on the swatches on the Colors page
- the "Considering line height and padding" screenshot above - the dark text is actually part of an image along with the the white/green padding lines beside it.  I've added a white background behind the image to maintain the color contrast of the text in the image.

packages/documentation-site/patternfly-docs/content/design-guidelines/styles/colors-for-charts/colors-grid.js
```diff
@@ -12,7 +12,7 @@ const ColorsGrid = () => (
         <GridItem span={10}>
           <div className="ws-content-chart-colors-gallery--information">
             <div>#8BC1F7</div>
```
mcoker, line 14: This is going to change between light/dark theme. I'm not sure what a good fix is. I don't think we have tokens currently for dark theme values - either a dark theme value in the current/existing token, or potentially a separate token for dark theme - but if we did, this whole grid bit could probably just loop through the chart color tokens and spit out the var/hex color from the token, and include the dark theme hex color and hide it in light theme and show it in dark theme?

mcoker (approved):
LGTM. Left a comment about the dark theme color values not updating. Here are the dark theme color changes we should see https://github.com/patternfly/patternfly/blob/main/src/patternfly/base/themes/dark/_chart-globals.scss

That said, the colors aren't updating in dark theme as they should. Looks like a var name issue in core - https://github.com/patternfly/patternfly/pull/5673

mcarrano (approved):
Looks good. Thanks for the updates @evwilkin !

evwilkin:
@mcoker thanks for the feedback - I've updated this to import the Charts stylesheet, and to grab the CSS variable values directly from the root element and loop through them as you suggested.  I've also added a mutation observer that listens for the class to change on the root element & get the updated CSS var values, to update when the dark theme class is toggled.

packages/documentation-site/patternfly-docs/content/design-guidelines/styles/spacers/spacers.css
```diff
@@ -57,23 +57,23 @@
 }
 
 .ws-content-correct {
-  color: #486b00;
+  color: var(--pf-v5-global--info-color--100);
```
mcoker, line 60: Did you mean to turn this content blue? It was green before. 

```suggestion
  color: var(--pf-v5-global--success-color--100);
```

packages/documentation-site/patternfly-docs/content/design-guidelines/styles/colors-for-charts/colors-grid.js
```diff
@@ -1,409 +1,83 @@
 import React from 'react';
 import { Grid, GridItem } from '@patternfly/react-core';
+import '@patternfly/patternfly/patternfly-charts-theme-dark.css';
```
mcoker, line 3: Just a nit, but should we move the base charts CSS import into this file, too, just for organization? Or does it need to be in the other file? https://github.com/patternfly/patternfly-org/blob/d93dc163789d545df723003a5e5604b4582971e2/packages/documentation-site/patternfly-docs/content/design-guidelines/styles/colors-for-charts/colors-for-charts.md?plain=1#L14

```suggestion
import '@patternfly/patternfly/patternfly-charts.css';
import '@patternfly/patternfly/patternfly-charts-theme-dark.css';
```

packages/documentation-site/patternfly-docs/content/design-guidelines/styles/colors-for-charts/colors-grid.js
```diff
@@ -1,409 +1,83 @@
...
-    </GridItem>
+const colorFamilies = [
+  {headingText: 'Blue', varColor: 'blue'},
+  {headingText: 'Green', varColor: 'green'},
+  {headingText: 'Cyan', varColor: 'cyan'},
+  {headingText: 'Purple', varColor: 'purple'},
+  {headingText: 'Gold', varColor: 'gold'},
+  {headingText: 'Orange', varColor: 'orange'},
+  {headingText: 'Red', varColor: 'red'},
+  {headingText: 'Grey', varColor: 'black'}
```
mcoker, line 13: I noticed in the main design guidelines color page, we're using "gray", and using "grey" here. Though I pinged @mceledonia about using "black" here instead, since that's the color name in the code, and also what designers tend to give developers (for example, "black-300" and "green-300"), and he suggested updating the wording to reflect the actual color name - so maybe we can change this to "black"? cc @mceledonia @mmenestr can you confirm?

mcoker (dismissed):
LGTM! Left a few comments, the only one I'm really interested in was the change of the green to blue text for the "correct" block.

Also if we ever wanted to update the layout a bit to space the color/var/value pair a bit, I mocked that up with just the blue colors in a way that might work well. Everything is just using the flex/grid layout classes, except for removing the margin from `.ws-content-chart-colors-gallery--information` as that shouldn't be necessary.

Screenshot 2023-06-23 at 1 40 29 PM

mcarrano (approved):
LGTM. Thanks @evwilkin !

evwilkin:
@mcoker pulled in your suggestions, thanks!

PR #3611, opened 2023-06-16
fix(docs): use alpha stylesheet for Core codesandbox examples in dev

### LINKED ISSUE #3591: Core codesandbox examples pull in V4 code
Core examples on codesandbox (from V5) don't include styling as it is pulling in V4 styles:

[https://github.com/patternfly/patternfly-org/blob/v5/packages/documentation-framework/helpers/codesandbox.js#L26-L31](https://github.com/patternfly/patternfly-org/blob/v5/packages/documentation-framework/helpers/codesandbox.js#L26-L31)

It would be helpful to update for testing until V5 is live, and confirm if this will still work as expected once V5 is live.

evwilkin:
Closes #3591 

Checks for development mode and uses alpha release of Core stylesheet for codesandbox examples in dev mode - remains latest release for prod examples.

### DIFF
--- a/packages/documentation-framework/helpers/codesandbox.js
+++ b/packages/documentation-framework/helpers/codesandbox.js
@@ -3,7 +3,12 @@ const versions  = require('../versions.json');
 const overpass = require('./fonts');
 const { capitalize } = require('./capitalize');
 
+
 const getStaticParams = (title, html) => {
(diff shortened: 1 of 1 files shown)


evwilkin:
This is no longer needed - Codesandbox will pull in V5 stylesheet when V5 is officially released, resolving the original issue.

PR #3633, opened 2023-06-30
fix(documentation): remove excludes from codemod command

gitdallas:
these are all merged into pf now and don't need to be excluded

also `--fix` instead of `-fix`


edonehoo (approved):
makes sense & looks good!

thatblindgeye (approved):

nicolethoen (approved):

PR #3646, opened 2023-07-12
Added when to use Progress stepper

### LINKED ISSUE #3571: Add "When to use progress stepper vs wizard vs progress bar" guidance
It seems users are seeing the wizard and the progress stepper as interchangeable components.

Should clarify how they are different and what their intended uses were. I think that the wizard is more for step by step flows where users have to complete forms, where as a progress stepper was more to show a process loading in the background - in other words, its a more detailed progress bar. However, this is just my perspective on this, and it might be wrong.

Task:
- look into what the progress stepper was intended for
- Brush up progress stepper guidelines to be more specific and not sound like an interchangeable component to wizard
- add "when to use X vs Y guidance" in all these guidelines

wolfeallison:
closes #3571

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/design-guidelines/components/progress-stepper/progress-stepper.md
+++ b/packages/documentation-site/patternfly-docs/content/design-guidelines/components/progress-stepper/progress-stepper.md
@@ -36,9 +36,13 @@ Use a progress stepper to convey to a user how many steps are required to comple
 
 ### When to use progress stepper vs. wizard
 
-Use a progress stepper to indicate to the user where they are in a step-by-step linear process on a single page. The steps in the process could be completed by the user or happening in the background without the user completing the task themselves. 
+Use a progress stepper:
+* To indicate to the user where they are in a step-by-step linear process on a single page. 
+* When the steps in the process could be completed by the user or happening in the background without the user completing the task themselves. 
 
-Use a [wizard](/v4/components/wizard) to guide the user through a multi-step flow where they must complete specific tasks in order. Use a wizard when the steps are complex enough to need multiple pages. 
+Use a [wizard](/v4/components/wizard):
+* To guide the user through a multi-step flow where they must complete specific tasks in order. 
+* When the steps are complex enough to need to broken up into smaller, more manageable steps and does not keep user on the same screen.
 
 ## Variations
 There are a few variations of the progress stepper that can be used for different use cases. All variations can be displayed horizontally or vertically, and the text on each step can be left justified or centered.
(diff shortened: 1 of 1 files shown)


mmenestr:
@andrew-ronaldson do you mind giving your thoughts on this PR? It's to add guidance on when to use a wizard vs a progress stepper?

packages/documentation-site/patternfly-docs/content/design-guidelines/components/progress-stepper/progress-stepper.md
```diff
@@ -34,6 +34,12 @@ Use a progress stepper to convey to a user how many steps are required to comple
 
 <img src="./img/progress-stepper-table-view-popover.png" alt="Image showing example of progress stepper in a table view with a popover." width="1600" />
 
+### When to use progress stepper vs. wizard
+
+Use a progress stepper to indicate to the user where they are in a step-by-step linear process on a single page. The steps in the process could be completed by the user or happening in the background without the user completing the task themselves.
```
mmenestr, line 39: I would maybe try to break this up into bullets to match how we've done it in other guidelines -

Use a progress stepper:
- To indicate to the user where they are in a step-by-step linear process on a single page
- When the steps are happening in the background without the user completing the task themselves

Use a wizard:
- To guide a user through a multi-step flow
- When the steps are complex enough to need multiple pages 


I would also maybe add something about wizards taking the user out of context, and progress keeping them in the same place? Technically, the wizard keeps someone on the same page, and the progress stepper is still a multiple step flow except the content below changes instead of fully flipping to the same page. Just thinking out loud...

wolfeallison: Just changed the formatting! I tried to update the content a bit but I'm not sure I'm understanding correctly, so it might need another pass!

andrew-ronaldson (changes requested):
Thanks @wolfeallison just noticed this read a little off. 
> When the steps are complex enough to need to broken up into smaller, more manageable steps and does not keep user on the same screen.

Can we update to something like what's below:

When the steps are complex enough they need to be broken up into smaller, more manageable steps and does not keep user on the same screen.

packages/documentation-site/patternfly-docs/content/design-guidelines/components/progress-stepper/progress-stepper.md
```diff
@@ -34,6 +34,16 @@ Use a progress stepper to convey to a user how many steps are required to comple
...
 
+### When to use progress stepper vs. wizard
+
+Use a progress stepper:
+* To indicate to the user where they are in a step-by-step linear process on a single page. 
+* When the steps in the process could be completed by the user or happening in the background without the user completing the task themselves. 
+
+Use a [wizard](/v4/components/wizard):
+* To guide the user through a multi-step flow where they must complete specific tasks in order. 
+* When the steps are complex enough to need to broken up into smaller, more manageable steps and does not keep user on the same screen.
```
wolfeallison, line 45:
```suggestion
* When the steps are complex enough they need to be broken up into smaller, more manageable steps and does not keep user on the same screen.
```

andrew-ronaldson (approved):

mmenestr (approved):

PR #3678, opened 2023-08-01
docs(colors): updates colors guidelines as part of content audit.

### LINKED ISSUE #3652: Review and update colors guidelines

edonehoo:
Closes #3652


edonehoo:
Lots of open thoughts/questions: 
- What is necessary in the swatch info? For ex, dark 300 and 400 say the same thing, so is that helpful in distinguishing the difference? Is it helpful to included example components if we are only mentioning a few?
- Hex codes aren’t pulled in for primary colors - should that be added for consistency? 
- Hex codes need to change when you switch to dark theme? OR there will need to be a whole separate dark theme section 
- Hex codes and potentially colors aren't updating in the color family palettes when in dark theme 
- Some swatches are honestly just hard to see on the dark background

Longer term, I think the format of this page should be changed to better incorporate image examples and to more effectively present color info. Plus the styling doesn’t match other PF pages. I tried to mock up some ideas, but this would definitely be something to collaborate with a designer on. May be able to save this for Penta?

andrew-ronaldson (approved):
LGTM, thanks!

mmenestr (approved):

Issue #3668, opened 2023-07-31
Bug - Dark theme switcher not responsive


nicolethoen:
![Screenshot 2023-07-31 at 10 23 15 AM](https://github.com/patternfly/patternfly-org/assets/26305082/96978209-c0c7-49d3-836d-cfca2035d6d3)

Issue #3629, opened 2023-06-26
Bug - Website H2 top margins missing


evwilkin:
Example:  Margin above this H2 to match the margin above the spacer is missing:
- https://patternfly-org-pr-3607-site.surge.sh/charts/colors-for-charts/#the-system

Style rule targets `:not(:first-child)` and these `h2` tags now are wrapped in a div so they are all first children:
- https://github.com/patternfly/patternfly-org/blob/main/packages/documentation-framework/templates/mdx.css#L195-L196

PR #3690, opened 2023-08-04
fix: clean up h2 and dark theme switcher styles

nicolethoen:
Closes https://github.com/patternfly/patternfly-org/issues/3629
Closes https://github.com/patternfly/patternfly-org/issues/3668

The dark theme switcher is being fixed with CSS overrides - likely we will want to actually address these issues on the core side in general.

I've also asked design whether the switch component was the right choice. I'll update this with their response when I get it.


mcoker (approved):
LGTM 👍

PR #3721, opened 2023-08-24
fix(mdx): added missing key on CSSVariables

mcoker:
fixes https://github.com/patternfly/patternfly-org/issues/3720


evwilkin (approved):
LGTM

PR #3742, opened 2023-09-01
RTL workspace enhancements

### LINKED ISSUE #3738: RTL workspace enhancements
* Update the RTL switch to toggle `dir="rtl"` instead of `.ws-dir-rtl` https://github.com/patternfly/patternfly-org/blob/8f4544ac3197d77deaad244c694865aff0e8404f/packages/documentation-framework/components/example/example.js#L171
* Use [`.pf-v5-m-dir-ltr`](https://github.com/patternfly/patternfly/blob/b674c5131b2663587073cbc5b0707dbe475820ba/src/patternfly/base/_common.scss#L37) on the box that holds the dark theme/RTL switches instead of `ws-dir-ltr` https://github.com/patternfly/patternfly-org/blob/8f4544ac3197d77deaad244c694865aff0e8404f/packages/documentation-framework/components/example/example.js#L164
* Remove `.ws-dir-[ltr/rtl]` https://github.com/patternfly/patternfly-org/blob/8f4544ac3197d77deaad244c694865aff0e8404f/packages/documentation-framework/components/example/example.css#L134-L140
* Nice to have - a toggle in the masthead on the main workspace view (like the dark theme switcher) that can toggle `.pf-v5-m-dir-rtl` on the example blocks, which would let us quickly scan RTL examples in the example list page. That should be an easy add. It could eventually toggle the whole page, too, not the example blocks, if we wanted to test the workspace this way.
  * We could probably just add a plain action in the masthead that toggles between the FA [align-left](https://fontawesome.com/icons/align-left?f=classic&s=solid) and [align-right](https://fontawesome.com/icons/align-right?f=classic&s=solid) icons?

evwilkin:
Closes #3738 

This PR:
1. Updates full page examples as follows:
    - Updates the existing full-page example RTL switch to toggle `dir="rtl"` instead of `.ws-dir-rtl` and removes the `.ws-dir-[rtl/ltr]` css styles.
    - Replaces `.ws-dir-ltr` with `.pf-v5-m-dir-ltr` on the flex container for the theme/RTL switches.
2. Updates the workspace masthead with a switch to toggle `.pf-v5-m-dir-rtl` on all instances of `.ws-example .ws-preview` on the page.
    - To ensure examples remained synced to the switch when navigating across pages, this state is set in `SideNavLayout` where the switch lives and then passed down through context to the `Example` component.
    - NOTE: this RTL switch is currently enabled by default for PR preview, ideally once review is complete we can update set `hasRTLSwitcher` to `false` and only use it in the workspaces, or disable it in prod by checking `isProd` and using reverse logic as the [Google Analytics ID.](https://github.com/patternfly/patternfly-org/blob/main/packages/documentation-framework/scripts/webpack/webpack.base.config.js#L140)


wise-king-sullyman (approved):

packages/documentation-site/patternfly-docs/patternfly-docs.config.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
   hasVersionSwitcher: true,
   hasDesignGuidelines: true,
   hasDarkThemeSwitcher: true,
+  hasRTLSwitcher: true,
```
mcoker, line 16: just a reminder to take this out prior to merge

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -87,6 +93,27 @@ const HeaderTools = ({
...
+                  buttonId="ws-rtl-toggle"
+                  isSelected={false}
+                  onChange={() => setIsRtl(isRtl => {
+                    const examples = document.querySelectorAll('.ws-example');
+                    examples.forEach(ex => ex.classList.toggle('pf-v5-m-dir-rtl'));
+                    return !isRtl;
+                  })}
+                />
+              </ToggleGroup>
+              <Switch id="ws-rtl-switch" label="Example direction" defaultChecked={false} onChange={() =>{
```
mcoker, line 111: I like the switch. to save some real estate we could switch the toggle between "RTL" and "LTR" text.

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -87,6 +93,27 @@ const HeaderTools = ({
...
+                const examples = document.querySelectorAll('.ws-example');
+                examples.forEach(ex => ex.classList.toggle('pf-v5-m-dir-rtl'));
```
mcoker, line 113: Not a big deal, but I don't think there's value in switching the example header (`.ws-example-header`). I _think_ we could target `.ws-preview` (or `.ws-example .ws-preview` if we need to scope it?)? If not, we could add the `.pf-v5-m-dir-ltr` class to the example header, which will keep it from flipping.

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -87,6 +93,27 @@ const HeaderTools = ({
...
+            <ToolbarItem>
+              <ToggleGroup aria-label="Example direction toggle">
+                <ToggleGroupItem
+                  icon={isRtl ? <AlignRightIcon /> : <AlignLeftIcon />}
+                  aria-label="Toggle direction"
+                  buttonId="ws-rtl-toggle"
+                  isSelected={false}
+                  onChange={() => setIsRtl(isRtl => {
+                    const examples = document.querySelectorAll('.ws-example');
+                    examples.forEach(ex => ex.classList.toggle('pf-v5-m-dir-rtl'));
```
mcoker, line 106: also not a big deal but when these vars are avialble in react tokens, you can use the token instead so you don't have to hardcode it (and update the `v5` for v6)

mcoker (dismissed):

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
evwilkin: Great callout! ➕➕

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -27,6 +27,8 @@ import {
 } from '@patternfly/react-core';
 import BarsIcon from '@patternfly/react-icons/dist/esm/icons/bars-icon';
 import GithubIcon from '@patternfly/react-icons/dist/esm/icons/github-icon';
+import AlignLeftIcon from '@patternfly/react-icons/dist/esm/icons/align-left-icon';
+import AlignRightIcon from '@patternfly/react-icons/dist/esm/icons/align-right-icon';
```
mcoker, line 31: I think you can remove these now.

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -87,6 +91,15 @@ const HeaderTools = ({
                 document.querySelector('html').classList.toggle('pf-v5-theme-dark')} />
             </ToolbarItem>
           )}
+          {hasRTLSwitcher && (
+            <ToolbarItem>
+              <Switch id="ws-rtl-switch" label={isRtl ? 'RTL' : 'LTR'} defaultChecked={false} onChange={() =>{
```
mcoker, line 96: Sorry, I know I said we could switch the label, but to be consistent with the dark theme switch, and the dark/RTL switch on full page examples, I think it looks better with a static "RTL" switch label.

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -87,6 +91,15 @@ const HeaderTools = ({
...
+                const examplePreviews = document.querySelectorAll('.ws-example .ws-preview');
+                examplePreviews.forEach(ex => ex.classList.toggle('pf-v5-m-dir-rtl'));
```
mcoker, line 98: Since this fires from the switch per page, it's easy for the switch's checked state to get out of sync with the examples. For example - go to the home page or some component page and switch RTL on, then go to another component page and the switch is checked but the examples are LTR. What if the switch just toggled some bool (`rtlEnabled` or something) the example example block had a conditional in it spit out the rtl class if `rtlEnabled`?

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
evwilkin: Thanks - pulled out 3 other unused image imports along with these ✅

mcoker (approved):
LGTM!!

packages/documentation-framework/components/example/example.js
```diff
@@ -161,14 +163,17 @@ export const Example = ({
       <div id={previewId} className={css(className, 'pf-v5-u-h-100')}>
         {livePreview}
         {(hasDarkThemeSwitcher || hasRTLSwitcher) && (
-          <Flex direction={{ default: 'column' }} gap={{ default: 'gapLg' }} className="ws-full-page-utils ws-dir-ltr">
+          <Flex direction={{ default: 'column' }} gap={{ default: 'gapLg' }} className="ws-full-page-utils pf-v5-m-dir-ltr ">
```
nicolethoen, line 166: @mcoker is this the only versioned modifier class name?

mcoker: @nicolethoen there are a few others, you can see them here - https://github.com/patternfly/patternfly/blob/825d0af0912528722280460f17c82acf5b66c241/src/patternfly/base/_common.scss

They're global classes that can be applied to anything. Since they aren't paired with a versioned component/layout block or element class, they need to be versioned.

nicolethoen (approved):

packages/documentation-site/patternfly-docs/patternfly-docs.config.js
evwilkin: Done - thanks!

mcoker (approved):
This is great!!! Thanks for jumping on this issue 👍👍

PR #3747, opened 2023-09-07
feat(extensions): Added extensions only build to patternfly org.

dlabaj:
Added the ability to perform an extension only build of the patternfly.org site.  This allows use to release builds to a staging site for extension development. 

Normal site build:
Screenshot 2023-09-07 at 3 26 31 PM

Extension site build:
Screenshot 2023-09-07 at 3 47 53 PM


package.json
```diff
@@ -8,17 +8,20 @@
     "analyze": "yarn workspace patternfly-org analyze",
     "develop": "yarn develop:v5",
     "develop:v5": "yarn workspace patternfly-org develop",
+    "develop:extensions": "EXTENSIONS_ONLY=true PRERELEASE=true yarn develop:v5",
     "build": "yarn build:v5",
     "build:analyze": "yarn workspace patternfly-org build:analyze && yarn copy:v5",
     "build:v4": "yarn build:v5",
     "build:v5": "yarn workspace patternfly-org build && yarn copy:v5",
+    "build:extensions": "EXTENSIONS_ONLY=true PRERELEASE=true yarn build:v5",
```
dlabaj, line 16: EXTENSIONS_ONLY env variable indicates that this is an extensions build.  PRERELEASE env variable indicates that you would like to include sections / docs that are not ready for release yet in this build (e.g. Component Groups section).

packages/documentation-site/patternfly-docs/patternfly-docs.config.js
```diff
@@ -1,34 +1,56 @@
...
-    { section: 'developer-resources' },
-    { section: 'training' }, 
-  ],
-  topNavItems: [],
-  port: 8003,
+// This module is shared between NodeJS and babelled ES5, if this extension only build
+// exclude the other side nav items.
+if (process.env.EXTENSIONS_ONLY === 'true') { 
+    module.exports = {
+      googleAnalyticsID: 'UA-47523816-6',
```
nicolethoen, line 6: I'm not sure if we want to use the same analytics ID and algolia apiKey for the staging site. It'll be using all new URLs as well, so the search won't work. I wonder if you can take out a lot of these extra fields so that it more closely matches the one in core: https://github.com/patternfly/patternfly/blob/main/patternfly-docs/patternfly-docs.config.js

evwilkin: Agreed here - the search returns results that would be hosted on an extensions-only site, most of the links in the footer aren't relevant (Components, Get started, etc pages won't exist on this site), and I'm not sure that the version switcher is relevant so we can likely strip this back further to keep the focus on the extensions.

packages/documentation-site/patternfly-docs/patternfly-docs.source.js
```diff
@@ -128,4 +114,43 @@ module.exports = (sourceMD, sourceProps, sourceFunctionDocs) => {
 
   sourceProps(path.join(consoleSrcPath, "/**/*.tsx"), reactPropsIgnore);
   sourceMD(path.join(consoleDocsPath, "/examples/*.md"), 'react');
+
+  // Component Groups extension (Currently in PRERELEASE, so only include in PRERELEASE builds)
```
evwilkin, line 118: NIT - this `if` case for the `PRERELEASE` flag is nested in the middle of a bunch of other extensions.  For organization (as this section could grow in the future), can we move this condition to the top or the bottom to group the code?

Something like

1. [always included] at the top of the source file, 
2. [if !EXTENSIONS_ONLY] in the middle, and 
3. [PRERELEASE] at the bottom?

evwilkin (changes requested):
2nd Nicole's comment about stripping out some more content from the `patternfly-docs.config.js` file, and a couple thoughts to keep the code organized as we start throwing conditions throughout the source file - open to your feedback on these ideas.

packages/documentation-site/patternfly-docs/patternfly-docs.source.js
```diff
@@ -6,102 +6,88 @@ module.exports = (sourceMD, sourceProps, sourceFunctionDocs) => {
...
-
-  // React-user-feedback MD
-  sourceMD(path.join(reactUserFeedbackPath, '/**/examples/*.md'), 'react');
-  sourceMD(path.join(reactUserFeedbackPath, '/**/design-guidelines/*.md'), 'design-guidelines');
-
-  // React OUIA MD
-  sourceMD(path.join(reactCorePath, '/**/helpers/OUIA/*.md'), 'react');
+  
+  if (process.env.EXTENSIONS_ONLY === 'true') { 
+    sourceMD(path.join(contentBase, 'extensions/**/*.md'), 'extensions');
```
evwilkin, line 12: This line is called whether or not the build is extensions-only or regular, as both builds include the extension docs, so it's repeated below on [line 20](https://github.com/patternfly/patternfly-org/pull/3747/files#diff-4f2188cb353f2766c010a081762bb64556eb7bf758649c76cb09e4b79c29beb8R20).

To clean up a bit can we pull this extensions sourcing outside of the `if/else` statement, and then flip the logic to only run the rest of the `sourceMD` lines `if(!(process.env.EXTENSIONS_ONLY === 'true'))`?  To me that makes it a little clearer that the extension files are always sourced, and then the rest are sourced only if extensions_only isn't true.

evwilkin:
FYI - unrelated to this PR as far as I know, but the Component groups => ErrorBoundary page crashes.

evwilkin (approved):
LGTM 🎉

nicolethoen (approved):

PR #3767, opened 2023-09-27
docs(rtl-handbook): Update documentation with more details.

### LINKED ISSUE #3758: RTL handbook - add any react specific documentation, other follow ups
Looks like we should probably add something about

* Chart legends - https://github.com/patternfly/patternfly-react/pull/9570#issuecomment-1725916867
* `<Icon shouldMirrorRTL />` - https://github.com/patternfly/patternfly-react/blob/ff467eef5db530f0245ba29ca01db1f4a239f274/packages/react-core/src/components/Icon/Icon.tsx#L27-L30
* Anything about things that use popper?
* Link the mention of logical properties in the handbook to a docs page on what logical properties are (https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values)
* Anything else?

cc @srambach @tlabaj @wise-king-sullyman @thatblindgeye @kmcfaul

edonehoo:
Closes #3758


packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
```diff
@@ -11,16 +11,29 @@ The default browser behavior displays content following LTR formatting. In order
...
+
+**Note:** When using self-defined custom legends and label components, RTL support must be handled manually as `legendDirection` does not apply to custom implementations.
+
+### Icons
+PatternFly components have also been written to reverse directional icons that are part of the component. For example, the caret indicating expansion of various components is properly reversed. 
+
+Other icons you may have in your interface will not be mirrored automatically, as it may be appropriate to mirror some but not others. Be sure to review all icons in your product and adjust appropriately. PatternFly has modifier classes listed in the following section that can help with that.
 
-Other icons you may have in your interface will not be mirrored automatically, as it may be appropriate to mirror some but not others. Be sure to review all icons in your product and adjust appropriately. PatternFly has modifier classes listed below that can help with that.
 <br />
```
edonehoo, line 32:
```suggestion
```

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
```diff
@@ -11,25 +11,32 @@ The default browser behavior displays content following LTR formatting. In order
 
 ## Support in PatternFly
 
-PatternFly components have been written to use logical properties - `block-start`, `inline-end`, `block-end`, and `inline-start` - rather than `top`, `right`, `bottom`, and `left`. This allows styles like margin and padding to be correctly laid out in both LTR and RTL. However, to avoid a sweeping breaking change, PatternFly variable names still use `top`, `bottom`, `left`, and `right`.
+PatternFly components have been written to use [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values) - `block-start`, `inline-end`, `block-end`, and `inline-start` - rather than `top`, `right`, `bottom`, and `left`. This allows styles like margin and padding to be correctly laid out in both LTR and RTL. However, to avoid a sweeping breaking change, PatternFly variable names still use `top`, `bottom`, `left`, and `right`.
```
edonehoo, line 14:
```suggestion
### Components 

PatternFly components have been written to use [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values) - `block-start`, `inline-end`, `block-end`, and `inline-start` - rather than `top`, `right`, `bottom`, and `left`. This allows styles like margin and padding to be correctly laid out in both LTR and RTL. However, to avoid a sweeping breaking change, PatternFly variable names still use `top`, `bottom`, `left`, and `right`.
```

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
```diff
@@ -11,25 +11,34 @@ The default browser behavior displays content following LTR formatting. In order
...
+
+Additionally, when in RTL, ensure that the `<ChartAxis>` representing a chart's x-axis uses the `invertAxis` property to flip the axis.
+
+**Note:** When using self-defined custom legends and label components, RTL support must be handled manually as `legendDirection` does not apply to custom implementations.
+
+### Icons
+PatternFly components have also been written to reverse directional icons that are part of the component. For example, the caret indicating expansion of various components is properly reversed. 
+
+Care should be taken to handle icons appropriately for RTL. Some icons you may have in your interface will not be mirrored automatically. [Sometimes an icon should be reversed in RTL](https://m2.material.io/design/usability/bidirectionality.html#mirroring-elements) (for example, when indicating a direction such as an expansion caret), and sometimes it should not be reversed (such as a logo, or an icon indicating time). When an icon *should* be reversed, but isn't reverse automatically, you can use the modifier `.pf-v5-m-mirror-inline-rtl`.
```
edonehoo, line 32:
```suggestion
Care should be taken to handle icons appropriately for RTL. Some icons you may have in your interface will not be mirrored automatically. [Sometimes an icon should be reversed in RTL](https://m2.material.io/design/usability/bidirectionality.html#mirroring-elements) (for example, when indicating a direction such as an expansion caret), and sometimes it should not be reversed (such as a logo, or an icon indicating time). When an icon *should* be reversed, but isn't reversed automatically, you can use the modifier `.pf-v5-m-mirror-inline-rtl`.
```

edonehoo:
@mcoker @srambach I added updates to the RTL handbook and also updated the structure. Can you lmk what you think?

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
```diff
@@ -11,25 +11,34 @@ The default browser behavior displays content following LTR formatting. In order
...
+### Components 
 
-PatternFly components have been written to reverse directional icons that are part of the component - for example, the caret indicating expansion of various components is properly reversed. 
+PatternFly components have been written to use [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values) - `block-start`, `inline-end`, `block-end`, and `inline-start` - rather than `top`, `right`, `bottom`, and `left`. This allows styles like margin and padding to be correctly laid out in both LTR and RTL. However, to avoid a sweeping breaking change, PatternFly variable names still use `top`, `bottom`, `left`, and `right`.
 
-Other icons you may have in your interface will not be mirrored automatically, as it may be appropriate to mirror some but not others. Be sure to review all icons in your product and adjust appropriately. PatternFly has modifier classes listed below that can help with that.
-<br />
+**Note:** An exception to this support is with any components using popper.js, like tooltips and popovers.
```
mcoker, line 18: Do we want to mention anything about our upgrade plan to floating-ui, or how to manage popper things in RTL for now? For example, something simple like "For now, the popper position/placement will need to be managed manually when used on pages with RTL content." cc @tlabaj @wise-king-sullyman

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
```diff
@@ -11,25 +11,34 @@ The default browser behavior displays content following LTR formatting. In order
...
+
+Additionally, when in RTL, ensure that the `<ChartAxis>` representing a chart's x-axis uses the `invertAxis` property to flip the axis.
+
+**Note:** When using self-defined custom legends and label components, RTL support must be handled manually as `legendDirection` does not apply to custom implementations.
+
+### Icons
+PatternFly components have also been written to reverse directional icons that are part of the component. For example, the caret indicating expansion of various components is properly reversed. 
+
+Care should be taken to handle icons appropriately for RTL. Some icons you may have in your interface will not be mirrored automatically. [Sometimes an icon should be reversed in RTL](https://m2.material.io/design/usability/bidirectionality.html#mirroring-elements) (for example, when indicating a direction such as an expansion caret), and sometimes it should not be reversed (such as a logo, or an icon indicating time). When an icon *should* be reversed, but isn't reversed automatically, you can use the modifier `.pf-v5-m-mirror-inline-rtl`.
```
mcoker, line 32: Do we want to mention `<Icon shouldMirrorRTL>` here, too? I know we're mentioning both the modifier class and the `<Icon>` component support below - should we maybe just update this to "... you can use one of the methods mentioned in our [RTL modifiers](#whatever-the-anchor-name-is-to-that-section)"?

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
```diff
@@ -11,25 +11,34 @@ The default browser behavior displays content following LTR formatting. In order
...
 
 ## PatternFly modifiers supporting RTL
```
mcoker, line 34: Since this isn't just a section about modifiers, should we call this helpers or something more generic?

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
```diff
@@ -11,25 +11,34 @@ The default browser behavior displays content following LTR formatting. In order
...
 
-Use the modifier class `.pf-v5-m-mirror-inline-rtl` to reverse an icon or other element only when in RTL.
+-To reverse an icon or other element when in RTL, use the modifier class `.pf-v5-m-mirror-inline-rtl` or the `shouldMirrorRTL` property of the `<Icon>` React component.
 
 To set direction manually on an element, use `.pf-v5-m-dir-ltr` or `.pf-v5-m-dir-rtl`.
```
mcoker, line 38: I wonder if it's worth noting that the preferred way of defining direction on an element is via the native `dir` html attribute, eg, `dir="rtl"` or `dir="ltr"`? wdyt @srambach

mcoker (approved):
LGTM, left some comments, nothing is a blocker though

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
tlabaj: Yes, I think we should mention that. I like the sentence you have here @mcoker

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
edonehoo: If someone uses dir for this, does that negate the need for the modifier class or the shouldMirror prop? If so - in what cases would a developer use one vs the other out of all these options?

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
edonehoo: yes, I like your suggestion better!

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
mcoker: Great question! You would use one or the other, so no need to use `dir` and either of those classes together.

Probably the best way to suggest when to use `dir` or a class would be - the classes are meant for:

* Changing the direction of something as a stylistic/presentational-only change. 
* On the off chance a dev is unable to change the markup (to add a `dir` attribute), they can use the classes as an alternative.

Otherwise, you'd always want to use `dir` for anything where the writing direction should change for language support.

edonehoo: okay, gotcha! Can you look at my recent changes and lmk if you think they effectively convey this info?

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
```diff
@@ -15,7 +15,7 @@ The default browser behavior displays content following LTR formatting. In order
 
 PatternFly components have been written to use [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values) - `block-start`, `inline-end`, `block-end`, and `inline-start` - rather than `top`, `right`, `bottom`, and `left`. This allows styles like margin and padding to be correctly laid out in both LTR and RTL. However, to avoid a sweeping breaking change, PatternFly variable names still use `top`, `bottom`, `left`, and `right`.
 
-**Note:** An exception to this support is with any components using popper.js, like tooltips and popovers. 
+**Note:** An exception to this support is with any components using popper.js, like tooltips and popovers. For now, the popper position and placement will need to be managed manually when used on pages with RTL content.
```
mcoker, line 18: 👍

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
```diff
@@ -33,9 +33,11 @@ Care should be taken to handle icons appropriately for RTL. Some icons you may h
 
 ## PatternFly RTL helpers
 
-To reverse an icon or other element when in RTL, use the modifier class `.pf-v5-m-mirror-inline-rtl` or the `shouldMirrorRTL` property of the `<Icon>` React component.
+The preferred approach for defining an element's direction is to use the `dir` HTML attribute: `dir="rtl"` or `dir="ltr"`. If you are unable to use this attribute for any reason, you can instead utilize one of the following approaches to reverse an icon or other element when in RTL:
 
-To set direction manually on an element, use `.pf-v5-m-dir-ltr` or `.pf-v5-m-dir-rtl`.
+- Use the `.pf-v5-m-mirror-inline-rtl` class
+- Manually set an element's direction with `.pf-v5-m-dir-ltr` or `.pf-v5-m-dir-rtl`
+- Use the `shouldMirrorRTL` property of the `<Icon>` React component
```
mcoker, line 40: The way this reads, I wonder if it's conflating a couple of ideas. I could be reading it wrong though 😅 So this may be known, but there is a difference between:
1. `dir`, `.pf-v5-m-dir-ltr`, and `.pf-v5-m-dir-rtl`
  * This changes the "writing direction", and can apply to the whole document or individual elements, and can be nested (the document can be RTL, a block within the document can be LTR, and a block within the LTR block can be RTL, and so on). It basically just tells the browser the writing mode is from right to left. It will attempt to change the order of things to start from the right and end at the left. It's a browser feature with a bunch of logical stuff that happens to try and accommodate an LTR language. Nothing will happen with an icon, it's more for the way content and boxes are laid out.
    * `dir` should be used when the writing mode of a thing needs to be set. The `.pf-v5-m-dir[...]` classes will change the writing direction, but only visually/presentationally.
3. `.pf-v5-m-mirror-inline-rtl` and `<Icon shouldMirrorRTL />`
  * They just flip something horizontally. They don't really have anything to do with RTL/LTR per se, aside from these helpers have been written to only apply when they're used in an RTL block.

mcoker (dismissed):

packages/documentation-site/patternfly-docs/content/developer-resources/rtl-handbook.md
edonehoo: no you're right! I was definitely understanding this differently. I think I'm _mostly_ following, so I just pushed some changes that hopefully incorporate some of this distinction! Pls keep correcting the content as needed though 🙏

mcoker (approved):
🎃👍

PR #3774, opened 2023-10-09
docs(accessibility-scorecard): Updates content as part of content audit.

### LINKED ISSUE #3773: Review and update accessibility scorecard page

edonehoo:
Closes #3773


packages/documentation-site/patternfly-docs/content/developer-resources/accessibility-scorecard.md
```diff
@@ -4,12 +4,16 @@ section: accessibility
 ---
 import { AccessibilityScorecard } from './accessibility-scorecard.js';
 
+## Scoring your accessibility
+
+To support the proper assessment of accessibility measures, we've created a scorecard that outlines the essential criteria to meet. You can use this scorecard as a guide to evaluate the state of accessibility for any UI built using PatternFly.
```
jessiehuff, line 9: I'm wondering if we should say something about how these are the way in which we test, not just criteria per say? Maybe something like, "we've created a scorecard that outlines the ways that we recommend testing for accessibility and how your UI meets those expectations." That probably needs word-smithing. 🙃

thatblindgeye (commented):
A good followup might be linking to the Google sheet of the report (can't remember if it's public viewable right now), and/or maybe expanding on each of the criteria (what it means, why, how, etc).

One nit would be that maybe we include some verbiage about how a scorecard should be done routinely rather than "do it once and you're good forever". 

Also wondering if we could convey how the scorecard relates to the testing page a bit more, like what the scorecard is for vs a more in depth accessibility audit. Maybe something like, "An accessibility scorecard is a higher level overview of the current state of accessibility, an accessibility audit is a piece of that scorecard" (except word bettered 😆 ). This isn't a blocker, though, thinking out loud as I'm trying to think whether a reader may wonder "well what do I need the a11y audit page for if I have this scorecard" or something.

Overall I think this is looking good, though! Let me know what you think of the above, nothing is really a blocker so I'd be good with approving it as-is if we want

edonehoo:
@thatblindgeye I do really like these suggestions! I made some updates to head that direction, but I can build them out a little more if that would be better. 

I couldn't find a Google Sheet like what you're describing - if you know where that is? Expanding on the criteria also sounds helpful - maybe even in an expandable table/data list? That could be worth saving for a followup, though

thatblindgeye (approved):
I think this is looking great as is! Definitely some followups can be done, nothing I would really block over here.

@jessiehuff do you recall where the scorecard (the one published on org or the template) lives?

jessiehuff (approved):
Love your suggestions, @thatblindgeye! The scorecards are currently google docs ([scorecard template](https://docs.google.com/document/d/11CgQi_xIrGHdX2st0eCBMf2lVAxcFV2iD4f7KC_Hmkg/edit) and [org's 2022 scorecard](https://docs.google.com/document/d/1stdDUUfJKAUGqSt2TBkfiAacQvKrgD-dToh0gMMv8nA/edit)). I hesitate to link to the doc for the exact reason you mentioned. I'm fairly positive that our google docs are available to all Red Hatters, but I don't think that they're publicly available.

PR #3789, opened 2023-10-20
chore(docs-framework): add class on example load

### LINKED ISSUE #3709: Indication that the window/page assets have loaded
The main motivation for this is with the screenshot/visual regression testing tool in our various workspaces. If an example/demo has a large image or some other kind of asset, it would be great if our screenshot tool could wait until everything on the page has fully loaded before grabbing the screenshot. Our screenshot tool supports `readySelector` (a class somewhere on the document) and `readyEvent` (a string logged to the console) as options to trigger a screenshot

I've `window.onload` in the past to do something like add a `.page-loaded` class to `<body>` or log "page loaded" to the console, but I'm not sure if that's valid or outdated in a react app. Is that something we could do for the pages served from the docs framework?

cc @nicolethoen

wise-king-sullyman:
Closes #3709


mcoker (commented):
To test, I put a link to this 1.7mb image on an example page and emulated a 3g connection in chrome, and the class was added to `<body>` before the image loaded. 

If you try that with a `window.load` event handler, it doesn't run until the image has finished downloading - https://codepen.io/mcoker/pen/MWLgdYw

Is there something similar using react?

mcoker (approved):
🦾🤖 bleep bleep boop!!

evwilkin (approved):
LGTM

PR #3803, opened 2023-11-06
chore(ExpandableSection): updated a11y docs to match template

### LINKED ISSUE #3800: Update a11y docs for Expandable Section

thatblindgeye:
Closes #3800 

Couple of things to address before this gets merged in:

- In the React implementation we set an aria-labelledby on the content component with a value of the toggle. Essentially an expandable section content will have an accessible name of "Show less", possibly announced as "Show less, region". Which isn't exactly descriptive. I'm wondering if we should be adding aria-labelledby at all (and a role="region"), or if we should recommend providing more descriptive toggle text, e.g. "Show more for [content subject]".
- In React we're also always setting `aria-controls`. Doesn't really hurt to have it, but wondering if we really need it outside teh detached examples where it makes more sense. A detached example could have several other things between the toggle and content for whatever reason, but a non-detached example will always have the toggle immediately before the content.

For either bullet above, the outcome will affect the Core section in this PR and possibly require updates to Core examples. Currently I don't mention aria-labelledby for the HTML attributes, and the aria-controls verbiage doesn't mention "...only for a detached expandable section".


nicolethoen (dismissed):

nicolethoen:
Opening an accessibility issue for the core/react docs is totally reasonable. I like both the suggestions you make int he bullets. I think that feels more straight forward for the consumers.

Do you need to make an update to this before it's merged?

nicolethoen (approved):

PR #3805, opened 2023-11-07
chore(HelperText): updated a11y docs to match template

### LINKED ISSUE #3804: Update a11y docs for Helper Text

thatblindgeye:
Closes #3804

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/accessibility/helper-text/helper-text.md
+++ b/packages/documentation-site/patternfly-docs/content/accessibility/helper-text/helper-text.md
@@ -3,50 +3,84 @@ id: Helper text
 section: components
 ---
 
-**Helper text** is text that often accompanies inputs and other form components. It can be static text that provides additional or clarifying information and will never change, or it can be dynamic text that gets rendered or updated based on user interaction or input.
+import { Checkbox, List, ListItem } from '@patternfly/react-core';
 
-**Keyboard users** should not be able to place focus on any helper text, as it is not an interactive element.
+## Accessibility
 
-**Screen reader users** should have static helper text announced to them via the `aria-describedby` attribute when focus is placed on a form component, and they should also have dynamic helper text announced to them via the `aria-live` attribute when the text content gets updated.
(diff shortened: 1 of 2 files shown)


packages/documentation-site/patternfly-docs/content/accessibility/helper-text/helper-text.md
```diff
@@ -3,50 +3,84 @@ id: Helper text
...
+    </HelperTextItem>
+    <HelperTextItem component="li">
+      Some other helper text content
+    </HelperTextItem>
+  </HelperText>
+  ```
+- Structure the helper text as a `ul` element and each helper text item as a `li` element if there is more than one helper text item.
+- Make the helper text component an `aria-live` region if the helper text item(s) have static text and styling, but will dynamically render.
+- Ensure there is visually hidden text for assistive technologies such as screen readers that conveys the status type if the helper text item has static text and will always be rendered, but the styling will dynamically update.
+- Avoid passing interactive content as helper text.
```
thatblindgeye, line 34: Currently [helper text design guidelines](https://www.patternfly.org/components/helper-text/design-guidelines#content-considerations) mention that a link to additional info can be provided, so maybe some input from @andrew-ronaldson or @mmenestr would be helpful here. Is that still an intended use case for a helper text? Just wondering whether helper text should include content that has to be navigated to on another page, or if it should only include content that is immediately relevant for a user.

mmenestr: Hmm to be honest, I'd say not. I might edit the guidelines to delete that bullet because I don't think it makes sense. It only makes sense to add a link to external information in a popover!

thatblindgeye: I can remove that bullet from the design guidelines as part of this PR if you'd like?

nicolethoen: @thatblindgeye go for it

nicolethoen (approved):

PR #3808, opened 2023-11-07
chore(JumpLinks): updated a11y docs to match template

### LINKED ISSUE #3807: Update a11y docs for Jump Links

thatblindgeye:
Closes #3807 

Not mentioned in the docs, but looking at our core and react examples, rather than using `aria-label` when the jump links have a visible label, we should consider utilizing `aria-labelledby` instead. Would just require a couple new props in React, and a followup needs to be opened in Core to add the `aria-current` attribute in examples, but interested what others think.

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/accessibility/jump-links/jump-links.md
+++ b/packages/documentation-site/patternfly-docs/content/accessibility/jump-links/jump-links.md
@@ -3,32 +3,82 @@ id: Jump links
 section: components
 ---
 
-**Jump links** provide a way to navigate through section headings on a page without scrolling.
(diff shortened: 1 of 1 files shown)


nicolethoen (approved):
LGTM 
We could open an issue to add an arialabelledby prop for usage when appropriate.

As part of that, would we also want to log a warning or error if there is NO aria-label nor aria-labelledby? it's required, right?

thatblindgeye:
@nicolethoen I feel like one or the other should be required, yeah. Even if there aren't any other nav elements on a page, it'd still be helpful to provide some context as to "what is this nav for?". Will open followup issues in both repos

PR #3866, opened 2024-01-17
fix: add drag drop deps and remove unused code editor deps

nicolethoen:


kmcfaul (commented):
Changes look good, glad it's building fine with the CodeEditor swap (it must have been just symlinking that was causing issues for me locally). 

I am seeing two react minified errors on the surge, but I'm not sure what's causing them because it's parsing down to a generic hydration error. If it's the drag/drop export warning, I'm working on that on the react side and this should be good to go.

nicolethoen:
@kmcfaul when I run this branch locally, I see the following error:
Screenshot 2024-01-18 at 3 25 41 PM

kmcfaul (approved):
I put up https://github.com/patternfly/patternfly-react/pull/10023 which should address the export warning.

tlabaj (dismissed):

Issue #3852, opened 2024-01-04
Bug - Beta tags problematic in navigation and title for Next components


tlabaj:
**Describe the problem**

When a next componet is introduced as Beta,  

- A `Beta` label is added to the navigation item for the componet even though only the next implementation is Beta. 
- A `Beta` label is added to the componet title on the componet page.

**Expected behavior**

Beta should only be applied to the next componet.  Adding a label at the top level implies both implementations are beta.

**Screenshots**
If applicable, add screenshots to help explain the issue.

![image](https://github.com/patternfly/patternfly-org/assets/39532947/b6c81d47-3711-4b7b-bc94-78dcf3e04adc)

**Any other information?**

PR #3868, opened 2024-01-18
fix: avoid applying beta tags to nav items for next components

nicolethoen:
Closes: https://github.com/patternfly/patternfly-org/issues/3852


evwilkin (approved):
LGTM, thanks!

Issue #3609, opened 2023-06-16
Restructure new homepage footer


thatblindgeye:
Followup to https://github.com/patternfly/patternfly-org/pull/3573

Currently the structure of the homepage is setup like so:

```
<header />
<main>
    <footer />
<main>
```
It'd be worth restructuring the layout so that the site wide footer content is outside of the main element:

```
<header />
<main />
<footer />
```

So that the footer content appears as a landmark in rotor menus for assistive tech. Additionally//alternatively, the footer content is followed by a final PageSection component (which houses the Red Hat logo, copyright info, etc) which could be combined into the `<footer>`. If we were to move the footer outside of the `main` element then combining the two makes more sense, but keeping the footer inside of `main` there may be less concern (([W3C](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-footer-element.html) mentions how footers don't _need_ to appear at the end of a section, but _usually_ do)

PR #3870, opened 2024-01-19
chore(footer): Restructure new homepage footer

jessiehuff:
Closes https://github.com/patternfly/patternfly-org/issues/3609

### DIFF
--- a/packages/documentation-framework/app.js
+++ b/packages/documentation-framework/app.js
@@ -37,7 +37,6 @@ const AppRoute = ({ child, title, path }) => {
   return (
     <React.Fragment>
       {child}
-      {process.env.hasFooter && <Footer />}
     </React.Fragment>
   );
 }
--- a/packages/documentation-framework/app.js
+++ b/packages/documentation-framework/app.js
@@ -3,7 +3,6 @@ import { createRoot, hydrateRoot } from 'react-dom/client';
 import { Router, useLocation } from '@reach/router';
 import 'client-styles'; // Webpack replaces this import: patternfly-docs.css.js
 import { SideNavLayout } from '@patternfly/documentation-framework/layouts';
-import { Footer } from '@patternfly/documentation-framework/components';
 import { MDXTemplate } from '@patternfly/documentation-framework/templates/mdx';
 import { routes, groupedRoutes, fullscreenRoutes, getAsyncComponent } from './routes';
 import { trackEvent } from './helpers';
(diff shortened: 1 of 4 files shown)


nicolethoen:
It seems this change does impact the layout of the website quite a bit. I wonder if there is some CSS overrides causing this behavior, or if there is some CSS (either in core, or in org) needed to address this.

WDYT @mcoker @srambach @andrew-ronaldson?

Before:
Screenshot 2024-01-30 at 3 15 57 PM

After:
Screenshot 2024-01-30 at 3 16 15 PM

mcoker:
@nicolethoen @jessiehuff looks like it's just because the footer was moved out of the page component (it was in the main content area), and now comes after the page component. 

The page component is set up so that it fills the entire viewport height, and then we scroll the contents of 1) the sidebar and 2) the main content area, if there is any overflow in those sections. So what I'm seeing is the page main content scrolling, then when you get to the bottom, the whole browser window scrolls to scroll the page component out of view and scroll the footer into view, which is why there are two scrollbars here:

Screenshot 2024-02-06 at 9 51 57 AM

Is the intent that the footer span the whole width of the viewport? Or can it go back in the bottom of the main content area?

nicolethoen:
@mcoker it's updating the markup because (according to the issue) the ideal markup for a page would be:
```
<header />
<main />
<footer />
```

but right now it's 
```
<header />
<main>
    <footer />
<main>
```

mcoker:
@nicolethoen ah sorry I should have reviewed the original issue. I'm not sure how we can add this without ending up with double scrollbars. It would be fine if the footer was fixed/sticky to the bottom of the viewport (like masthead is to the top), but it's too tall to do that.

I have one idea - if it's a matter of fixing the markup so it's more a11able, I wonder if we could do this below. Assuming this works, it seems pretty easy, keeps the existing structure and scrolling and all of that, just changes the html element used in 2 spots for the a11y goodness.
  * Change the existing `main.pf-v5-c-page__main` to `div.pf-v5-c-page__main`
  * The direct child of the current main element is `div#ws-page-content-router` - change that to `main#ws-page-content-router`
  * Append the `<footer>` to `div.pf-v5-c-page__main`, so it comes after `main#ws-page-content-router`

Otherwise I'd probably defer to @lboehling @andrew-ronaldson on a recommendation or example somewhere to refer to for what the scrolling behavior should be while retaining a good UX - what elements scroll and when, what elements are "sticky" and when, stuff like that

mcoker (approved):
Nice!! 🎃👍

andrew-ronaldson (approved):
sweet, LGTM

mcoker:
Gave this feedback to @nicolethoen and @jessiehuff offline but just for reference, a page section adjacent to the other sections with `.pf-m-sticky-top` should work without any modifications to the styles.

Screenshot 2024-03-19 at 6 20 21 PM

It can probably be moved out of `#ws-page-content-router` and included directly under `.pf-v5-c-page__main`, too - from what I tested, that seems OK and I didn't see any issues.

mcoker:
💪🐰🥕

mcoker (approved):

PR #3871, opened 2024-01-19
chore(SkipToContent): updated a11y docs to template

### LINKED ISSUE #3825: Update a11y docs for Skip To Content

thatblindgeye:
Closes #3825

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/accessibility/skip-to-content/skip-to-content.md
+++ b/packages/documentation-site/patternfly-docs/content/accessibility/skip-to-content/skip-to-content.md
@@ -24,7 +24,7 @@ At a minimum, a skip to content should meet the following criteria:
     <Checkbox id="skip-to-content-a11y-checkbox-2" label={<span>The skip to content's <code className="ws-code">href</code> is linked to the main content of the page.</span>} />
   </ListItem>
   <ListItem>
-    <Checkbox id="skip-to-content-a11y-checkbox-3" label="Standard keyboard navigation can be used to navigate between the skip to content and other focusable elements." description={<span><kbd>Tab</kbd> navigates to the skip to content when at the beginning of a page or the next focusable element thereafter, and <kbd>Shift</kbd> + <kbd>Tab</kbd> navigates to the previous focusable element or the skip to content.</span>} />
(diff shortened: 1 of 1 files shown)


packages/documentation-site/patternfly-docs/content/accessibility/skip-to-content/skip-to-content.md
```diff
@@ -3,12 +3,46 @@ id: Skip to content
...
+
+<List isPlain>
+  <ListItem>
+    <Checkbox id="skip-to-content-a11y-checkbox-1" label="The skip to content is the first focusable element on the page." description="When content is repeated on several pages - such as actions in a header or navigation links in a sidebar - users navigating via keyboard or other assistive technologies would need to navigate through that content every time. Placing a skip to content as the first focusable element helps users avoid this repeated content." />
+  </ListItem>
+  <ListItem>
+    <Checkbox id="skip-to-content-a11y-checkbox-2" label={<span>The skip to content's <code className="ws-code">href</code> is linked to the main content of the page.</span>} />
+  </ListItem>
+  <ListItem>
+    <Checkbox id="skip-to-content-a11y-checkbox-3" label="Standard keyboard navigation can be used to navigate between the skip to content and other focusable elements." description={<span><kbd>Tab</kbd> navigates to the skip to content when at the beginning of a page or the next focusable element thereafter, and <kbd>Shift</kbd> + <kbd>Tab</kbd> navigates to the previous focusable element or the skip to content.</span>} />
```
jessiehuff, line 27: I find this description just a tad confusing. I previously tried to explain what to expect with the keyboard interactions surrounding the skip to content, that it was essentially still in the tab order and that users could interact with it or move past it/that it doesn't force the user into using it. I think the part that I find particularly confusing is this line, "Shift Tab navigates to the previous focusable element or the skip to content." For some reason this feels odd, like we're just describing general keyboard interactions?

thatblindgeye: That's a fair point, yeah this is basically the keyboard navigation content that is used in other components' a11y docs. It does end up a bit wordy because it's sort of a special case in that the first Tab on a page should focus the Skip to Content.

Do you think this item is necessary, or should we just omit the Shift + Tab part? Could also reword it a little, though not sure if this really improves it significantly:

```
Depending on where focus is on the page, <kbd>Tab</kbd> navigates to the skip to content or the next focusable element, and <kbd>Shift</kbd> + <kbd>Tab</kbd> navigates to the previous focusable element or the skip to content. Since the skip to content should be the first focusable element on a page, the previous focusable element may typically be an item in the browser toolbar.
```

Basically trying to convey that Tab and Shift + Tab work and what the expected focus should be.

jessiehuff: Yeah, I think the edit makes it a lot clearer. It adds context so I better understand what you're saying. I'm good with that, but if you think it's better to remove it altogether, that's fine with me too.

jessiehuff (approved):
Awesome job as always! 🎉

PR #3876, opened 2024-01-30
Algolia fix

### LINKED ISSUE #3683: v4-archive site search returns v5 pages
Searching on v5 seems to work fine, but when I'm searching on v4, it gets  me back to v5 results.

jessiehuff:
Closes #3683 
Adding new version number to bump v4 staging site

### DIFF
--- a/packages/documentation-framework/versions.json
+++ b/packages/documentation-framework/versions.json
@@ -1,9 +1,30 @@
 {
   "Releases": [
     {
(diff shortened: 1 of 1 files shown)


evwilkin:
FYI - this is just adding an entry to versions.json that we're adding before bumping the V4 archive site with the correct algolia ID (which was added to Org in a prior PR, _after_ the site was updated)

PR #3880, opened 2024-02-05
chore(Page): updated a11y docs to new template

### LINKED ISSUE #3874: Page

thatblindgeye:
Closes #3874

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/accessibility/page/page.md
+++ b/packages/documentation-site/patternfly-docs/content/accessibility/page/page.md
@@ -3,30 +3,81 @@ id: Page
 section: components
 ---
 
-The **page** component is used to define the basic layout of a page with either vertical or horizontal navigation. Page layouts are defined using page sections, such as the header, body, and footer of a page. The basic layout of a page depends on whether your page uses vertical or horizontal navigation.
+import { Checkbox, List, ListItem } from '@patternfly/react-core';
 
-A page can be combined with many other components, so to create an accessible page, refer to the accessibility recommendations for each component.
+## Accessibility
 
-**Keyboard users** should be able to navigate through the masthead of the page and into other interactive elements of the page using **Tab** and **Shift + Tab** to move forward and backward through interactive elements. 
+To implement an accessible PatternFly **page**:
 
-**Screen reader users** should be able to navigate through and interact with the masthead of the page using the same interactions as keyboard users. Elements of a page that do not have visible descriptive text, such as toggle icons, should have alternative text for screen readers.
+- Provide a [skip to content](/components/skip-to-content) if more than one page shares the same heading or sidebar content.
+- Give each `nav` element on the page a unique `aria-label`.
+- Ensure there is only 1 `main` element on the page.
+- Ensure each page section has an `aria-labelledby` attribute linked to a heading element within the section.
+  - If there is no heading within the section, provide an `aria-label` instead.
+- Follow any accessibility documentation for other components used within a page, such as a [breadcrumb](/components/breadcrumb), [notification drawer](/components/notification-drawer), or [navigation](/components/navigation).
 
-Consider using a **skip to content** component on your page so that assistive technology users don’t need to make their way through the entire navigation menu each time they go to another page.
+## Testing
 
-## To make a page accessible:
-- Use semantic elements when possible, as these will have default roles. If this is not possible, manually add roles to identify different regions of the page as needed.
-- If there are multiple instances of one type of semantic element (for instance, two `<nav>` items), label the elements.
-- For every element that does not contain text, add an `aria-label` attribute that contains alternative text.
-- If using notifications in the masthead, refer to notification badge accessibility guidelines.
+At a minimum, a page should meet the following criteria:
 
-The following props/attributes have been added for you or are props/attributes that can be customized:
+<List isPlain>
+  <ListItem>
+    <Checkbox id="page-a11y-checkbox-1" label="If more than one page has the same header and/or sidebar content, a skip to content is passed to the page." description="This allows users to skip repetitive content, as otherwise they would have to navigate through each heading and sidebar item on every page." />
+  </ListItem>
+  <ListItem>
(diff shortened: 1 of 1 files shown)


packages/documentation-site/patternfly-docs/content/accessibility/page/page.md
```diff
@@ -3,30 +3,81 @@ id: Page
...
+## Accessibility
 
-**Keyboard users** should be able to navigate through the masthead of the page and into other interactive elements of the page using **Tab** and **Shift + Tab** to move forward and backward through interactive elements. 
+To implement an accessible PatternFly **page**:
 
-**Screen reader users** should be able to navigate through and interact with the masthead of the page using the same interactions as keyboard users. Elements of a page that do not have visible descriptive text, such as toggle icons, should have alternative text for screen readers.
+- Provide a [skip to content](/components/skip-to-content) if more than one page shares the same heading or sidebar content.
+- Give each `nav` element on the page a unique `aria-label`.
+- Ensure there is only 1 `main` element on the page.
+- Ensure each page section has an `aria-labelledby` attribute linked to a heading element within the section.
```
jessiehuff, line 15: This is a really good point, and potentially unrelated to this PR, but are we actually following this recommendation in our Page demos?

thatblindgeye: Can't recall off the top of my head but erring on the side of caution I want to say we may not be. That would require a follow up in Core and React, which I can evaluate which, if any, examples/demos need to be updated and create any necessary followups. Probably need to go through Org as well.

You're referring to the last bullet regarding aria-labelledby, correct?

packages/documentation-site/patternfly-docs/content/accessibility/page/page.md
```diff
@@ -3,30 +3,81 @@ id: Page
...
 
-A page can be combined with many other components, so to create an accessible page, refer to the accessibility recommendations for each component.
+## Accessibility
 
-**Keyboard users** should be able to navigate through the masthead of the page and into other interactive elements of the page using **Tab** and **Shift + Tab** to move forward and backward through interactive elements. 
+To implement an accessible PatternFly **page**:
 
-**Screen reader users** should be able to navigate through and interact with the masthead of the page using the same interactions as keyboard users. Elements of a page that do not have visible descriptive text, such as toggle icons, should have alternative text for screen readers.
+- Provide a [skip to content](/components/skip-to-content) if more than one page shares the same heading or sidebar content.
+- Give each `nav` element on the page a unique `aria-label`.
```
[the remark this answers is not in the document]
thatblindgeye: Ah good catch, agreed we should be consistent in the verbiage.

Though now I'm slightly conflicted, in that I'm wondering if we should say that a nav should have a unique label regardless of amount, i.e. whether there's 7 or only 1. I guess I'm thinking in terms of rotor menus/wherever a list of X element types are listed out (lists, headings, etc). Even with only 1 nav (or list, or whatever) on a page, I feel like providing an aria-label can only help, whereas omitting one could possibly hurt?

For example, if the nav is announced as just "nav' to a user, it could be site navigation, page navigation, breadcrumb navigation, etc. So what the user would be jumping to could be a bit of an unknown.

What do you think, though?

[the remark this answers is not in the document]
jessiehuff: I think it definitely could be helpful, but I'm wondering if it's fully required. I wonder if we could explain how it can be helpful but say that it's only required when there's multiple? (And maybe we could even give the examples you just stated.)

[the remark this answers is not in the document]
thatblindgeye: Maybe, "If there are multiple `nav` elements on the page, give each one a unique `aria-label`." as the checkbox label, then a description of, "You should also consider providing an `aria-label` when there's only a single `nav` on the page, if the purpose of that nav is not the primary navigation."

Or something along those lines? Not sure how I feel about "primary navigation"; "primary site navigation", really just wanting to convey "something that isn't the typical navigation for a page"

packages/documentation-site/patternfly-docs/content/accessibility/page/page.md
thatblindgeye: I thought it was an MDN docs page but maybe not, it may have been traversing through VO's rotor menu again. Or one of our a11y pages that mention sections shouldn't be used if there's no label/our example of providing a label for scrollable content.

If we do want to keep this it would probably make sense to update it to say use either aria-labelledby or aria-label if no visible text is present, but it may actually make sense to remove (even if for now)

packages/documentation-site/patternfly-docs/content/accessibility/page/page.md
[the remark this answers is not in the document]
jessiehuff: Yeah, I think that makes sense! And I think calling it a primary navigation isn't inherently bad. You could even work with those words a bit between both and say something like, "You should also consider providing an `aria-label` when there's only a single nav on the page, if the purpose of that nav isn't a typical, primary navigation for a page." (To be honest though, I think all of these options are probably fine though and hopefully get the point across.)

packages/documentation-site/patternfly-docs/content/accessibility/page/page.md
jessiehuff: Yeah, it makes sense to do, and I'm always up for adding more/better context to help users. I just hesitate to make recommendations if we aren't already walking that walk, so to speak. 🙂

thatblindgeye: Definitely true. I think my thoughts were get a11y docs up to date and update core/react examples as follow ups according to the updated docs, though it may make more sense to not include stuff that we aren't following yet ourselves.

What do you think about me removing the verbiage about labeling page sections from this PR, opening issues in core/react to add consistent labeling, then once those are closed out we open a followup in Org to add in the verbaige about labeling sections?

jessiehuff: I think that would make sense to me!

thatblindgeye:
@jessiehuff https://github.com/patternfly/patternfly-react/issues/10173 and https://github.com/patternfly/patternfly/issues/6431 regarding convo above, otherwise updates have been made here

jessiehuff (approved):
LGTM!

Issue #3551, opened 2023-05-24
Update information about react-tokens to docs


nicolethoen:
react-tokens map to most or all CSS variables in patternfly, and there is no way for a consumer to look them up if they want to use them. This will become ever more important as we used versioned prefixes in our CSS, since the CSS selectors and variables will change every major release, but our react-tokens wont. 

our CSS variables tables in component docs and [Global CSS Variables page](https://www.patternfly.org/v4/developer-resources/global-css-variables) should include information about the corresponding react-tokens.

PR #3886, opened 2024-02-12
global vars - remove selector prop, show selector column

evwilkin:
Towards #3551 

Note: please leave any suggestions for additional enhancements/changes, I wasn't clear if this was all we need currently.

This PR:
- Bumps versions to latest alphas
- Removes the [`selector`](https://github.com/patternfly/patternfly-org/blob/v6/packages/documentation-framework/pages/global-css-variables.md?plain=1#L105) prop used in global-css-variables.md that was filtering to only show rules matching selector `:where(:root)`
  - it now shows all vars, but we could add in new selector `:root` to only show new vars (or leave as-is showing everything and wait until core cleanup?)
- Unhides the "selector" column from the global vars table
  - though all the new vars map to the `:root` selector, this column enables rows to be expanded (where applicable) to track the variable chain.
  - Screenshot 2024-02-12 at 6 47 03 PM

### DIFF
--- a/packages/documentation-framework/package.json
+++ b/packages/documentation-framework/package.json
@@ -80,10 +80,10 @@
     "webpack-merge": "5.8.0"
   },
   "peerDependencies": {
-    "@patternfly/patternfly": "6.0.0-alpha.38",
-    "@patternfly/react-code-editor": "6.0.0-alpha.5",
-    "@patternfly/react-core": "6.0.0-alpha.5",
-    "@patternfly/react-table": "6.0.0-alpha.5",
+    "@patternfly/patternfly": "6.0.0-alpha.78",
+    "@patternfly/react-code-editor": "6.0.0-alpha.29",
+    "@patternfly/react-core": "6.0.0-alpha.29",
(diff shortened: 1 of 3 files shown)


nicolethoen:
This is similar to what I built into the design token docs. I think the biggest difference (other than a few visual differences) is that building it from the design-tokens repo means that they are up to date with the latest in the tokens repo - rather than reading design tokens as passed from design-tokens -> core -> react -> org (which will always be a few steps behind).

nicolethoen:
I was also planning to have the token docs page go in the new 'tokens' subnav item that Erin was building.

evwilkin:
@nicolethoen that all makes sense and sounds great to me, would you suggest I close this PR out in favor of the momentum you have going in your related issues?

nicolethoen:
It's possible? Though I hadn't even thought of how the Global CSS vars might be the same or almost the same as the design tokens once all is said and done. It also may be worth a conversation to see what we should do with this page on the v6 site.

evwilkin:
No longer needed

PR #3902, opened 2024-02-27
feat(templates): add templates package to versions.json

### LINKED ISSUE #3898: Add react-templates to latest versions.json
Required when the new react-templates package is added to build the docs.

kmcfaul:
Closes #3898 

Adds the templates package to `versions.json` to allow the docs to build successfully. Currently version is set to `1.0.0-alpha.0` but may be changed depending on what number we want to start on (1 vs 5 to match other PF packages).

### DIFF
--- a/packages/documentation-framework/templates/mdx.js
+++ b/packages/documentation-framework/templates/mdx.js
@@ -23,6 +23,7 @@ const MDXChildTemplate = ({
     optIn,
     beta,
     deprecated,
+    template,
     newImplementationLink,
     functionDocumentation = []
   } = Component.getPageData();
@@ -82,6 +83,12 @@ const MDXChildTemplate = ({
             </React.Fragment>
           )}
           {' '}To learn more about the process, visit our <Link to="/get-started/about#major-release-cadence">about page</Link>.
+        </InlineAlert>
+      )}
+       {(template || source === 'react-template') && (
+        <InlineAlert title="Templates" variant="info">
+          {`This section showcases templates for the ${id} component. A template is a wrapped ${id} component use case that has built in logic with a streamlined API, and some limited customization.
+          For custom use cases that fall outside of a template's design, please use the ${id} component suite directly.`}
         </InlineAlert>
       )}
     </React.Fragment>
--- a/packages/documentation-framework/templates/mdx.js
+++ b/packages/documentation-framework/templates/mdx.js
@@ -87,7 +87,7 @@ const MDXChildTemplate = ({
       )}
        {(template || source === 'react-template') && (
         <InlineAlert title="Templates" variant="info">
-          {`This page showcases templates for the ${id.toLowerCase()} component. A template consists of the component and additional logic supporting the use case, and has a streamlined API with some limited customization.`}
+          {`This page showcases templates for the ${id.toLowerCase()} component. A template combines a component with logic that supports a specific use case, with a streamlined API that offers additional, limited customization.`}
         </InlineAlert>
       )}
     </React.Fragment>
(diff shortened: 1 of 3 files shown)


kmcfaul:
Added logic for a new `template` flag to be passed in via `md` to show an inline alert for a template page. LMK if `info` is appropriate for this alert or if another variant should be used instead.

packages/documentation-framework/templates/mdx.js
```diff
@@ -84,6 +85,12 @@ const MDXChildTemplate = ({
           {' '}To learn more about the process, visit our <Link to="/get-started/about#major-release-cadence">about page</Link>.
         </InlineAlert>
       )}
+       {(template || source === 'react-template') && (
+        <InlineAlert title="Templates" variant="info">
```
tlabaj, line 89: @edonehoo Can you please review the content here?

evwilkin (approved):
LGTM - fyi here's a screenshot of how it looks, adding `template: true` to the frontmatter of a React example .md file:
Screenshot 2024-02-29 at 3 50 14 PM

When we have templates to pull in we'll need to update the [source file](https://github.com/patternfly/patternfly-org/blob/main/packages/documentation-site/patternfly-docs/patternfly-docs.source.js) to make the build aware of them and map them to the correct tab.

packages/documentation-framework/templates/mdx.js
```diff
@@ -84,6 +85,12 @@ const MDXChildTemplate = ({
...
+          {`This section showcases templates for the ${id} component. A template is a wrapped ${id} component use case that has built in logic with a streamlined API, and some limited customization.
```
edonehoo, line 90:
```suggestion
          {`This page showcases templates for the ${id} component. A template contains a wrapped component that has built-in logic, a streamlined API, and some limited customization.
```

What is the intended use of a template? Are they like fill-in-the-blank components? 

Having trouble understanding the second sentence well enough to suggest phrasing, but maybe that's because I am not familiar with templates as a concept

packages/documentation-framework/templates/mdx.js
edonehoo: - can we transform the component name variable to be lowercase?

packages/documentation-framework/templates/mdx.js
kmcfaul: Templates will be pre-wrapped demos basically, with a small API. Instead of a user needing to define additional boiler plate code like state and toggles (for select), they have the option to use a template which would only need some initial data and self-contain the rest of the logic to get it to function.

Updated and tweaked the second sentence a little to represent that a bit better, LMK what you think and if it needs 
 further updates 👍

wise-king-sullyman (approved):

packages/documentation-framework/templates/mdx.js
```diff
@@ -84,6 +85,11 @@ const MDXChildTemplate = ({
           {' '}To learn more about the process, visit our <Link to="/get-started/about#major-release-cadence">about page</Link>.
         </InlineAlert>
       )}
+       {(template || source === 'react-template') && (
+        <InlineAlert title="Templates" variant="info">
+          {`This page showcases templates for the ${id.toLowerCase()} component. A template consists of the component and additional logic supporting the use case, and has a streamlined API with some limited customization.`}
```
edonehoo, line 90:
```suggestion
          {`This page showcases templates for the ${id.toLowerCase()} component. A template combines a component with logic that supports a specific use case, with a streamlined API that offers additional, limited customization.`}
```
@kmcfaul thanks for clarifying -- I'm following better now! wdyt about this phrasing?

kmcfaul: Sounds great!

jessiehuff (approved):

nicolethoen (approved):

wise-king-sullyman (approved):

PR #3906, opened 2024-02-28
chore(dark theme styling): fix ws dark theme styles

andrew-ronaldson:
Closes https://github.com/patternfly/patternfly-org/issues/3910

### DIFF
--- a/packages/documentation-framework/components/sideNav/sideNav.css
+++ b/packages/documentation-framework/components/sideNav/sideNav.css
@@ -4,7 +4,7 @@
   --pf-v6-c-nav--m-light__link--focus--BackgroundColor: var(--pf-v6-global--BackgroundColor--light-200);
   --pf-v6-c-nav--m-light__link--m-current--BackgroundColor: var(--pf-v6-global--BackgroundColor--light-200);
   --pf-v6-c-nav--m-light__link--active--BackgroundColor: var(--pf-v6-global--BackgroundColor--light-200);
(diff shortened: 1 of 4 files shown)


srambach (approved):
🌔 LGTM

nicolethoen (approved):

nicolethoen (commented):
I approved, but then I noticed that at smaller screen widths the expandable jumplinks are looking incorrect in dark theme:
Screenshot 2024-03-19 at 2 14 43 PM
Screenshot 2024-03-19 at 2 14 54 PM

nicolethoen:
after the last change, something wonky is happening to the minified react. 
I'm seeing the same weird side effect when I run things locally. I'll see if I can track down what's happening...

andrew-ronaldson:
Screenshot 2024-03-21 at 10 39 23 AM
Noticed this bit of weirdness when the ToC gets expanded. It's working well enough for Alpha but good to fix in a follow up

nicolethoen (approved):

evwilkin:
FYI the collapsible ToC being cut off bug flagged above looks to be due to [this line applying `overflow: auto` to the component](https://github.com/patternfly/patternfly-org/blob/main/packages/documentation-framework/components/tableOfContents/tableOfContents.css#L26).  Removing the line looks to fix the issue, and I cannot confirm if/why that code needs to be present at all, so believe we can simply strip that one line out.

PR #3916, opened 2024-03-07
chore(TreeView): updated a11y docs to template

### LINKED ISSUE #3875: Update a11y docs for Tree View

thatblindgeye:
Closes #3875 

Some of the additions I've made here currently are not things implemented in either core or React, but either things I noticed were missing or things gleamed from some of [W3's tree view examples](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/). Some of this may depend on if we would want to make such changes, namely:

- `aria-multiselectable`: in React the "active items" of a tree view are intended to be placed into an array, suggesting multiple items can be selected. Currently we don't indicate that to assistive tech.
- A mention about custom interactive content, based on https://github.com/patternfly/patternfly-react/issues/8255 and the link mentioned in that issue
- When a search input must be included (7 root nodes) based on W3's tree view examples (linked above); this may be a bit arbitrary of a number, as even 4 root nodes may require a search if each node has 10 children (with each children having 5 of its own children)
- Again similar to W3 tree view examples, not allowing Tab to navigate between tree view items/content. Currently in React you can use either arrows or Tab, sometimes you need to use Tab. The complexity of some of the tree view examples may make it difficult to not use Tab (should Right Arrow either expand a node if there's no other interactive content, otherwise Right Arrow focuses each interactive content - expand button, selection button, action button - and only expands a node once it's reached the last interactive item?)

Other things like aria-label and aria-selected are things we need to add in, but some additions I'd be open to discussing further to see if they make sense overall.

There were several issues I noted that I'll use as part of the component audit for tree view and make issues in their respective repos as necessary.

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/accessibility/tree-view/tree-view.md
+++ b/packages/documentation-site/patternfly-docs/content/accessibility/tree-view/tree-view.md
@@ -3,32 +3,87 @@ id: Tree view
 section: components
 ---
 
-A **tree view** is a structure that displays data in a hierarchical view. It can be used in a dropdown, drawer, primary-detail, modal, or wizard. Tree views are often used when grouping similar decisions, action items, or navigation elements together.
+import { Checkbox, List, ListItem } from '@patternfly/react-core';
 
-A default tree view provides a tree structure with items and arrows to expand or collapse child items. Every tree view can be extended with these optional capabilities.
+## Accessibility
 
-**Keyboard users** should be able to interact with and navigate through the interactive tree view. Initial focus should be set on the first focusable element in the tree view. The user should be able to focus on one element of the tree with **Tab**. Once focused on a tree item, the user should also be able to move between visible nodes using **Up arrow** and **Down arrow**. **Left arrow** can be used to collapse an expanded node, or move to the parent of an already-collapsed or childless node. **Right arrow** should expand the focused node if it is collapsed but has child nodes, move to the first child of an already-expanded node, or do nothing on an end node.
+To implement an accessible PatternFly **tree view**:
 
-Keyboard interactions for tree view:
+- Provide a unique label for the tree view via `aria-label` or `aria-labelledby`.
+- Only pass custom interactive content as a tree view item's secondary action, and never as part of a tree view item's displayed name.
+- Ensure any interactive actions for a tree view item follow any applicable accessibility guideliens.
+- Provide a typeahead search when the tree view has more than 7 items at the root of the tree view.
 
-|  |  |
-| -- | -- |
-| Tab | Focus on the first (or previously selected) element of the tree |
-| Up arrow | Move up through visible nodes |
-| Down arrow | Move down through visible nodes |
-| Left arrow | Collapse focused node if expanded, or move to the parent of a collapsed/childless node |
-| Right arrow | Expand focused node if collapsed, move to first child if already expanded, or do nothing if end node |
+For the PatternFly React library:
 
-**Screen reader users** should be able to interact with the tree view using their mouse or keyboard. Nodes of a tree view should have descriptive text labels whenever possible. If a tree view node only contains a graphic (such as an icon), it should have alternative text: `aria-label` is the most common way to add this. A tree view should be identified as such when the screen reader first focuses on it. When users navigate to a new node, indicate its branch level , number of child nodes, and number of sibling nodes.
+- Enure the root TreeView component's `isNested` prop is false, and that any nested TreeView components' `isNested` prop are true.
 
-## To make tree view accessible:
-- For the search variant, use an aria-label describing what the user can search. For example, `searchProps={{ id: 'input-search', name: 'search-input', 'aria-label': 'Search input example' }}`.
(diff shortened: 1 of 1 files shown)


jessiehuff:
Eric and I were talking through this, and we agree that there are a number of improvements that could be made to tree view. 

In core and react:
- Mention `aria-multiselectable` is needed when the tree view is intended to have multiple items selected. So that'd be more a docs update in Core, and a new prop added in React (isMultiSelectable)
- Add `aria-selected` to tree view items when selection made (we should probably look into whether `aria-checked` should be applied instead of `aria-selected` for checkbox tree views)
- Add aria-labeling to gives context to what the tree view is of/for (either `aria-labelledby` to a heading or `aria-label`)

Docs updates:
- Recommend that search is present when 7 or more tree nodes are present (we'd like to have future discussions though about whether enforcing a search when there are enough nodes makes sense in the future)

packages/documentation-site/patternfly-docs/content/accessibility/tree-view/tree-view.md
```diff
@@ -3,32 +3,89 @@ id: Tree view
...
+| `role="treeitem"` | `li.pf-v5-c-tree-view__list-item` | Identifies the element as a group of nested tree view items. **Required** `ul` elements nested within the root `ul[role="tree"]`. |
+| `tabindex="-1"` | `button.pf-v5-c-tree-view__node, button.pf-v5-c-tree-view__node-toggle, button.pf-v5-c-tree-view__node-text, .pf-v5-c-tree-view__node-check > input[type="checkbox"], .pf-v5-c-tree-view__action > button` | Makes the element focusable without including it in the tab sequence of the page. **Required** on all tree items, toggles, checkboxes, and action buttons, except for the element that is currently or was last focused within the tree view. This must be updated as a user navigates through a tree view. |
+| `tabindex="0"` | `button.pf-v5-c-tree-view__node, button.pf-v5-c-tree-view__node-toggle, button.pf-v5-c-tree-view__node-text, .pf-v5-c-tree-view__node-check > input[type="checkbox"], .pf-v5-c-tree-view__action > button` | Includes the element in the tab sequence. **Required** on the tree item, toggle, checkbox, or action button that is currently or was last focused within the tree view. This must be updated as a user navigates through a tree view. |
```
thatblindgeye, line 84: The core docs currently (but once my Core PR goes in, previously) mentioned the tabindex for the `li[role="treeitem"]` element. This matches what WAI examples are doing, like their [file tree view example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1a/).

However, what the React implementation is doing is updating the tabindex on a button element that is inside these li elements:

image

Or at least that's the intent when the examples get more complicated (separate expansion/selection, checkboxes, and actions).

I included all of the elements that could receive focus in a tree view above, but since the examples in React don't currently work that way (I opened an issue regarding this, it may be a bit complicated to decide how we want to go with it), I could just mention the tree view nodes here for now, e.g. the `button.pf-v5-c-tree-view__node` elements in the most basic examples.

That said, do we want to also update core/react implementations to omit the tabindex on the li elements?

packages/documentation-site/patternfly-docs/content/accessibility/tree-view/tree-view.md
```diff
@@ -3,32 +3,89 @@ id: Tree view
...
+  </ListItem>
+  <ListItem>
+    <Checkbox id="treeview-a11y-checkbox-4" label={<span>Within the context of the tree view, each tree view item can be navigated to and interacted with via keyboard. <kbd>Up Arrow</kbd> and <kbd>Down Arrow</kbd> will navigate to the next or previous item without expanding or collapsing any, respectively. <kbd>Left Arrow</kbd> will collapse an expanded item or navigate to its parent. <kbd>Right Arrow</kbd> will expand a collapsed item or navigate to its first child. <kbd>Enter</kbd> and <kbd>Space</kbd> will trigger a default action, typically selecting an item and/or expanding/collapsing it.</span>} />
+  </ListItem>
+  <ListItem>
+    <Checkbox id="treeview-a11y-checkbox-5" label={<span>When focus is outside of the tree view. <kbd>Tab</kbd> places focus on the first interactive item in the tree view or the last interactive element to have focus within the tree view. When focus is inside of the tree view, <kbd>Tab</kbd> places focus on the first focusable element outside of the tree view.</span>} />
+  </ListItem>
```
thatblindgeye, line 44: This is made more true in the React PR I have up, but the more complicated examples may not work this way. It's trickier because a basic tree view should not have its items tabbable except when first entering the tree view, but things like separate selection/expansion, checkboxes, and action items complicate things.

Tabbing may be required in some of these instances because by default the Right and Left arrows have 2 roles: expanding/collapsing a parent node, and navigating to the first child/the parent when on a non-expandable or collapsed node. So using Right/Left arrow to get to a separate selection button, a checkbox, or an action item becomes difficult without omitting that default behavior, and if we do that it might be confusing that this behavior works on some tree views but not others.

jessiehuff (approved):
Thanks for all of your work on this!!! 😄

PR #3925, opened 2024-03-20
chore(): monaco-editor dependency

kmcfaul:
After bumping the docs framework version in react for the templates package, I'm running into an issue building the code-editor package. I think we do still need `monaco-editor` for it to build without manual dependency installation.


kmcfaul:
@nicolethoen  Or should `monaco-editor` be a dependency of our `react-code-editor` package instead?

nicolethoen (approved):

kmcfaul:
Closed because monaco-editor will be added as dependency of the code editor package.

PR #3934, opened 2024-03-24
chore: simplify Babel configuration

jonkoops:
Simplifies the Babel configuration by removing transformations that are no longer needed, and replacing them with relevant presets.

This change is needed to land https://github.com/patternfly/patternfly-react/pull/10189


packages/documentation-framework/package.json
jonkoops, line 1: I've removed all the individual Babel transformations, since `@babel/preset-env` will already handle these automatically. I've also replaced `@babel/plugin-transform-react-jsx` with `@babel/preset-react`, which can do more in-depth optimization.

packages/documentation-framework/scripts/webpack/webpack.base.config.js
```diff
@@ -55,20 +55,12 @@ module.exports = (_env, argv) => {
             options: {
               cacheDirectory: '.cache/babel',
               cacheCompression: false,
-              presets: [['@babel/preset-env', {
-                loose: true,
-                corejs: 3,
-                useBuiltIns: 'entry',
```
jonkoops, line 61: I've removed the `core-js` auto-polyfilling and related settings here. The reason for that is with `useBuiltIns` set to `entry` it is required to add an explicit import to `core-js` in the application code ([see documentation](https://babeljs.io/docs/babel-preset-env#usebuiltins-entry)).

Since the aforementioned import never existed in the application code this essentially means that `core-js` polyfills were also never included. So I have opted to plain out disable it to make this more explicit.

packages/documentation-framework/scripts/webpack/webpack.base.config.js
```diff
@@ -55,20 +55,12 @@ module.exports = (_env, argv) => {
...
-                exclude: ['transform-regenerator', 'transform-async-to-generator'],
```
jonkoops, line 62: I've removed these transforms, as they are now implicitly no longer included by updating the `targets` field, as well as updating the [browserlist database](https://github.com/browserslist/update-db?tab=readme-ov-file#update-browserslist-db) (change included in updated lockfile).

packages/documentation-framework/scripts/webpack/webpack.base.config.js
```diff
@@ -55,20 +55,12 @@ module.exports = (_env, argv) => {
...
-                modules: false,
```
jonkoops, line 63: I've removed the `modules` field here, as this should be left to `auto` (the default) when using it in combination with a bundler such as WebPack + `babel-loader`, which is the case here ([see documentation](https://babeljs.io/docs/babel-preset-env#modules-auto)).

packages/documentation-framework/scripts/webpack/webpack.base.config.js
```diff
@@ -55,20 +55,12 @@ module.exports = (_env, argv) => {
...
-                '@babel/plugin-transform-react-jsx',
-                '@babel/plugin-proposal-class-properties',
-                '@babel/plugin-proposal-optional-chaining',
-                ["@babel/plugin-proposal-private-methods", { "loose": false }],
-                ["@babel/plugin-proposal-private-property-in-object", { "loose": false }]
+              presets: [
+                '@babel/preset-react',
+                ['@babel/preset-env', {
+                  loose: true,
+                  targets: '> 1%, not dead'
```
jonkoops, line 62: I've narrowed the support target, as this [was matching](https://browsersl.ist/#q=%3E+0.25%25%2C+not+dead) a lot of unsupported browsers. The new target still has [wide support](https://browsersl.ist/#q=%3E+1%25%2C+not+dead) for various browsers, but prevents excessive transpilation.

packages/documentation-framework/scripts/webpack/webpack.base.config.js
```diff
@@ -55,20 +55,12 @@ module.exports = (_env, argv) => {
...
+                }],
               ],
```
jonkoops, line 64: All these plugins are already supported by `@babel/preset-env`, so there is no need to specify them individually.

nicolethoen (approved):

wise-king-sullyman (approved):
🚀

PR #3947, opened 2024-03-27
feat(Templates): add to org site

kmcfaul:
Adding the templates to the org MD.

@nicolethoen Do I need to add the package as a dependency somewhere, or is the templates package not published yet? I'm not seeing how we got the drag and drop package found.


nicolethoen:
The templates package will get to documentation-site via its react-docs dependency. So that'll have to be updated to the latest react-docs version in order to find the templates documentation.

That said, I don't want this to merge before the alpha release is live, so that we aren't exposing the 5.3 release ahead of promoting the 5.3 packages, and I think the alpha is going to be released first. This PR can be all primed and ready, though :)

PR #3961, opened 2024-04-11
feat(docs): update beta button style

### LINKED ISSUE #3859: Beta label styling issues
Beta labels are in plain buttons, and at least the labels next to a beta example have plain button hover styles that shouldn't show. Do those need to be in buttons at all? If so, we have a "no padding" plain button that would probably be more appropriate.

Screenshot 2024-01-12 at 9 16 33 AM

kmcfaul:
Closes #3859.

### DIFF
--- a/packages/documentation-framework/templates/mdx.js
+++ b/packages/documentation-framework/templates/mdx.js
@@ -270,7 +270,7 @@ export const MDXTemplate = ({
                   {isDeprecated && (
                     <FlexItem spacer={{ default: 'spacerSm' }}>
                       <Tooltip content="Deprecated components are available for use but are no longer being maintained or enhanced.">
-                        <Button isInline component="span" variant="plain" hasNoPadding>
+                        <Button variant="plain" hasNoPadding>
                           <Label color="grey">Deprecated</Label>
                         </Button>
                       </Tooltip>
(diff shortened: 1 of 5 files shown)


jessiehuff (approved):

packages/documentation-framework/templates/mdx.js
```diff
@@ -270,7 +270,7 @@ export const MDXTemplate = ({
                   {isDeprecated && (
                     <FlexItem spacer={{ default: 'spacerSm' }}>
                       <Tooltip content="Deprecated components are available for use but are no longer being maintained or enhanced.">
-                        <Button isInline component="span" variant="link">
+                        <Button isInline component="span" variant="link" hasNoPadding>
```
mcoker, line 273: For the 3 updates in this file, these are inline link span buttons and `hasNoPadding` only works on plain buttons. That said, I think these should all be plain buttons? Currently the label is technically taller than the button because of the display types between the span (display inline) and a label (display inline-flex), and that would be fixed by using a plain button.

Screenshot 2024-04-23 at 8 41 54 PM

mcoker (approved):
LGTM, just one small change unless there is a reason you think we should leave it as is.

mcoker (approved):
👍

wise-king-sullyman (approved):
🚀

Issue #3979, opened 2024-04-23
A11y report bugs


evwilkin:
Epic tracking the issues found in 4/23/24 a11y report - https://pf-a11y_4-23-24.surge.sh/

```[tasklist]
### Testing-your-accessibility
- [ ] <li> elements must be contained in a <ul> or <ol>
- [ ] li elements must be contained in a ul or
```

```[tasklist]
### Panel/HTML
- [ ] Scrollable region must have keyboard access
```

```[tasklist]
### Sidebar/html
- [ ] Scrollable region must have keyboard access
```

```[tasklist]
### components/table
- [ ] Scrollable region must have keyboard access
```

```[tasklist]
### components/table/html
- [ ] Scrollable region must have keyboard access
```

```[tasklist]
### extensions/component-groups/shortcut-grid
- [ ] id attribute value must be unique
```

```[tasklist]
### training/html-fundamentals-training
- [ ] IDs used in ARIA and labels must be unique
```

nicolethoen:
replaced by: https://github.com/patternfly/patternfly-org/issues/4374

PR #3980, opened 2024-04-23
chore(a11y): updated a11y scorecard

### LINKED ISSUE #3744: Run a11y report for new Org site & publish results

evwilkin:
Closes #3744 

This PR updates the a11y scorecard with a new report link from 4/23/24 and links to a new epic (#3979) with 24 new bugs.

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/developer-resources/a11y-scores.js
+++ b/packages/documentation-site/patternfly-docs/content/developer-resources/a11y-scores.js
@@ -2,21 +2,24 @@ import React from 'react';
 
 export const a11yScores = [
   {
-    'Nov 21, 2022': [
+    'April 23, 2024': [
       {
-        criteria: 'Preliminary accessibility report generated using axe-core based tooling.',
+        criteria:
+          'Preliminary accessibility report generated using axe-core based tooling.',
         status: '✅',
         notes: (
           <>
-            <a href="https://pf-a11y_5-16-22.surge.sh/">Initial report (May 16th, 2022)</a>
+            <a href="https://pf-a11y_5-16-22.surge.sh/">
+              Initial report (May 16th, 2022)
+            </a>
             <p>1066 violations found</p>
           </>
-        )
+        ),
       },
       {
         criteria: 'Manual keyboard and screen reader testing conducted.',
         status: '--',
-        notes: ''
+        notes: '',
       },
       {
         criteria: `Accessibility violations in product itemized and prioritized. Notes should indicate how violations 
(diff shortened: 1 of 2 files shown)


packages/documentation-site/patternfly-docs/content/developer-resources/a11y-scores.js
```diff
@@ -1,77 +1,198 @@
...
+      {
+        criteria:
+          'Preliminary accessibility report generated using axe-core based tooling.',
+        status: '✅',
+        notes: (
+          <>
+            <a href="https://pf-a11y_5-16-22.surge.sh/">
+              Initial report (May 16th, 2022)
+            </a>
+            <p>1066 violations found</p>
```
thatblindgeye, line 15: Wondering if this should be the information for the day this new report was ran for April 23 instead. I feel like the context for each new report becomes any a11y issues found as part of that report, rather than the very first report we ever ran.

So the baseline becomes the 24 violations found in this new report, and the `Critical and major accessibility violations identified in item #3 resolved (%).` criteria item basically becomes "N/A" until we go through to resolve issues.

Mainly wondering because as time goes on, the initial 1066 violations become less pertinent to newer reports. Curious what @jessiehuff @nicolethoen think, though.

evwilkin: Good point, that "initial" data isn't too relevant anymore.  What about simply "prior" report?  I think either we:
- List the last report that was done (in this case, our live website shows Nov 21, 2022)
- Or we wait to update the scorecard until after work has been done on what we find.  Ex: I just found 24 violations, how many have we fixed yet?
  - In this case, either it would always show "N/A" as Eric mentioned until we address & merge PRs to make the fixes (and put up another new scorecard changing from N/A to 0), or we'd wait to publish the new scorecard at all until the work is done - when do we call it done, do we have to get to 0 issues before publishing a new scorecard?

thatblindgeye: That is also a good point. Maybe we update the "violations resolved" field whenever a new report is ran, rather than when we get to X issues remaining? If we plan to run and update this report more regularly that might be easier, say something like:

1. Q1: run initial report. "Violations resolved" is N/A
2. Q2: run a new report, and update the "Violations resolved" for the Q1 report
3. Q3: run a new report, and update the "Violations resolved" for the Q2 report
4. ...etc

That would paint the picture of how many a11y issues we're introducing/resolving between reports, but also keep us a bit accountable since we'd have a schedule of when we need to update that "violations resolved" field. Though that would mean altering that "Violations resolved" field since currently it's expecting to be filled out based on the current report.

Otherwise we'd probably have to do something like run the report Q1 start, but wait to actually upload it until Q1 end with whatever number of violations that have been resolved.

evwilkin: Thanks for the feedback!  I've updated the wording based on this conversation to remove "preliminary" from "preliminary accessibility report", and replace "initial report" with "Q2 2024" which can be updated with each new report.

jessiehuff (approved):
I think this looks good with the changes!

thatblindgeye (approved):

PR #3989, opened 2024-04-29
fix(typescript): bump version to fix dependency issue and topology build

### LINKED ISSUE #3988: Bug - incompatibility between typedoc version and typescript version
**Describe the problem**
A clear and concise description of the problem. Which pages are affected?
Typedoc 0.23 has [dropped support](https://github.com/TypeStrong/typedoc/issues/1945#issuecomment-1143061412) for any Typescript version < 4.6.

See installation requirements [here](https://typedoc.org/guides/installation/#requirements).

Topology also lists Typedoc as a dependency and the [build fails](https://github.com/patternfly/react-topology/issues/183) since we are pulling in these conflicting versions from the docs framework.

We'll have to bump the Typescript version up to at least 4.6 (topology and [other extensions](https://github.com/patternfly/react-console/blob/main/package.json#L54) use 4.7.4).

**Expected behavior**
A clear and concise description of the expected behavior.
Topology build succeeds with latest v6 alphas.

jenny-s51:
Closes #3988

### DIFF
--- a/packages/documentation-framework/package.json
+++ b/packages/documentation-framework/package.json
@@ -60,7 +60,7 @@
     "style-to-object": "0.3.0",
     "to-vfile": "6.1.0",
     "typedoc": "0.23.0",
-    "typescript": "4.3.5",
+    "typescript": "4.7.4",
     "unified": "9.1.0",
     "unist-util-remove": "2.0.0",
     "unist-util-visit": "2.0.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -41,6 +41,11 @@
     "@babel/highlight" "^7.24.2"
     picocolors "^1.0.0"
 
+"@babel/compat-data@^7.17.10", "@babel/compat-data@^7.17.7", "@babel/compat-data@^7.20.5":
+  version "7.24.4"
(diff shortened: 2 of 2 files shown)


packages/documentation-framework/package.json
```diff
@@ -60,7 +60,7 @@
     "style-to-object": "0.3.0",
     "to-vfile": "6.1.0",
     "typedoc": "0.23.0",
-    "typescript": "4.3.5",
+    "typescript": "4.7.4",
```
dlabaj, line 63: This update looks fine to me.  However long term I wondering if most of the dependencies should be moved to dev and peer.  @wise-king-sullyman Do you see any issues with moving these dependencies in the future?

dlabaj (approved):
LGTM

packages/documentation-framework/package.json
wise-king-sullyman: Yeah I'm not sure how many should be moved, but it is probably something that should be investigated.

wise-king-sullyman (approved):

jessiehuff (approved):

jenny-s51 (commented):
There seems to be a parsing issue that's causing the build error... investigating...

jenny-s51:
Closing in favor of #4004

Issue #3945, opened 2024-03-27
Identify & resolve any custom styles


kmcfaul:
We should be able to remove all the CSS copied from core's pf-v6-c-content: https://github.com/patternfly/patternfly-org/blob/v6/packages/documentation-framework/templates/mdx.css#L90.

Looks like we can remove custom styles for the sidebar nav, which should resolve this issue https://github.com/patternfly/patternfly-org/issues/3953

mcoker:
Once https://github.com/patternfly/patternfly/issues/5966 is complete, assuming those styles are pretty much a copy of the core `pf-[v]-c-content` (aka, react `<Text>` and `<TextContent>`) component, ideally, we should be able to yank out those styles and swap `.ws-[p/dl/hr/etc]` with classes from the content/text/textcontent component - eg, `.pf-v6-c-content--[p/dl/hr/etc]`.

PR #3996, opened 2024-05-03
fix(styles): Remove and replace global styles

wise-king-sullyman:
Towards #3945

This PR updates global styles in the quickest way I could so that their removal in core didn't break things and prevent further progress on the style updates or block us from pulling in the latest alphas.

### DIFF
--- a/packages/documentation-framework/components/autoLinkHeader/autoLinkHeader.css
+++ b/packages/documentation-framework/components/autoLinkHeader/autoLinkHeader.css
@@ -3,8 +3,8 @@
 }
 
 .ws-heading-anchor {
-  color: var(--pf-v6-global--Color--100);
-  transform: translate(calc(-100% - var(--pf-v6-global--spacer--xs)), -50%);
+/*   color: var(--pf-v6-global--Color--100); */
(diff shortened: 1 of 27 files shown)


jessiehuff (approved):
Not seeing any huge issues :)

srambach:
All of these can be follow ups, just documenting -

- Cookie banner/message needs to have a background.
- The blue links in the black section at the bottom of the page aren't accessible. Are we keeping the black background I wonder?
- Get started->Community page - the vertical spacing of sections, headers, etc, is missing
- I don't see code snippet highlights on react component pages
- Code snippet highlights run into each other (HTML page) Do we want a little more space or no?
- Training pages, the cards are messed up https://patternfly-org-pr-3996-site.surge.sh/training/html It looks like instead of a label, we have a tiny card?
- Utilities - Dark theme dotted outlines on some examples are too light, e.g. https://patternfly-org-pr-3996-site.surge.sh/utility-classes/flex They seem to be good on the layouts.

nicolethoen (approved):

mcoker (approved):
🐢🍕

PR #4004, opened 2024-05-06
fix(build): bump ts version and fix docs build for v6

jenny-s51:
Unblocks https://github.com/patternfly/react-topology/issues/183. Towards #3988 

This change comments out the `visit` function which was imported by `unist-util-visit`, which was causing an error in the build step following the upgrade to Typescript 4.7.4.

 cc @nicolethoen per our debugging session this afternoon


jenny-s51:
Opened a follow up issue: #4005

nicolethoen (approved):

PR #4006, opened 2024-05-06
add comparison table to table guidelines

### LINKED ISSUE #3963: Add comparison table design guidelines
We have no guidelines around what a comparison table should look like in PF.
Need to add guidelines that cover the following things:
- Where the value vs parameter should be placed
- Whether to use vertical divider lines between items being compared
- Whether to use grey background color to distinguish parameter values / sticky column from rest of table, and if so, which shade

mmenestr:
added these guidelines https://docs.google.com/document/d/1Z6LyyqYTxKxCtYwRVywttHjL7v15AFt7FyAXJB8zNK0/edit to table guidelines

closes #3963


jpuzz0 (approved):

packages/documentation-site/patternfly-docs/content/design-guidelines/components/table/table.md
```diff
@@ -118,6 +118,32 @@ The actionable table provides checkboxes or radio buttons that enable users to s
 #### When not to use
 * Users can not take any actions on table items/rows.
 
+### Comparison table
```
edonehoo, line 121: should we also add this as a list item on line 77?

edonehoo (approved):
one small comment, but otherwise looks good!

packages/documentation-site/patternfly-docs/content/design-guidelines/components/table/table.md
mmenestr: great catch!! Just updated

edonehoo (approved):
🤘❗️

jessiehuff (approved):

Issue #4009, opened 2024-05-07
[Documentation framework] - "page loaded" class doesn't always fire


mcoker:
We added a `.page-loaded` class when all of the page assets have loaded as part of https://github.com/patternfly/patternfly-org/pull/3789

Looks like the class isn't always being added to the page. According to @evwilkin 

> well i can’t exactly say why it happens in incognito and not elsewhere, but i believe the TLDR is the window.load event listening is being added to the page after it’s already loaded.  The listener is always there, but only triggers a “load” outside of Chrome incognito
> hard to google what causes that and why it’s working in Chromium but not Chrome for me, but luckily I think there’s an easy fix - before we add the event listener we check document.readyState  - if it’s “complete” we go ahead and add the class, if not we add the listener as we do now

And Evan said this was working for him

```
if (isFullscreenPreview) {
    isFullscreen = false;
    if (document?.readyState === 'complete') {
      document.body.classList.add('page-loaded');
    } else {
      window.addEventListener('load', () => {
        //append a class to the document body to indicate to screenshot/automated visual regression tools that the page has loaded
        document.body.classList.add('page-loaded');
      });
    }
  }
```

Once that is working, we'll want to add that to the core repo's backstop config as the `readySelector`, outlined in https://github.com/patternfly/patternfly/issues/6602

PR #4012, opened 2024-05-09
chore(docs): fix page-loaded class

evwilkin:
Towards #4009 

This PR adds extra logic to check if page has already finished loading (and if so, adds the page-loaded class) before adding window.load event listener.  
- This fixes the race case condition where the page-loaded class is sometimes not added in incognito windows where the listener is never triggered because it is added after the page has already loaded.

### DIFF
--- a/packages/documentation-framework/components/example/example.js
+++ b/packages/documentation-framework/components/example/example.js
@@ -1,6 +1,15 @@
 import React, { useContext } from 'react';
 import { useLocation } from '@reach/router';
-import { Button, CodeBlock, Flex, CodeBlockCode, debounce, Label, Switch, Tooltip } from '@patternfly/react-core';
+import {
+  Button,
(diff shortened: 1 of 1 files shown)


mcoker (dismissed):
LGTM! I did pull it down and test with a big image on an example and the event is firing before the image has fully loaded, but I also added this to the `<head>` and as the last thing in `<body>` and got the same results, so I think the change in this PR is good, though we'll have to see if we still have any issues with images loading

```
    <script>
      window.onload = function() {
        console.log("window.load fired");
      };
    </script>
```

wise-king-sullyman (dismissed):
I haven't been able to replicate the issue here but this change looks like it should be a great fix 🚀

mcoker (approved):
LGTM! Verified on a handful of examples it's firing and looks to be at the right time

PR #4013, opened 2024-05-09
chore(docs): fix page-loaded class

evwilkin:
Towards #4009 

This PR wraps the window.load listener in a check to see if the page has already completed loading before adding that listener (and if so, adds the page-loaded class).
- This fixes the race condition where that event listener sometimes does not fire in incognito when it is added after the page has loaded.
- Mirrors PR #4012 going into `main` branch.

### DIFF
--- a/packages/documentation-framework/components/example/example.js
+++ b/packages/documentation-framework/components/example/example.js
@@ -1,6 +1,15 @@
 import React, { useContext } from 'react';
 import { useLocation } from '@reach/router';
-import { Button, CodeBlock, Flex, CodeBlockCode, debounce, Label, Switch, Tooltip } from '@patternfly/react-core';
+import {
+  Button,
+  CodeBlock,
+  Flex,
+  CodeBlockCode,
+  debounce,
+  Label,
+  Switch,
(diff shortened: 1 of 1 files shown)


mcoker (approved):

wise-king-sullyman (approved):

evwilkin:
@mcoker @wise-king-sullyman FYI pushed one additional change, to wrap this code in `useEffect` as an additional check to delay execution until the Example has loaded into the DOM

mcoker (approved):
Manually applied to my local dev server (until docs framework full page examples are fixed) - LGTM!

wise-king-sullyman (approved):
I remember having some funky behavior when I tried using a useEffect for this in the initial approach, but I also didn't have the readyState check which I'm guessing resolves that issue since if it _isn't_ done doing the initial paint when the effect runs it should add the listener, and when it is it should just load the class.

I do wonder if there might be an edge case here if the effect starts running before painting is complete, but not in time for the listener to catch the event, but we can always cross that bridge if/when we get there.

PR #4015, opened 2024-05-09
fix(docs-framework): adjust back to top z-index

mcoker:
fixes https://github.com/patternfly/patternfly-org/issues/4014

I think back to top should overlay everything, similar to skip-to-content. currently when it shows up when the jump links on the right are visible, the back to top button is visible, but clicking on it triggers a click on one of the jump links.

### DIFF
--- a/packages/documentation-framework/templates/mdx.css
+++ b/packages/documentation-framework/templates/mdx.css
@@ -308,7 +308,6 @@
   max-width: 832px;
 }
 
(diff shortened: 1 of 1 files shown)


wise-king-sullyman (approved):

srambach (approved):
LGTM

jessiehuff (approved):

PR #4024, opened 2024-05-15
fix: Incorrectly formatted link

Venefilyn:
There was a link that was incorrectly formatted and let to a 404 page


jessiehuff (approved):
Thanks for catching this! :)

PR #4049, opened 2024-05-30
feat(docs-framework): Update AutoLinkHeader to use the Text component

### LINKED ISSUE #4035: Update AutoLinkHeader to use the Text component rather than Title
Followup from #4011 

See this comment for more info
https://github.com/patternfly/patternfly-org/pull/4011#discussion_r1604128254

Blocked by #4034

wise-king-sullyman:
Closes #4035


jessiehuff (approved):

mcoker (changes requested):
LGTM though the anchor link disappeared with removing `position: relative` from `.ws-heading`. Also if we wanna make the anchor icon the default icon color, the token for that is `var(--pf-t--global--icon--color--regular)`. Looks like the existing style is commented out, so the icon is inheriting its color from the parent `<a>` which is set to our link color https://github.com/patternfly/patternfly-org/blob/4b783773dc5f30f16820cfd33433aa6a8cfd4546/packages/documentation-framework/components/autoLinkHeader/autoLinkHeader.css#L2

Left another comment about the flex layout used around the autolinkheader and how that will keep any heading in it from applying margin from the content/text component since it's the only-child. Probably out of scope for this PR to do anything about, but probably worth looking at somewhere in this update of the org styles? At least the content heavy pages that likely will not have metadata could probably benefit from not having the parent flex layout that will enable the heading margin/spacing to apply.

FWIW Here's the [about ux writing page](https://patternfly-org-pr-4049-site.surge.sh/ux-writing/about-ux-writing/) if I move all of the headings out of the flex layouts, remove a random "undefined" string showing up in the classlist for the paragraphs and lists and things, and change the page title from a 4xl title to the content/text component's `<h1>` (that last one I'm assuming we don't want to do, the title is way too small)

Screenshot 2024-05-30 at 5 13 20 PM

And here's the page as it is in the PR for reference

Screenshot 2024-05-30 at 5 13 36 PM

nicolethoen (approved):

mcoker (approved):
LGTM! One thing we should follow up with is this little gap between the anchor and heading text, if you move your mouse slowly from the heading text to the anchor, the anchor will disappear

Screenshot 2024-06-17 at 11 12 08 AM

Issue #3794, opened 2023-11-01
Bug - address vulnerabilities in documentation-framework


nicolethoen:
### critical:
Introduced through: sharp@0.30.6
Fixed in: sharp@0.32.6

### High:
Introduced through: @patternfly/documentation-framework@5.3.0 › remark-parse@8.0.3 › trim@0.0.1
Fix: Upgrade to remark-parse@9.0.0

Introduced through: codesandbox@2.2.0
Fixed in: axios@1.6.0

Introduced through: codesandbox@2.2.0
Fixed in: ansi-regex@3.0.1, @4.1.1, @5.0.1, @6.0.1

Introduced through: @babel/core@7.18.2, @babel/preset-env@7.18.2 and others
Fixed in: semver@5.7.2, @6.3.1, @7.5.2

Introduced through: codesandbox@2.2.0
Fixed in: ssri@6.0.2, @7.1.1, @8.0.1

wise-king-sullyman:
Before addressing we should probably validate how much of an issue this really is given that we don't ship org code to consumers.

The first step for trying to resolve this should be updating deps.

PR #4079, opened 2024-06-06
fix: remove codesandbox integration to fix vulnerabilities

nicolethoen:
progress toward: https://github.com/patternfly/patternfly-org/issues/3794


dlabaj (approved):

dlabaj:
@nicolethoen Looks like there's a yarn conflict.

nicolethoen:
yeah this shouldn't be merged without more conversations... idk if we want to remove our codesandbox integration without an alternative...

PR #4081, opened 2024-06-07
fix: cleaning out example.css

nicolethoen:
Closes: https://github.com/patternfly/patternfly-org/issues/4060

this 'primarily' impacts the component docs/examples pages. The layout for each example on the page is now using Stacks with gutters rather than custom class names.

I had to leave overrides for:

- creating the dark theme switcher at the bottom of full page examples
- modifying the layout of the code editor at the bottom of each example
- beta tags and required stars in the props tables

### DIFF
--- a/packages/documentation-framework/components/example/example.css
+++ b/packages/documentation-framework/components/example/example.css
@@ -2,19 +2,6 @@
   --ws-code-editor--tooltip--MaxWidth: 16ch;
 }
 
-.ws-example {
-  margin-top: var(--pf-t--global--spacer--lg);
-  margin-bottom: var(--pf-t--global--spacer--lg);
-}
-
-.ws-example > .ws-example-header {
-  padding: var(--pf-t--global--spacer--md);
-}
-
-.ws-example-header > .ws-example-heading:not(:last-child) {
-  margin-bottom: var(--pf-t--global--spacer--md);
-}
-
 .ws-code-editor:not(.ws-example-code-expanded) > .pf-v6-c-code-editor__header::before {
   --pf-v6-c-code-editor__header--before--BorderBottomWidth: 0;
 }
@@ -28,18 +15,6 @@
 
 .ws-code-editor-control.pf-v6-c-button {
   --pf-v6-c-button--after--BorderWidth: 0;
-}
-
-.ws-preview {
-  padding: var(--pf-t--global--spacer--md);
-/*   background-color: var(--pf-v6-global--BackgroundColor--100); */
-/*   border-bottom: var(--pf-v6-global--BorderWidth--sm) solid var(--pf-v6-global--BorderColor--300); */
-  transition: width .2s ease-in-out;
-}
-
-.ws-core-c-page.ws-preview > .ws-preview-html,
-.ws-react-c-page.ws-preview > .pf-v6-c-page {
-  overflow: hidden;
 }
 
 .ws-editor {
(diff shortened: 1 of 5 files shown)


packages/documentation-framework/components/example/example.css
```diff
@@ -101,32 +67,32 @@
 }
 
 .pf-v6-c-badge.ws-beta-badge {
```
mcoker, line 69: Do you know where this badge is used? I poked around for a minute on the PR preview site and couldn't find this `ws-beta-badge` class.

packages/documentation-framework/components/example/example.css
```diff
@@ -101,32 +67,32 @@
...
+  color: var(--pf-t--global--text--color--status--danger--default);
 }
 
 .ws-full-page-utils {
   position: fixed;
   right: 0;
   bottom: 0;
   padding: var(--pf-t--global--spacer--lg);
-/*   z-index: var(--pf-v6-global--ZIndex--2xl); */
+  z-index: var(--pf-t--global--z-index--600);
```
mcoker, line 84: whenever possible it's better to use the most semantic version of a token. The ones that end in numbers like `--[100, 200, etc]` are base tokens and we're not protecting those from breaking changes, not like it's a big deal for org though. And for z-index it's a 1:1 match with the previous tokens, just with the different prefix (`pf-t--global` instead of `pf-[v]-global`). 

```suggestion
  z-index: var(--pf-t--global--z-index--2xl);
```

packages/documentation-framework/components/example/example.css
```diff
@@ -101,32 +67,32 @@
...
 
 .ws-full-page-utils::before {
   position: absolute;
   inset: 0;
   content: "";
-/*   background-color: var(--pf-v6-global--BackgroundColor--100); */
+  background-color: var(--pf-t--global--background--color--floating--default);
   opacity: 0.8;
-/*   box-shadow: var(--pf-v6-global--BoxShadow--sm); */
+  box-shadow: var(--pf-t--global--box-shadow--color--sm);
```
mcoker, line 93: box-shadow--color--sm is just a color. Box shadows are another where it's a 1:1 replacement with the new token syntax... with the exception of `pf-v5-global--BoxShadow--inset` and `pf-v5-global--BoxShadow--xl`, which don't exist in v6.

```suggestion
  box-shadow: var(--pf-t--global--box-shadow--color--sm);
```

packages/documentation-framework/components/example/example.css
```diff
@@ -101,32 +67,32 @@
...
 }
 
 .pf-v6-theme-dark .ws-theme-switch-full-page::before {
```
mcoker, line 96: Do you know where this is on the site? Poked around some on the PR preview site and couldn't find it.

mcoker (approved):
LGTM, left some comments but nothing is a blocker

packages/documentation-framework/components/example/example.css
nicolethoen: I believe its used to mark props in prop tables as beta like here:
https://patternfly-org-pr-4081-site.surge.sh/components/menus/select#selectlist

packages/documentation-framework/components/example/example.css
nicolethoen: They are the dark theme switchers in the bottom right corner of the full page examples:
https://patternfly-org-pr-4081-site.surge.sh/components/page/react-demos/sticky-section-group/

mcoker: Hmm... I don't see it in the markup on that page. IIRC I renamed `.ws-theme-switch-full-page` to `.ws-full-page-utils` in https://github.com/patternfly/patternfly-org/commit/fc1f9bacab352cf2db7418ae98b2f3c98cd47302, and probably missed removing this selector.

packages/documentation-framework/components/example/example.css
mcoker: This looks good but I wonder if we could just remove these styles and use a (compact?) label out of the box. We have a blue/outline label, but also if it's more of an "info" label (formerly blue, updated to purple for v6), maybe that would be a good use case.

packages/documentation-framework/components/example/example.css
nicolethoen: I'll take a look

PR #4082, opened 2024-06-07
fix: clean up gdpr banner styles

nicolethoen:
Closes: https://github.com/patternfly/patternfly-org/issues/4062


mcoker (approved):
LGTM, but you could also use a panel with a flex layout - https://codepen.io/mcoker/pen/GRaOXNP

srambach (approved):
👍

PR #4083, opened 2024-06-07
fix: clean up toc/JumpLinks styling

nicolethoen:
Closes: https://github.com/patternfly/patternfly-org/issues/4064

I cannot figure out how to get the mobile jumplinks/table of contents to scroll properly.
I tried to do overflow-y: scroll on the `.ws-toc.pf-m-expanded .pf-v6-c-jump-links__main` but it is clipping the toggle arrow. Very open to suggestions.

### DIFF
--- a/packages/documentation-framework/components/tableOfContents/tableOfContents.css
+++ b/packages/documentation-framework/components/tableOfContents/tableOfContents.css
@@ -1,12 +1,13 @@
 .ws-toc {
   align-self: flex-start;
   position: sticky;
-  width: calc(100% + var(--pf-v6-c-page__main-section--PaddingLeft) + var(--pf-v6-c-page__main-section--PaddingRight));
+  width: calc(100% + var(--pf-v6-c-page__main-section--PaddingInlineStart) + var(--pf-v6-c-page__main-section--PaddingInlineEnd));
+  background-color: var(--pf-t--global--background--color--secondary--default);
   z-index: 201;
   margin-block-start: calc(var(--pf-v6-c-page__main-section--PaddingTop) * -1);
   margin-block-end: var(--pf-t--global--spacer--md);
-  margin-inline-start: calc(var(--pf-v6-c-page__main-section--PaddingLeft) * -1);
-  margin-inline-end: calc(var(--pf-v6-c-page__main-section--PaddingRight) * -2);
+  margin-inline-start: calc(var(--pf-v6-c-page__main-section--PaddingInlineStart) * -1);
+  margin-inline-end: calc(var(--pf-v6-c-page__main-section--PaddingInlineEnd) * -2);
   padding-block-start: var(--pf-t--global--spacer--md);
   padding-block-end: var(--pf-t--global--spacer--md);
   padding-inline-start: var(--pf-t--global--spacer--md);
@@ -28,7 +29,6 @@
   .ws-toc .pf-v6-c-jump-links__header {
     position: sticky;
     top: 0;
-    background-color: var(--pf-t--global--background--color--secondary--default);
     z-index: 2;
   }
 }
--- a/packages/documentation-framework/components/tableOfContents/tableOfContents.css
+++ b/packages/documentation-framework/components/tableOfContents/tableOfContents.css
@@ -64,19 +64,3 @@
   }
 }
 
-/* .ws-toc-link { */
-.ws-toc-item .pf-v6-c-jump-links__link {
-  position: relative;
-  display: block;
-  color: var(--pf-v6-global--Color--300);
-  font-size: var(--pf-t--global--font--size--body--sm);
-  margin: -4px 0;
-}
-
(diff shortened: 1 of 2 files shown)


mcoker:
@nicolethoen the jump links toggle caret is getting cut off because of an odd, yet intentional style in jump links from v5. In reviewing the figma jump links designs, looks like that style was removed and the update to remove it in v6 was overlooked. I'll put up a PR in core to fix it.

packages/documentation-framework/components/tableOfContents/tableOfContents.css
```diff
@@ -2,8 +2,8 @@
   align-self: flex-start;
   position: sticky;
   width: calc(100% + var(--pf-v6-c-page__main-section--PaddingLeft) + var(--pf-v6-c-page__main-section--PaddingRight));
-  background-color: var(--pf-t--global--background--color--primary--default);
-  z-index: 501;
+  background-color: var(--pf-t--global--background--color--200);
+  z-index: var(--pf-t--global--z-index--sm);
```
mcoker, line 6: I would probably try "201" here. We use `--sm` (200) for random little component things that need to stack on top of other things. We use `--md` (300) for sticky sections. `--lg` (400) for backdrop, `--xl` (500) for modal. My guess is 201 might interfere the least with the TOC overlapping examples?

packages/documentation-framework/components/tableOfContents/tableOfContents.css
```diff
@@ -12,23 +12,24 @@
...
 .ws-toc.pf-m-expanded {
-/*   box-shadow: var(--pf-v6-global--BoxShadow--sm-bottom) */
+  box-shadow: var(--pf-t--global--box-shadow--sm--bottom)
 }
 
 /* Mobile jumplinks */
 @media (max-width: 1450px) {
   .ws-toc.pf-m-expanded .pf-v6-c-jump-links__main {
     max-height: 65vh;
+    overflow-y:scroll;
```
mcoker, line 26: `overflow: scroll` will _always_ draw a scrollbar gutter on a box, whether it has overflow/scrollable content or not. `overflow: auto` will conditionally render the gutter if there is scrollable content and a scrollbar.

packages/documentation-framework/components/tableOfContents/tableOfContents.css
```diff
@@ -12,23 +12,24 @@
...
   }
 
   .ws-toc .pf-v6-c-jump-links__header {
     position: sticky;
     top: 0;
-    background-color: var(--pf-t--global--background--color--primary--default);
+    background-color: var(--pf-t--global--background--color--200);
```
mcoker, line 32: If you want this to be the grey background, use `--pf-t--global--background--color--secondary--default`

packages/documentation-framework/components/tableOfContents/tableOfContents.css
```diff
@@ -67,8 +68,8 @@
 .ws-toc-item .pf-v6-c-jump-links__link {
   position: relative;
   display: block;
-/*   color: var(--ws-toc-link--Color, var(--pf-v6-global--Color--300)); */
-/*   font-size: var(--pf-v6-global--FontSize--sm); */
+  color: var(--pf-v6-global--Color--300);
+  font-size: var(--pf-t--global--font--size--body--sm);
```
mcoker, line 72: None of these styles will do anything anymore so you can just remove 'em

packages/documentation-framework/components/tableOfContents/tableOfContents.css
```diff
@@ -67,8 +68,8 @@
...
   margin: -4px 0;
```
mcoker, line 73: You can keep this I reckon, but can we remove it and see if anyone complains?

packages/documentation-framework/components/tableOfContents/tableOfContents.css
```diff
@@ -2,8 +2,8 @@
   align-self: flex-start;
   position: sticky;
   width: calc(100% + var(--pf-v6-c-page__main-section--PaddingLeft) + var(--pf-v6-c-page__main-section--PaddingRight));
-  background-color: var(--pf-t--global--background--color--primary--default);
-  z-index: 501;
+  background-color: var(--pf-t--global--background--color--200);
```
mcoker, line 5: I haven't really dug into the styles too much, but looks like since this applies to `.ws-toc`, you shouldn't need to specify a different background on the jumplinks header here https://github.com/patternfly/patternfly-org/blob/27938f190eff1be7b8fdd9b82b256454155047fe/packages/documentation-framework/components/tableOfContents/tableOfContents.css#L32

Though whether you keep it or not, this background on `.ws-toc` should use the semantic var unless there's a reason you don't think it's the right var?

```suggestion
  background-color: var(--pf-t--global--background--color--secondary--default);
```

mcoker (approved):
Just one little, itty-bitty nit/question.

PR #4084, opened 2024-06-10
fix: clean up mdx.css (after examples.css merges)

nicolethoen:
Closes: https://github.com/patternfly/patternfly-org/issues/4066

### DIFF
--- a/packages/documentation-framework/templates/mdx.js
+++ b/packages/documentation-framework/templates/mdx.js
@@ -91,7 +91,7 @@ const MDXChildTemplate = ({
       {toc.length > 1 && (
         <TableOfContents items={toc} />
       )}
-      <div className={isComponentCodeDocs && 'pf-v6-l-stack pf-m-gutter'} style={{...(source !== 'landing-pages' && {maxWidth: "825px"})}}>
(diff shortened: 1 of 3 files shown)


packages/documentation-framework/templates/mdx.js
```diff
@@ -90,28 +83,27 @@ const MDXChildTemplate = ({
...
   const ChildComponent = () => (
-    <div className="pf-v6-u-display-flex ws-mdx-child-template">
+    <div className={source !== 'landing-pages' ? 'pf-v6-l-flex' : ''}>
       {toc.length > 1 && (
         <TableOfContents items={toc} />
       )}
-      <div>
-        <div className={innerContentWrapperClass()}>
-          {InlineAlerts}
+      <div className={isComponentCodeDocs && 'pf-v6-l-stack pf-m-gutter'} style={{...(source !== 'landing-pages' && {maxWidth: "825px"})}}>
```
mcoker, line 94: I'm assuming you can't use `<Stack>` here easily because of the conditional classes/styles, but it would be great to not have the classes hardcoded 🥸👍

mcoker (approved):
Just one thing to update the mono font token if you can. Otherwise LGTM

PR #4086, opened 2024-06-11
docs(patterns): Adds actions pattern.

### LINKED ISSUE #3986: New guideline - Actions - Patterns
Create a new "Actions" guideline page under Patterns.

The guideline should include topics covered in this doc:
https://docs.google.com/document/d/1TA6uBuNIyA5F1X2rmz0pSUs0vkc7TYT1Rp-_fkQg-xo/edit

edonehoo:
Closes #3986


edonehoo:
@mmenestr can you lmk what you think about this draft when you get a sec! I'm not sure how I feel about the info organization -- it felt a little awkward, but I also didn't really like any of the rearrangements I tried. So curious for thoughts/feedback in general!

mmenestr:
I think the organization is fine if that's all that's gonna be in there! I think this page would also be good for things like "how to deal with actions" ie: what they trigger - do they trigger a modal? Does the action happen immediately? Which I think was an issue that Pankaj was working on which I reassigned to you, if I remember correctly (?)

So for example, "Deletion" actions. This would be a good place to specify that destructive actions should be red and that clicking on a delete button should always pop-up a modal before actually deleting the thing. So in that sense it belongs more in the "Pattern" page because you're explaining a pattern that should be followed of Clicking on a delete button --> Showing a modal --> And then actually deleting the item (as opposed to clicking delete immediately deleting the thing). 

And you can go further by comparing it to a "Remove" action which wouldn't necessarily require a red button (or even a modal potentially) because it's something that can easily be recovered.

mmenestr:
Also - I noticed you used purple little number markers instead of the usual pink. If that's not something you're going to change across every image, I would keep to the classic pink we usually use!

edonehoo:
@mmenestr oo yes tysm for the ideas on how to expand! I did still have that other issue in my backlog and forgot to include it -- will expand to add all that 🤘

edonehoo:
@mmenestr pushed some updates here! I added deletion info from that other issue, and also from looking around on uxd hub (so I may have grabbed some opinionated info). Lmk if the guidance I added rings true from your perspective 👀

packages/documentation-site/patternfly-docs/content/design-guidelines/patterns/actions/actions.md
```diff
@@ -0,0 +1,88 @@
...
+section: patterns
+---
+
+An **action** is any process that a user can trigger by clicking or selecting a linked component. Common actions include adding, deleting, editing, filtering, and submitting, for example. In PatternFly, most actions are shown in the UI as action buttons, action items (in menus), and action cards. These components are incorporated into the design of other components, such as a toolbar with action buttons, a dropdown menu with action items, and so on.
+
+## Elements 
+
+The following image of a card view demonstrates some of the more common uses of actions that you will find in a UI, but does not cover all scenarios. 
+
+![A card view, where the toolbar action buttons, actionable card, ellipsis icon, and action menu item are marked with annotations.](./img/actions-pattern.png)
```
mmenestr, line 12: Would remove the icon in front of the menu item "Delete card" because it's not a common thing to do for menu items - don't want to confuse people!

packages/documentation-site/patternfly-docs/content/design-guidelines/patterns/actions/actions.md
```diff
@@ -0,0 +1,88 @@
...
+1. A **warning icon** beside the title.
+1. A **labeled text input** to ask users to confirm the resource name. 
+    - An error will be shown if the user does not enter a resource name, or if the resource name does not match the resource being deleted.
+1. A **danger button** that submits the modal.
+
+Once a resource has been successfully deleted, show an [inline success alert:](/components/alert#inline-alerts-variants)
+
+![Inline success alert for deletion.](./img/critical-deletion-success.png)
+
+#### Reversible deletion
```
mmenestr, line 51: I wonder if instead of saying "reversible deletion" it would make more sense to say "removal" to align with the way you lay it out above. Deletion = do a modal, and have the red button and maybe a check box if it's really impactful deletion. But removal is more chill, may not require modal, doesn't need a red button etc

mmenestr (changes requested):
Left a couple comments but looks great so far!!

packages/documentation-site/patternfly-docs/content/design-guidelines/patterns/actions/actions.md
```diff
@@ -0,0 +1,91 @@
...
+1. A **labeled text input** to ask users to confirm the resource name. 
+    - An error will be shown if the user does not enter a resource name, or if the resource name does not match the resource being deleted.
+1. A **danger button** that submits the modal.
+
+Once a resource has been successfully deleted, show an [inline success alert:](/components/alert#inline-alerts-variants)
+
+![Inline success alert for deletion.](./img/critical-deletion-success.png)
+
+#### Removal
+When the action of deleting a resource can be reversed, it has a lower impact on the user experience. A confirmation dialog isn't necessary because users can easily retrieve the deleted resource.
```
mmenestr, line 55: Just need to update text in body to match the title - so instead of saying "action of deleting", would write "action of removing"

edonehoo: ah yes, good catch ty!

packages/documentation-site/patternfly-docs/content/design-guidelines/patterns/actions/actions.md
```diff
@@ -0,0 +1,91 @@
...
+1. A **labeled text input** to ask users to confirm the resource name. 
+    - An error will be shown if the user does not enter a resource name, or if the resource name does not match the resource being deleted.
+1. A **danger button** that submits the modal.
+
+Once a resource has been successfully deleted, show an [inline success alert:](/components/alert#inline-alerts-variants)
+
+![Inline success alert for deletion.](./img/critical-deletion-success.png)
+
+#### Removal
+When the action of removing a resource can be reversed, it has a lower impact on the user experience. A confirmation dialog isn't necessary because users can easily retrieve the deleted resource.
```
mmenestr, line 55: there's another "delete" later in the paragraph to edit 😀

edonehoo: oml 😅 let me push that too. Thanks for bearing with me!

mmenestr (approved):

PR #4110, opened 2024-06-25
chore(styles): Remove custom styles for the css vars table

### LINKED ISSUE #4059: Review cssVariables.css styles for replacing custom styles with existing styling options or tokenization
We should evaluate these styles to determine if they're still needed and remove them if not.

If they are still needed we should update the currently commented out global styles with their proper token replacements.

wise-king-sullyman:
Closes #4059


packages/documentation-framework/components/cssVariables/cssVariables.js
```diff
@@ -110,10 +100,9 @@ export class CSSVariables extends React.Component {
                   key={`${rowKey}_2`}
                   className="pf-v6-l-flex pf-m-column pf-m-align-self-center"
                 >
-                  <span
-                    className="ws-color-box"
-                    style={{ backgroundColor: value }}
-                  />
+                  <Icon>
+                    <CircleIcon color={value}/>
```
mcoker, line 104: Looks like the previous circle had a border and box shadow, which we probably want to retain at least one of those for when the circle's color is the same or really close to the background color.

And this is probably something for later, but we're showing the hex value for the color both as the text in the table (`#0066cc`) but also as the color of the icon (`color="#0066cc"`), which isn't accurate in dark theme. Technically this applies to other global vars, too, but mostly colors. For colors, you could just use the component var the table row is referencing for the circle color since component vars are available at the `:root` document scope. 

Screenshot 2024-06-27 at 8 57 51 PM

Screenshot 2024-06-27 at 8 58 10 PM

mcoker (approved):
LGTM! Just one thing about the circle potentially disappearing if it's the same color as the background behind the circle. And another comment for later I imagine.

nicolethoen:
Seeing this, I think it may not be super accessible the way you've removed the hex values. They might need to be there?

nicolethoen (approved):

PR #4155, opened 2024-07-25
docs(homepage): Update homepage for v6

edonehoo:
Closes [ #13](https://github.com/patternfly/patternfly-design/issues/1303)

May be finished post-launch?

### DIFF
--- a/packages/documentation-site/package.json
+++ b/packages/documentation-site/package.json
@@ -17,7 +17,7 @@
     "screenshots": "pf-docs-framework screenshots"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


edonehoo:
Basic/early attempt here. Would love to do this kind of thing:
image

But I can't get the code for split cards to work. 

Footer could use some updates, whether that's in this pr or a followup

edonehoo:
this isn't needed--working with Bekah to put together design mockups for this in Figma, which will need to be implemented eventually by a dev in a new pr

Issue #4073, opened 2024-06-04
Review commented out inline style in ColorSwatch.js


wise-king-sullyman:
In the above referenced file there is an inline style which references a removed global style value and is currently commented out.

That style should be reviewed, it should be removed if it's no longer needed and updated to use a token if it is needed.

nicolethoen:
Closed by #4157

Issue #4067, opened 2024-06-04
Review ColorFamily.css and ColorFamily.js styles for replacing custom styles with existing styling options or tokenization


wise-king-sullyman:
We should evaluate these styles to determine if they're still needed and remove them if not.

Along with the styles in ColorFamily.css, the inline styles in ColorFamily.js which reference removed global styles should also be addressed under this issue.

If they are still needed we should update the currently commented out global styles with their proper token replacements.

nicolethoen:
Per conversation with design:
- remove references to shadows
- color families should not show palette or base tokens, only semantic tokens
- see if we can have two columns under each accordion in the color families. One column for light theme semantic tokens which use these colors, one column for dark theme semantic tokens which use these colors
- pick the three-ish most common colors for each category in the color swatch section

nicolethoen:
Closed by #4157

PR #4157, opened 2024-07-25
fix: update styles on color page

nicolethoen:
Closes https://github.com/patternfly/patternfly-org/issues/4067
Closes https://github.com/patternfly/patternfly-org/issues/4073


packages/documentation-site/patternfly-docs/content/design-guidelines/styles/colors/colors.md
```diff
@@ -174,4 +188,43 @@ Status and state colors are indicators that communicate data and actions to user
...
+To check the contrast between background and text colors, use a <a href="https://color.a11y.com/?wc3" target="_blank" className="pf-m-link">WCAG AA-compliance tool.</a>
+
+<Divider/>
+
+## PatternFly palettes
+Our color palettes are created as a system, designed to reinforce content and workflows. Our palettes are carefully chosen to support effective communication and adapt to the needs of your UI.
+
+Note that bright colors are used sparingly and are reserved for specific interactions, such as communicating status.
+
+Our color palettes are organized into "families" that contain different shades of the same hue. In the following palettes, you can expand each value color or shadow in a family to see more details.
```
edonehoo, line 200:
```suggestion
## Color families 
Our color palettes are organized into "families" that contain different shades of the same hue. In the following families, you can expand each color to see related tokens.
```

Did we decide to keep the color families in addition to linking out to the Red Hat palettes? I think I removed the whole section in my pr, so would update the text to something like this if we are going to keep them

edonehoo:
A few things I noticed: 

1. In dark mode, the text on the color families doesn't update and can't be read in some cases: 
image

2. Also in the families, there are sometimes no tokens listed (which maybe there are just no tokens related to the color? but it looks like there should be some listed): 
image

3. Related to ^ Should we change the label from "CSS variables" to "Tokens" in the expanded family sections?

4. This is small, but the primary background color swatch has a background, even before you hover:
image

packages/documentation-site/patternfly-docs/content/design-guidelines/styles/colors/colors.md
```diff
@@ -174,4 +188,43 @@ Status and state colors are indicators that communicate data and actions to user
 
 [Our goal](/accessibility/patternflys-accessibility/) is for PatternFly to meet [level AA requirements in the Web Content Accessibility Guidelines 2.1](https://www.w3.org/WAI/standards-guidelines/wcag/new-in-21/). To achieve level AA accessibility, your UI contrast ratios must be at or above 4.5:1 for normal text, 3:1 for large text, and 3:1 for graphics and other UI components. Additionally, on hover, link text color should have ample contrast from both the background color and the default state link color.
 
-To check the contrast between background and text colors, use a <a href="https://color.a11y.com/?wc3" target="_blank" className="pf-m-link">WCAG AA-compliance tool.</a>
\ No newline at end of file
+To check the contrast between background and text colors, use a <a href="https://color.a11y.com/?wc3" target="_blank" className="pf-m-link">WCAG AA-compliance tool.</a>
+
+<Divider/>
```
edonehoo, line 193:
```suggestion
```

nicolethoen:
> 4\. This is small, but the primary background color swatch has a background, even before you hover:

That's because without a background color, the color swatch is invisible! 😆 I'm open to other color suggestions!

> Also in the families, there are sometimes no tokens listed (which maybe there are just no tokens related to the color? but it looks like there should be some listed):

I can make a 'no semantic tokens message if there are none?

edonehoo:
> That's because without a background color, the color swatch is invisible!
Oh, duh! 😆 I can't think of a different approach that's ~the golden ticket, so I think we can just stick with this one, just to get the info updated? And then maybe revisit with a designer to see if we want to change it

edonehoo (approved):
I like it 🗣️

packages/documentation-site/patternfly-docs/content/design-guidelines/styles/colors/colors.md
```diff
@@ -174,4 +188,37 @@ Status and state colors are indicators that communicate data and actions to user
...
+    <ColorFamily title="Gray family" family="gray" />
+  </GridItem>
+  <GridItem xl2={4} xl={6} sm={12}>
+    <ColorFamily title="Blue family" family="blue" />
+  </GridItem>
+  <GridItem xl2={4} xl={6} sm={12}>
+    <ColorFamily title="Green family" family="green" />
+  </GridItem>
+  <GridItem xl2={4} xl={6} sm={12}>
+    <ColorFamily title="Cyan family" family="teal" />
```
edonehoo, line 207:
```suggestion
    <ColorFamily title="Teal family" family="teal" />
```
wait, just noticed this. I think we call this teal now

edonehoo (changes requested):

edonehoo (approved):

PR #4166, opened 2024-07-30
fix(docs): removed extra link tags in card title

### LINKED ISSUE #4165: Bug - Homepage card have extra link
The cards on the homepage have an extra `<a>` tag in the `CardTitle` that is forcing the text to have link styling (with underlines).  These tags can be removed in favor of the `selectableActions` prop which is already implemented correctly in the `CardHeader`:

- https://github.com/patternfly/patternfly-org/blob/v6/packages/documentation-site/patternfly-docs/pages/home.js#L102
- https://github.com/patternfly/patternfly-org/blob/v6/packages/documentation-site/patternfly-docs/pages/home.js#L149

[https://staging-v6.patternfly.org/](https://staging-v6.patternfly.org/)
Screenshot 2024-07-30 at 1 38 55 PM

evwilkin:
Closes #4165 

Depends on https://github.com/patternfly/patternfly/pull/6949 & React follow up - will require version bump with those changes.

This PR removes extra `<a>` tags in the actionable cards on the docs homepage which are causing link styling to be applied - these tags are not needed.


kmcfaul (approved):

nicolethoen (commented):
The keyboard interactions on the medium article cards were not ideal before this change, but they are now not working at all without the <a> tags...

evwilkin:
@nicolethoen thanks for the catch - it looks like this is happening with the Actionable cards example too, where you tab into the Cards and navigate between them with arrow keys but moving to a card with a link immediately triggers the link - does this look like a React bug?

- https://patternfly-org-pr-4166-site.surge.sh/components/card#actionable

We can close this PR out or let it sit for now if this is a larger issue that would need to be fixed in React 👀 

@tlabaj FYI

nicolethoen (approved):

PR #4192, opened 2024-08-15
docs(switch): Updates design guidelines.

### LINKED ISSUE #4121: Switch design guidelines need updating
The [design guidelines for Switch](https://www.patternfly.org/components/switch/design-guidelines) look to be either quite outdated and/or incorrect. Images reference 2 labels being used with a Switch, one "form label" and a "switch label". Right now for v5 there's only ever 1 label for a Switch, and in v6 we're updating the API to discourage/prevent dynamic labeling for "on" and "off" states.

This should definitely be done for v6 (we need to also add a11y docs in v6 so whomever would want to work on the design guidelines I could pair up with to have parity between the 2). A v5 update would eb great as well, though may be slightly different due to the API still allowing dynamic labeling.

**Describe the enhancement**

Include screenshots or links to Marvel or other mockups.

**Any other information?**

edonehoo:
Closes #4121 for v6

Can open up a v5-follow-up once this PR is approved

Note: need to check with Design to assess the accuracy of the updated conclusions I reached in these guidelines


packages/documentation-site/patternfly-docs/content/design-guidelines/components/switch/switch.md
```diff
@@ -3,80 +3,83 @@ id: Switch
...
 
+### When to use a switch vs. a radio button 
 
+If the available options represent anything other than “on” or “off”, use a [radio button](/components/forms/radio/design-guidelines) instead of a switch. 
 
-### When to use switches versus checkboxes
+### When to use a switch vs. a checkbox
 
-A switch changes an option and saves it simultaneously, while checkboxes require a separate action to save the selection, such as pressing a “Submit” or “Save” button.
+Use checkboxes when additional action is required to save a user's selections. A switch changes and saves an option simultaneously, while checkboxes do not. For example, the following image requires users to click the **Save** button:
```
thatblindgeye, line 28: Would it be better to convey this as a recommendation, e.g. "A switch should change and save an option simultaneously, while checkboxes should not"?

Not a blocker, just something I noticed when checking out the link/image alt texts on the page

thatblindgeye (approved):
Thank you! ✨ (giving a ☑️ more for a11y perspective than anything)

andrew-ronaldson (approved):
good stuff! 🔱

dlabaj (approved):

Issue #4085, opened 2024-06-10
Scroll position issues in v6 workspace


mcoker:
1. If you scroll down on a component example page, then navigate to another component example page via the vertical nav, the window stays scrolled instead of scrolling back to the top.
2. If you click on a jump link, the section it scrolls you to will be in the middle of the main page area. It should scroll so that the anchored section is at the top of the main page area.

cc @wise-king-sullyman

evwilkin:
Closed by #4197

PR #4197, opened 2024-08-15
fix(docs): updated scroll selector for jumplinks, router links

evwilkin:
Closes https://github.com/patternfly/patternfly-org/issues/4085

Related to https://github.com/patternfly/patternfly-react/issues/10860 - scroll selector needed to be updated due to new wrapper component.

This PR updates the scrollable element passed to JumpLinks in the Table of Contents component, as well as references in the Link component (triggers scroll to top of page after link click), to `.pf-v6-c-page__main-container`, fixing:
- Scrollspy not working (active jumplink in the table of contents doesn't change as you scroll page)
- Scrolling down one page then navigating to another page does not scroll back to the top of new page

### DIFF
--- a/packages/documentation-framework/components/link/link.js
+++ b/packages/documentation-framework/components/link/link.js
@@ -2,16 +2,19 @@ import React from 'react';
 import { Link as ReachLink, navigate } from '@reach/router';
 import { getAsyncComponent } from '../../routes';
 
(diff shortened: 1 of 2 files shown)


mcoker (approved):
woot! Looks great now 🥳

Issue #3912, opened 2024-03-05
Add search bar back for final release


jessiehuff:
Once we are more stable with v6, we should look into adding the search bar back, likely around our beta at the earliest, maybe the full release of v6

PR #4199, opened 2024-08-16
fix(search): add search bar back in

rebeccaalpert:
Fixes https://github.com/patternfly/patternfly-org/issues/3912.

`initialVersion` looks like it was removed in https://github.com/patternfly/patternfly-org/commit/85d90f8422be775caa468d29443b8cf9ca6cd97e; dropping that and some unused imports while I'm in here.


nicolethoen:
This cannot be merged until we are ready to publish the v6 site on patternfly.org because the search is programmed to search the pages on patternfly.org, not the staging site 👍🏻 

But this will work great when we get to that point! :)

nicolethoen (approved):

PR #4191, opened 2024-08-14
feat(docs-framework): Add support for next subpaths in react-charts

wise-king-sullyman:
To enable https://github.com/patternfly/patternfly-react/pull/10854

PR #4245, opened 2024-09-09
feat(docs-framework): Add support for next subpaths in react-charts

wise-king-sullyman:
V6 version of https://github.com/patternfly/patternfly-org/pull/4191


dlabaj (approved):
LGTM

Issue #3410, opened 2023-02-15
Update multiple guidelines to reflect new usage of labels and label groups as chips


mmenestr:
Get rid of chips guidelines, and write new guidelines for labels as chips.

Design issue here: https://github.com/patternfly/patternfly-design/issues/1222

Background: We have decided to fully deprecate chips in favor of labels, which will cause the following changes:
- Getting rid of the chip and chip group components
- Creation a new compact label group component **(will want to document this)**
- Change the standard label group component to match the new compact label group styling **(will want to update images for this)**
- Update all Filter demos and table demos that include chips in them **(will want to update images in Filters to match)**
- Update [Multiple select component](https://www.patternfly.org/v4/components/select#multiple) to include labels instead of chips inside the toggle **(will want to update pictures of this in guidelines too)**

All relevant details and mocks can be found here: https://www.sketch.com/s/d7e287f4-39e3-4c17-a302-2131d8f3ea87

mcarrano:
This is on hold pending completion of https://github.com/patternfly/patternfly/issues/5344 which has been moved to the post-V5 backlog.

mmenestr:
@edonehoo - this seems to be something that is going to happen for PF6. In the staging, chips now have more of a label styling, but I guess we'll have to see exactly how we want to communicate this change in the guidelines. Leaving it on hold for now. Also created a PF6 design guidelines label to flag that this applies to the future state of PF!

edonehoo:
This will move towards completion as images are updated as part of a few of these https://github.com/patternfly/patternfly-org/issues/4235

PR #4258, opened 2024-09-13
docs(labels): Updates design guidelines.

edonehoo:
Works towards #3410


edonehoo:
Just kept getting build errors over and over. Reopened here: https://github.com/patternfly/patternfly-org/pull/4265

Issue #3907, opened 2024-02-28
Document list of breaking changes for penta


mcoker:
Placeholder for working on a list of breaking changes for penta, similar to this table we created for v5 - https://www.patternfly.org/get-started/upgrade/release-notes

The majority of this will likely be the changes from the old global variable system to tokens, and there may not be a 1:1 list of "v5 global var `--pf-v5-some-var` changed to to v6 token `--pf-t-some-var`". So we should discuss whether we need to try and create a v5->v6 map of vars/tokens, or if we can just point folks to the token documentation as a reference for how to update old global vars to tokens.

kmcfaul:
Delete old release notes page on v6 org - this references v5 still.

PR #4259, opened 2024-09-14
feat: add react breaking changes list to release notes

nicolethoen:
Closes https://github.com/patternfly/patternfly-org/issues/3907

view it running on: https://patternfly-org-pr-4259-site.surge.sh/get-started/upgrade/release-notes

### DIFF
--- a/packages/documentation-site/package.json
+++ b/packages/documentation-site/package.json
@@ -17,7 +17,7 @@
     "screenshots": "pf-docs-framework screenshots"
   },
   "dependencies": {
-    "@patternfly/documentation-framework": "6.0.0-alpha.90",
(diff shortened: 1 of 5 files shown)


nicolethoen:
I don't think it's super quick to fix the backticks so they are styled like other markdown pages. I think that could be a follow up if needed.

edonehoo (approved):
I can still see some md table formatting when I view the preview link, but I can also see that you removed it in your pr, so I think it's just the preview showing that. I don't think the backtick/code formatting really matters honestly--especially since it's time consuming to edit--devs will understand it regardless. I can edit in a followup if we do want the visual consistency, though!

PR #4260, opened 2024-09-16
docs(design-foundations): Documents tabular font modifier.

### LINKED ISSUE #4200: Document tabular numbers in "Numerics" section
We have a modifier available to display numbers using `.pf-v6-m-tabular-nums`. This modifier sets `font-variant-numeric: tabular-nums;`. However, it is currently not documented anywhere. It seems like the Numerics page would be a good place.
https://staging-v6.patternfly.org/ux-writing/numerics

It may also be useful to mention this in the design guidelines for Table, as this would be a common place to want tabular numbers.
https://staging-v6.patternfly.org/components/table/design-guidelines

edonehoo:
Closes #4200


edonehoo:
The original issue mentions adding these details to the numerics page, but I added the bulk of the docs to our typography page instead, and linked to it from numerics (and also table). Can swap this around, though! But I thought it may make more sense for font-related details to live with the other typography info 

Also, maybe this would benefit from an image to demonstrate the difference, but I'm struggling to understand what the actual styling difference is (slashed 0 or just a width change. The "code" text style in Figma doesn't work for me, if that's the same idea

edonehoo:
Replacing with https://github.com/patternfly/patternfly-org/pull/4266 because I couldn't get the issues to fix

PR #4266, opened 2024-09-18
docs(design-foundations): Documents tabular font modifier.

edonehoo:
Replaces https://github.com/patternfly/patternfly-org/pull/4260 to fix the many build issues. 

---

The original issue mentions adding these details to the numerics page, but I added the bulk of the docs to our typography page instead, and linked to it from numerics (and also table). Can swap this around, though! But I thought it may make more sense for font-related details to live with the other typography info

Also, maybe this would benefit from an image to demonstrate the difference, but I'm struggling to understand what the actual styling difference is (slashed 0 or just a width change?). The "code" text style in Figma doesn't work for me, if that's the same idea

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/design-guidelines/styles/typography/typography.md
+++ b/packages/documentation-site/patternfly-docs/content/design-guidelines/styles/typography/typography.md
@@ -20,7 +20,7 @@ You can place text content on a page using the content or title component:
 We use 3 fonts in PatternFly:
 - **Red Hat Display:** Used for larger text, such as headings.
 - **Red Hat Text:** Used for smaller text, subheadings, and body text. More readable for long-form text. 
-- **Red Hat Mono:** Used to format text as code.
+- **Red Hat Mono:** A tabular font used to format text as code. [Read more about our use of tabular font styling.](#tabular-font)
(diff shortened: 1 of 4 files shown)


packages/documentation-site/patternfly-docs/content/design-guidelines/styles/typography/typography.md
```diff
@@ -20,7 +20,7 @@ You can place text content on a page using the content or title component:
 We use 3 fonts in PatternFly:
 - **Red Hat Display:** Used for larger text, such as headings.
 - **Red Hat Text:** Used for smaller text, subheadings, and body text. More readable for long-form text. 
-- **Red Hat Mono:** Used to format text as code.
+- **Red Hat Mono:** A tabular font used to format text as code. [Read more about our use of tabular font styling.](#tabular-font)
```
edonehoo, line 23:
```suggestion
- **Red Hat Mono:** A tabular font used to format text as code. [Read more about our use of tabular font styling.](#tabular-font-styling)
```

srambach (approved):
Looks great! 📑

nicolethoen (approved):

PR #4271, opened 2024-09-23
feat(site): fix build for prerelease, use editorial content

### LINKED ISSUE #4169: Use long form text variation
A users has noted that overall, the text size in PatternFly has decreased. This is in line with an overall adjustment to font sizing coming from the larger design community. https://github.com/patternfly/patternfly/issues/6830 gives details and some a11y concerns were addressed in https://github.com/patternfly/patternfly/pull/6846, which introduces rem units to allow text resizing.

We have identified that there are tokens intended for a larger long-form text, but the large size is currently not used outside the text-size utility class.

We need to determine if the org site should adjust itself to use the larger text token or use a utility class; if the content component should be modified to use the larger text for body text, or if a large variant of body text should be created.

kmcfaul:
Closes #4169.


kmcfaul:
@tlabaj The changes in `ColorFamily` (minus the `isEditorial` flag) and `ColorSwatch` were the changes I had to make to get the build to work with the updated prerelease versions

packages/documentation-framework/components/sectionGallery/sectionDataListLayout.js
```diff
@@ -55,10 +55,10 @@ export const SectionDataListLayout = ({
                       </DataListCell>
                     ),
                     <DataListCell width={5} key="text-description">
-                      <Split className={hasListText ? 'pf-v6-u-mb-md' : null}>
+                      <Split className={hasListText ? "pf-v6-u-mb-md" : null}>
                         <SplitItem isFilled>
-                          <Content>
-                            <Content component={ContentVariants.h2}>
+                          <Content isEditorial>
+                            <Content component={ContentVariants.h2} isEditorial>
```
mcoker, line 61: Same comment as before - you should only need `isEditorial` on one of those. If it's on the parent `<Content isEditorial>` then any content component children will get editorial styling without having to use `isEditorial` on the children specifically. Since there is only one child, as long as you want to keep the wrapping `<Content />` (removing it would add a margin to the `h2` if there is an element in the DOM that comes after the `h2`), it wouldn't matter which element has `isEditorial`

packages/documentation-framework/components/sectionGallery/TextSummary.js
```diff
@@ -4,28 +4,28 @@ import { Content } from "@patternfly/react-core";
...
   }
 
   return (
-    <Content>
-      <Content component="p">
+    <Content isEditorial>
+      <Content component="p" isEditorial>
         <SummaryComponent id={id} itemsData={itemsData} />
       </Content>
     </Content>
```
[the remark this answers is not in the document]
kmcfaul: Yeah I couldn't 100% remember if it was that way and erred on the side of caution when I was adding it. I can remove the flag from the nested contents 👍

packages/documentation-site/patternfly-docs/pages/home.js
```diff
@@ -187,6 +187,7 @@ const HomePage = () => (
             <Content
               component={ContentVariants.p}
               className="pf-v6-u-font-size-lg"
+              isEditorial
```
mcoker, line 190: This element has the utility class `className="pf-v6-u-font-size-lg"`, which will override `isEditorial` - want to try removing the utility class font-size and see if the `isEditorial` styling gives either the same or an appropriate font-size?

lboehling:
clicked aroudn through several pages and from what i can tell this looks good from the design side!

mcoker (approved):
👍👍

mcoker (approved):
🐸👍

nicolethoen (approved):

PR #4272, opened 2024-09-23
fix(chokidar): update chokidar watch calls as v4 does not support globs

wise-king-sullyman:
Per changes called for in https://github.com/paulmillr/chokidar/?tab=readme-ov-file#upgrading

We updated to v4 of chokidar because of a security issue, but it seems that broke file watching. This PR should address that.


dlabaj (approved):
LGTM

PR #4293, opened 2024-10-02
docs(release-highlights) Adds last updated date for each release.

### LINKED ISSUE #3864: Add date to release notes
The release notes list each version but don't give any indication of the date they were released, we should include this

edonehoo:
Closes #3864

### DIFF
--- a/packages/documentation-site/package.json
+++ b/packages/documentation-site/package.json
@@ -17,7 +17,7 @@
     "screenshots": "pf-docs-framework screenshots"
   },
   "dependencies": {
-    "@patternfly/documentation-framework": "6.0.0-alpha.108",
+    "@patternfly/documentation-framework": "6.0.0-alpha.109",
     "@patternfly/react-catalog-view-extension": "6.0.0-alpha.7",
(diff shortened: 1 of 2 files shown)


edonehoo:
@evwilkin Hello! Assuming we want to communicate the date release highlights were _published_, I think it's probably safest to just go with month and year, since specific release dates are difficult to nail down ahead of time? wdyt about how this looks?

evwilkin:
Hey @edonehoo that works!  I know we have a date listed in the [`versions.json`](https://github.com/patternfly/patternfly-org/blob/main/packages/documentation-framework/versions.json#L28-L29) file that includes each package version for a release [which we list in the highlights](https://www.patternfly.org/get-started/release-highlights/#patternfly-5.2) so was envisioning we'd reference & automate that somehow, but since we have to manually create the release highlights it doesn't seem too much extra to put the month/year there like you've done 👍  and I agree - sometimes the actual day varies as we go through releasing each package, so using the month works well.

evwilkin (approved):
LGTM

nicolethoen (approved):
🤩

PR #4302, opened 2024-10-04
chore(calendar-month): update v6 images

### LINKED ISSUE #4301: Guideline updates: Calendar month

edonehoo:
Closes #4301


edonehoo:
A few things I noticed, just fyi -- 

In Figma, some of the calendar day states cut off at the container border. I just adjusted my image to get around that, but may need fixing somewhere: 
image

Also, when you select a date in the React example, it is styled like this:
image

Assuming the double line is because it is "selected and in focus" that styling in Figma looks diff?
image

nicolethoen (approved):

mcoker (approved):
LGTM! A couple of notes, but don't have to be done in this PR. From this screenshot:

Screenshot 2024-10-07 at 11 06 24 AM

* The native resolution is pretty big, much bigger than it will display in the browser. The file size isn't huge, but I bet it could be much smaller if we optimized it and dropped the image size. The widest it should display is 825px, so we could use that or maybe 1000px to be safe.
* There is a bunch of white-space built into the left/right of the image, which will keep the "drawn/graphical" portion of the image from being able to scale to the size of its container. This is in a narrow viewport. IMO the image would be better and more readable if it were closer to being as wide as the content
  * Screenshot 2024-10-07 at 11 06 00 AM

If we only ever want the image to be 825px wide (or even less, say 600px), we can make it that size, trim the white-space out of the image on the left/right, and center it in its container. You could do that by wrapping the markdown for the image tag in `<div class="pf-v6-u-text-align-center">[ markdown for img ]</div>`

edonehoo:
Oh I see! Okay I think the issue is just when you actually select that state in the component in Figma, it loads a solid, large circle instead of the double-lined style
image

andrew-ronaldson:
> LGTM! A couple of notes, but don't have to be done in this PR. From this screenshot:
> 
> 
> 
> Screenshot 2024-10-07 at 11 06 24 AM
> 
> 
> 
> * The native resolution is pretty big, much bigger than it will display in the browser. The file size isn't huge, but I bet it could be much smaller if we optimized it and dropped the image size. The widest it should display is 825px, so we could use that or maybe 1000px to be safe.
> 
We typically export it at 2x for better image quality. The last one I did at 1x ended up blurry. We can look at an alternative to get the size down but the right quality.

edonehoo:
^ what Andrew said about blurriness, though I typically was going 3 or 4x even, because I wanted it to be as clear as possible and didn't think about file size mattering 😅 

The extra whitespace I added to help with centering, but I see what you mean at smaller window sizes.

If I remove it, though, I feel like this huge image isn't great
image

But when I try to resize via something like `<div class="pf-v6-u-text-align-center" style="width: 500px">` or `<div class="pf-v6-u-text-align-center" style="width: 75%">`, it seems to downscale without centering:
image

Is there a better way for me to go about doing that?

mcoker:
> We typically export it at 2x for better image quality. The last one I did at 1x ended up blurry. We can look at an alternative to get the size down but the right quality.
> ^ what Andrew said about blurriness, though I typically was going 3 or 4x even, because I wanted it to be as clear as possible

Ah! Sorry, I didn't know that was a practice. It's all good if it's intentional, usually when I've seen images with a native resolution this large, it wasn't intentional. I'm curious though - was the 1x image blurry on its own? Or just when it's used in the docs on org? And are you exporting using Figma?

> But when I try to resize via something like <div class="pf-v6-u-text-align-center" style="width: 500px"> or <div class="pf-v6-u-text-align-center" style="width: 75%">, it seems to downscale without centering:

This is a good method of sizing/centering, but with a constrained width on the container, we'll need to center the container itself. So I'd make 2 updates:
* add the class `pf-v6-u-mx-auto`, which will center the `<div>` itself in the available horizontal space by using a margin-left and margin-right of "auto" (a CSS thing to center a block element).
* depending on how you want the image to scale in different viewports, you could add `min-width` and/or `max-width` to the mix here. Say you want the image to fill the available space but never grow more than 500px, you'd use `max-width: 500px`. Or if you only ever want the image to take up 75% of the available space, you can just use `width: 75%` like you have. Or if you want it to use 90% of the available space but don't let it grow past 500px, use `width: 90%; max-width: 500px`.

If this is a repeatable pattern we want to use for images in design guidelines (and potentially other docs pages), we could add a class to org that does all of this for you, and you can just do something like:

```html
<div class="ws-docs-content-img">
  ![Calendar month elements](./img/calendar-month-elements.png)
</div>
```

And we can add this to the org styles:

```css
.ws-docs-content-img {
  text-align: center;
  margin-inline: auto;
  width: 90%;
  max-width: 500px;
}
```

edonehoo:
> . I'm curious though - was the 1x image blurry on its own? Or just when it's used in the docs on org? And are you exporting using Figma?

Yes, exporting via Figma! This is a 1x exported png, so it's just blurry straight out of Figma: 
calendar-month-elements-x1. Hmm but maybe we could start trying SVGs instead? I think they're less blurry for the same export settings?

I can tinker with some of those styles! If we want to avoid the wider white border I've added, then I think we'd want to set up a class like that to make it quicker/easier

PR #4323, opened 2024-10-15
docs(page): Mention need for drop shadow with sticky page headers.

### LINKED ISSUE #3751: Page: Add guidance for sticky page headers
In the Page component products can choose to make the page header sticky and all other content scrolls below the header. In this case the sticky header should have a dropshadow but it is not clearly defined in the docs that his should occur.

Recommend adding a one-liner about this for designers so they know to include it in their specs for developer implementation.

edonehoo:
Closes #3751 

Very small change -- do we need any additional details? 

Sticky footers don't need drop shadows, right?


andrew-ronaldson (approved):
Crazemazing!

PR #4351, opened 2024-10-30
docs(develop): Adds information about breaking changes.

### LINKED ISSUE #3494: Document what a breaking change is
Ultimately this will live on org. We should have some basic guidelines for what constitutes a breaking change. Some ideas:

* A breaking change is basically _any change with code shipped in our dist package that requires users update something on their end, otherwise it will be/appear broken.* This includes:
  * Visual changes - colors, borders, text appearance, padding/margin/spacing, etc
    * Significant/non-trivial changes are breaking. Small adjustments are OK.
      * Can we get a definition for how significant a visual change has to be to be "breaking"? cc @mceledonia @mcarrano
    * Things that will cause text to wrap where it may not have before the change. eg, larger text, larger inline padding/margin/spacing, maybe a smaller bounding box, etc
     * Things that may push content down on the page (out of the viewport), making it less visible than it was before the change. eg, larger text, larger block padding/margin/spacing, etc
  * Changes to class names
  * Changes to elements - is this true, assuming the element is not directly tied to the styling in any way? Changing a div to a span or paragraph, for example, or using a heading element in place of a generic one - I would expect those to be OK. But table styling dependent upon the element/tag name should be preserved since the styling is tied to the element name.
  * Changes to selectors - unless this changes specificity in a way that could break the role of a custom property in the selector, I would also expect this to be OK.
  * Changes to HTML structure - probably implied, but in a way that breaks things - the structure before the change will be broken with the change we're making. Using optional elements, or different structures of existing elements that work and are supported to work would be fine.
  * Changes to var names and their role <sup>[1]</sup>

1. This is the big one. CSS vars/custom properties are the public facing API for styling PF. 
  * They are preserved in their name and function and we do not break them
  * In the future we may consider public vs internal vars in the future where internal vars can break, but public vars will not
  * Users should avoid:
    * Styling specific selectors and regular properties directly as those are subject to change
    * Using the values of CSS vars directly. eg, to apply a white background that matches the default/primary background used in PF, using `#fff/white/--pf-global--palette--white` instead of `--pf-global--BackgroundColor--100`. Reasons:
      * The primary background color may change to a different shade/color in the existing theme.
      * The primary bacakground color may change to a specific color in a different (eg, dark) theme
    * Using a global var, or more base level var, when a more specific one is applicable. eg, using `--pf-global--primary-color--100` instead of `--pf-c-button--m-primary--BackgroundColor` in a button to reference the background color of a primary button. Reasons - same as previous bullet point
    * A github issue should be added to request a themeable var if one does not already exist.

cc @mattnolting @srambach

edonehoo:
Closes #3494

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/get-started/develop.md
+++ b/packages/documentation-site/patternfly-docs/content/get-started/develop.md
@@ -136,3 +136,38 @@ PatternFly is supported on the 2 latest major versions of the following browsers
 - Firefox
 - Safari
 - Edge
+
+## Breaking changes
+
+A breaking change is any modification that is significant enough to require code updates in order to maintain proper functionality in a UI. To avoid major and unexpected disruptions, PatternFly typically reserves breaking changes for major version releases, but critical bug fixes can require breaking changes outside of the major release schedule. 
+
+Breaking changes can occur in both the code and the design of PatternFly features, including our beta offerings. 
+
+### Design breaking changes
+
+Breaking changes in PatternFly’s design include visual changes that would alter the UI and significantly disrupt the user experience. To restore the proper visual functioning of a UI, intentional updates are required to account for the changes.
+
+These changes include: 
+- Significant changes to a component’s style, which may be different enough to confuse users. 
+- Increases in the size of text or spacers that forcibly change content layout in a UI in ways that aren’t desired.
+
+### Core CSS breaking changes 
+
+Breaking changes to our core CSS library are the result of code changes that include:
+- New class names. 
+- New requirements for elements and classes. We will not require any elements that were not previously required in order for a component to function properly. Any new requirements will be offered as an “opt-in”. 
+- Renamed or removed CSS variables.
+
+### React breaking changes
+
+Breaking changes to our React library are the result of code changes that include:
+- Component props: 
+  - Removing a prop. 
+  - Changing the name, type, role, or function of a prop. 
+  - Making a prop required that was not previously required.  
+- Dependencies: 
+  - Changes that alter component behavior. 
+  - Changes that alter a consumer’s build process. 
+  - Changes that impact bundle size.
+  - Changes to the signature of an event handler
+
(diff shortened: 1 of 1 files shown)


packages/documentation-site/patternfly-docs/content/get-started/develop.md
```diff
@@ -136,3 +136,38 @@ PatternFly is supported on the 2 latest major versions of the following browsers
 - Firefox
 - Safari
 - Edge
+
+## Breaking changes
+
+A breaking change is any modification that is significant enough to require code updates in order to maintain proper functionality in a UI. To avoid major and unexpected disruptions, PatternFly typically reserves breaking changes for major version releases, but critical bug fixes can require breaking changes outside of the major release schedule. 
+
+Breaking changes can occur in both the code and the design of PatternFly features, including our beta offerings.
```
mcoker, line 144: I think the exception about beta components/features allowing for breaking changes gets a little lost here. Since we have sections below for design, I wonder if we could combine this with the intro paragraph?

```
A breaking change is any modification that is significant enough to require code updates in order to maintain proper functionality in a UI, and can occur in both the code and design of PatternFly features. To avoid major and unexpected disruptions, PatternFly typically reserves breaking changes for major version releases, with the following exceptions:
- Critical bug fixes can require breaking changes outside of the major release schedule.
- Components and features labeled as "beta."
```

mcoker (approved):
LOVE IT!

packages/documentation-site/patternfly-docs/content/get-started/develop.md
```diff
@@ -136,3 +136,42 @@ PatternFly is supported on the 2 latest major versions of the following browsers
...
+
+Breaking changes to our React library are the result of code changes that include:
+- Component props: 
+  - Removing a prop. 
+  - Changing the name, type, role, or function of a prop. 
+  - Making a prop required that was not previously required.  
+- Dependencies: 
+  - Changes that alter component behavior. 
+  - Changes that alter a consumer’s build process. 
+  - Changes that impact bundle size.
```
tlabaj, line 175: Can we remove this line about bundle size. It is ambiguous and arguable.

tlabaj (changes requested):

andrew-ronaldson (approved):
Apart from @tlabaj's feedback this is good to go. Thanks!

tlabaj (approved):

PR #4358, opened 2024-10-30
fix: Updated with patch from virtual assistant.

dlabaj:
Updated with virtual assistant verison 2.0.2


packages/documentation-site/patternfly-docs/patternfly-docs.css.js
```diff
@@ -1,6 +1,8 @@
 // React-specific CSS
 import "@patternfly/react-styles/src/css/components/Table/inline-edit.css"; // inline-edit extension styles
 import "@patternfly/react-styles/src/css/layouts/Toolbar/toolbar.css";
+// Patternfly virtual assistant
+import '@patternfly/virtual-assistant/dist/css/main.css';
```
rebeccaalpert, line 5: Line 5 needs to be line 20 again; otherwise this make total sense!

rebeccaalpert (changes requested):
Just need to change the import order to fix the styles and I'd be happy to approve. :)

PR #4364, opened 2024-10-31
fix(docs): fix scroll selector to scroll to top on page change

### LINKED ISSUE #4363: Bug - Scroll position doesn't change between pages
Clicking between pages on the site does not put the scroll position back to the top.

Scrolling position on first page
Image
Retained on second page.
Image

evwilkin:
Closes #4363 

This PR fixes pages not scrolling to top when navigating to a new page.  

It updates the Link component and TableOfContents component to revert the scrollSelector change from https://github.com/patternfly/patternfly-org/pull/4197 which is no longer needed after Core update https://github.com/patternfly/patternfly/pull/7025 was merged.


mcoker (approved):
LGTM!

It looks like there can be a slight delay for the page loading and the page title swap can happen after the anchor change, which feels a little weird. Is that a separate issue, or something we should look at in this PR?


https://github.com/user-attachments/assets/79bc39c2-5f65-4e2a-b508-e4fea74b3d78

evwilkin:
@mcoker good callout, I've moved the page scrolling inside the promise to trigger after page navigation 👍

mcoker (approved):
🚀

packages/documentation-framework/components/tableOfContents/tableOfContents.js
```diff
@@ -117,7 +91,7 @@ export const TableOfContents = ({ items }) => {
     <JumpLinks
       label="Table of contents"
       isVertical
-      scrollableSelector=".pf-v6-c-page__main-container"
+      scrollableSelector="#ws-page-main"
       className="ws-toc"
       style={{ top: stickyNavHeight }}
       offset={width > 1450 ? 108 + stickyNavHeight : 148 + stickyNavHeight}
```
srambach, line 97: Probably a new issue - but with the jumplinks collapsed, I notice that the scroll position puts the title underneath the jumplink toggle. The page switch looks 👌🏻 though!

srambach (approved):

packages/documentation-framework/components/tableOfContents/tableOfContents.js
evwilkin: @srambach I'm only able to recreate if I navigate to a jumplink on desktop layout and then shrink the screen down so the jumplinks move from the side to the top as seen in the video, did you see it anywhere else?  Navigating from this point forward seems to work as expected at the new screen size.


https://github.com/user-attachments/assets/1ce8fa16-9645-467c-8a7a-d3a7cdb1c12b

Issue #4365, opened 2024-10-31
Bug - Dark and light mode are not saved when refreshing the browser


dlabaj:
**Describe the problem**
When using the website it was noticed that if you refresh the page or navigate away from the page the  users current choice is not save in the browser.  This information should persist upon page refresh or when navigating back to the site. 

**Expected behavior**
Selected theme persists  for the user within their browser.

PR #4367, opened 2024-10-31
fix: Adds ability to persist theming to local storage.

dlabaj:
Fixes issue #4365 by persisting theme setting to local storage.


jonkoops (commented):
Might I suggest using the [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) selector to determine the default value? The logic would then essentially be to use the stored value only if it exists, and otherwise fall back to this logic we have [Keycloak](https://github.com/keycloak/keycloak/blob/d853dcab7d9e076a9290f11c48d19ae6b177f1ba/themes/src/main/resources/theme/keycloak/welcome/index.ftl#L11-L27) (but with the`pf-v6-theme-dark` class).

dlabaj:
> Might I suggest using the [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) selector to determine the default value? The logic would then essentially be to use the stored value only if it exists, and otherwise fall back to this logic we have [Keycloak](https://github.com/keycloak/keycloak/blob/d853dcab7d9e076a9290f11c48d19ae6b177f1ba/themes/src/main/resources/theme/keycloak/welcome/index.ftl#L11-L27) (but with the`pf-v6-theme-dark` class).

Thanks for the suggestion will look to see how to use it.  Thanks.

Issue #4249, opened 2024-09-10
All components page: Update component illustrations 


andrew-ronaldson:
Follow up to https://github.com/patternfly/patternfly-design/issues/1301

Goal is to replace the visuals with something more inline with v6 visual updates. This work includes some minor design updates to the layout and page content. See [mocks](https://www.figma.com/design/qnXKjGtXlqUk4PycBot9W7/pf.org%3A-All-Components?node-id=0-1&t=9MznqXhAq5ZPru9e-1) for details

bekah-stephens:
Here's the link to the updated illustrations - both light and dark (PNGs): https://drive.google.com/drive/folders/1oseJuG1fW2goTXTHNlRBAGMPu0713K9R

PR #4396, opened 2024-11-26
Component graphics v6 updates

andrew-ronaldson:
Follow up to design issue https://github.com/patternfly/patternfly-org/issues/4249
and Closes https://github.com/patternfly/patternfly-org/issues/4249


packages/documentation-site/patternfly-docs/images/component-illustrations/accordian.png
evwilkin, line 1: Typo - this looks to be a duplicate of accordion.png except with a typo in the file name (accordian vs. accordion)

packages/documentation-site/patternfly-docs/images/component-illustrations/about-model.png
evwilkin, line 1: Duplicale of about-modal.png except with typo, this file can be deleted.

evwilkin (changes requested):
These illustrations look great!  A few clarifications:
- `about-model.png` and `accordian.png` are duplicates with typos in the file names and can be deleted (the correctly-spelled versions do exist in this PR and are working correctly).
- `All components --splash --light.png` is the only net-new illustration but is not exported in the index file or used anywhere, is this supposed to be displayed somewhere?
- Prior illustrations were 500x500px, new illustrations are 1288 x 704px.  This aspect ratio works better to not make the cards/list rows unnecessarily tall, but the largest I am able to make the images display on the site is 639 x 319.5px (list view at 767px screen width).  The illustrations appear to be created at ~2x this resolution, confirming this is intentional for high dpi displays?  If not needed we can likely reduce file size with smaller images
Screenshot 2024-11-26 at 3 41 00 PM

andrew-ronaldson:
The images were 4x scale so now they are back down to 2x. I removed the extra images that were misspelled and the ALL components image as it will be addressed in another issue.

edonehoo:
image

Just noticed that it looks like some of the cards are full height, while some are more narrow. But as I adjust my window to be smaller, the full-height cards become narrow and the originally-narrow cards swap to full height 🤔

image

evwilkin:
@edonehoo the cards are in each row should be the same height, and are set to match the height of the tallest card in that row - so in your second screenshot you'll see Application Launcher is extra tall to fit the "Demo" tag at the bottom so all other cards stretch so the height is universal across the row.  

As cards shift between rows, any rows without those extra tags don't need the extra space to accommodate them and so don't stretch vertically - so rows aren't a uniform height, but cards within a row should all match.  Hopefully that makes sense, but if it's still acting wonky or there's a different suggestion for how to handle that we can certainly revisit this logic 😄

andrew-ronaldson:
I noticed a couple images that contained drop shadows were increase height on certain images. I don't think it will make a big difference but wanted them to be consistent

Issue #4404, opened 2024-12-05
Bug - dark theme toggle issues


mcoker:
In https://github.com/patternfly/patternfly-org/pull/4389, we added this `<meta>` tag to set [`color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) https://github.com/patternfly/patternfly-org/blob/6493439f89d03c8f33e7a5d14574582dde5a7b7b/packages/documentation-framework/templates/html.ejs#L6

Basically what that does is tells the page that it can render in either light or dark mode (preferring light as the default, since it comes first), based on the device/browser's preference (if `@media (prefers-color-scheme: dark)`, it will set `color-scheme: dark`). However we also allow you to toggle dark theme independent of the OS setting, so if you're looking at patternfly.org in light theme and your OS/browser is set to dark theme, you'll get light theme PF with `color-scheme: dark` stuff, like dark checkboxes and the default html/body background color as black instead of white. Exhibits A & B below:

([checkbox component page](https://staging.patternfly.org/components/forms/checkbox/))
![Image](https://github.com/user-attachments/assets/e80452d3-54a7-41c6-ac21-37199045bda6)

([full page checkbox example](https://staging.patternfly.org/components/forms/checkbox/react/controlled/))
![Image](https://github.com/user-attachments/assets/a9a5761a-9c21-42c3-a8ec-06b37cf23cb5)

That's the problem. Here are some ideas for a fix.

PF CSS currently [sets `color-scheme: dark`](https://github.com/patternfly/patternfly/blob/d3213a69149a204bf614ec02af4d9464589f196b/src/patternfly/base/normalize.scss#L125-L127) when you add `.pf-v6-theme-dark` to the `<html>` element. One fix could be for core to set `color-scheme: light` as a default (when dark theme is not enabled).

We could also just remove the `<meta name="color-scheme" content="light dark">` tag, but that begs the question for why you would use `<meta name="color-scheme">` in the first place. AFAIK the main benefit of `<meta>` over CSS is that the `color-scheme` changes from a `<meta>` tag apply pretty much immediately as the page is loading, since it's in the doc `<head>`, where `color-scheme` in CSS relies on the CSS loading first. So with a CSS only approach like we currently have, the page may load with no `color-scheme` set (defaults to a light theme), only to switch to `color-scheme: dark` after the CSS loads, potentially creating a UI change after stuff is already on the page.

[...]

PR #4408, opened 2024-12-09
fix(dark-theme): fixed issue with incorrect color-scheme being applied

mcoker:
fixes https://github.com/patternfly/patternfly-org/issues/4404

### DIFF
--- a/packages/documentation-framework/templates/html.ejs
+++ b/packages/documentation-framework/templates/html.ejs
@@ -3,7 +3,6 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width,initial-scale=1,shrink-to-fit=no">
-    <meta name="color-scheme" content="light dark">
     <meta name="description" content="PatternFly is Red Hat's open source design system. It consists of components, documentation, and code for building enterprise applications at scale.">
     <title><%= title %></title>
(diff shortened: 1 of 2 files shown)


nicolethoen (approved):

dlabaj (approved):

kmcfaul (approved):

packages/documentation-framework/templates/html.ejs
```diff
@@ -3,7 +3,6 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width,initial-scale=1,shrink-to-fit=no">
-    <meta name="color-scheme" content="light dark">
```
jonkoops, line 6: This metadata indicates that dark mode as well as light mode is supported, why remove it?

dlabaj (approved):

packages/documentation-framework/templates/html.ejs
kmcfaul: This was creating an issue where the system preference was dictating the color of checkboxes and radio buttons, superseding the actual selected theme.

jonkoops: Talked to @mcoker about it and reverting this is fine, we'll have to get back to a proper fix at a later time :+1:

PR #4387, opened 2024-11-20
chore: upgrade Yarn to the latest version

jonkoops:
Upgrades Yarn to a version that is not end-of-life. This is just to make my life easier as I can't get the old Yarn to work on my system for some reason.

PR #4413, opened 2024-12-11
chore: upgrade Yarn to the latest version

jonkoops:
Upgrades Yarn to a version that is not end-of-life. This PR is a reworked version of #4387, which ran into issues after merging. This turned out to be because of a bad rebase, which caused the versions of `@patternfly/documentation-framework` to be out of sync (a common issue that happens on `main` as well).

This version improves over the original PR as the `@patternfly/documentation-framework` package is now using the [workspace protocol](https://yarnpkg.com/features/workspaces#cross-references), meaning it will never be out of sync again, as it is directly linked to the version present in the workspace. This also mitigates the need to manually bump this version when it is changed.

~Note the CI will likely fail for the reasons outlined under https://github.com/patternfly/patternfly-org/pull/4387#issuecomment-2489310204 due to Corepack not being present on `main`. I will see if I can split this into two separate PRs to increase the confidence in the merge.~


jonkoops:
I've rebased the PR on top of `main` after merging in #4414, this should be good to go now. I assume the Snyk failure is due to the changes made by Yarn, likely some dependency is now being picked up properly, but I cannot see the report to validate this.

Either way, nothing has changed in terms of dependency versions, so this should not really be a reason to hold this up, marking this as ready to review.

nicolethoen (approved):
I compared the snyk report to the current state of snyk in the main branch. It has not changed. We still need to resolve the vulnerabilities, but this PR is not having any overall affect on that. I resolved the snyk report in this PR so it's not a blocker

jonkoops:
Nice, thanks for the merge Nicole! :tada:

PR #4438, opened 2025-01-21
Bug - Charts - CodeSandbox default imports

dlabrecq:
Add victory dependency to chart CodeSandbox examples

Closes https://github.com/patternfly/patternfly-react/issues/11441


evwilkin (approved):
LGTM - thanks for the fix!

PR #4448, opened 2025-01-24
docs(charts): Updates colors for charts page.

edonehoo:
Closes #4412

PR #4451, opened 2025-01-29
fix: update colors grid

### LINKED ISSUE #4444: Bug - Colors for Charts page uses old color tokens
The page is using some of the old color tokens that have been replaced in v6:
- `gold` should be `yellow`
- `cyan` should be `teal`
- `red` should be `red-orange`

### LINKED ISSUE #4447: Bug - PF tokens table is not up to date
The tokens table on our docs is not up to the date with the latest tokens used in PF6 and needs to be updated.

(example: `--pf-t--global--font--weight--heading` is listed but the token is actually named `--pf-t--global--font--weight--heading--default`)

logonoff:
fixes #4444

before:
image
image

after:
image
image

I had to change the grid size as "Red Orange" simply takes up too much space:
image


logonoff:
Duplicate #4448, I'll close this PR

PR #4488, opened 2025-02-27
docs(charts): Adds developer notes to about-charts page

### LINKED ISSUE #2974: Chart styles documentation
Now that a lot of chart styles have reference to CSS variables, we've had the question of where the vars are defined and how to use/reference them. Seems like this could be a good addition to the charts section on the website - similar to (or part of) the "about" page?

edonehoo:
Closes #2974 

As a follow up to this PR, I'll need to remove redundant info that I copied from the dark theme handbook.

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/design-guidelines/charts/about/about.md
+++ b/packages/documentation-site/patternfly-docs/content/design-guidelines/charts/about/about.md
@@ -8,7 +8,7 @@ _Charts are only available in React_
 A **chart** visualizes data in an application. The type of chart you use will depend on your use case and the type of data you need to display.
 
 ## Elements
-<img src="./img/axis-labels.png" alt="Chart with axis labels" width="713"/> 
+<img src="./img/axis-labels.png" alt="Chart with axis labels" width="713"/> 
 
 1. **Title:** Titles should be left-aligned and appear above your chart.
 2. **Axis labels and scale values:**  We recommend using tick marks to clearly mark scale values on the y-axis. The text for axis labels and scale values should be 12px font in standard text color. All scale values should be equally distributed across the axis and the axis label should fall outside the scale values, centered with the axis line.
@@ -16,7 +16,7 @@ A **chart** visualizes data in an application. The type of chart you use will de
 4. **Chart tooltip:** A tooltip will appear on hover over certain elements of a chart, like bars in a bar chart or segments in a donut chart. [Chart tooltips](/charts/tooltips) should display the specific values related to the element the user is hovering over. 
 
 ## Chart types
-<img src="./img/chart-legend.png" alt="Image displaying what chart types to use when" width="1024"/> 
+<img src="./img/chart-legend.png" alt="Image displaying what chart types to use when" width="1024"/> 
 
 - [Area chart](/charts/area-chart/design-guidelines): Use to show (potentially multiple) trends over a continuous scale (usually time).
 - [Bar chart](/charts/bar-chart/design-guidelines): Use to show and compare categories and their respective data point/value. Bar charts can be plotted vertically or horizontally. The axis that labels the bar is the category, the axis that marks the end point of the bar displays the data points/values.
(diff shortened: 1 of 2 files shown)


edonehoo:
@mcoker can you review when you have time? 

I'm curious if it's clear enough, especially since I mention both the React tokens and the standard design tokens. Do you think we need to expand more on if/when to use the chart design tokens instead of or in addition to the React tokens? And if so, that's an area I think I could use clarification too 😊

packages/documentation-site/patternfly-docs/content/design-guidelines/charts/about/about.md
```diff
@@ -65,3 +65,15 @@ On click
 
 1. **Legend:** When a user clicks on a legend label, it becomes disabled and the color swatch is replaced with an eye-slashed icon.
 2. **Chart:** Data corresponding to the clicked legend label is hidden from view.
+
+## Develop with charts 
+
+Default styles in the [@patternfly/react-charts package](https://www.npmjs.com/package/@patternfly/react-charts) are aligned with our light theme. To use chart styles, you must also have the [@patternfly/patternfly package](https://www.npmjs.com/package/@patternfly/patternfly) installed.
```
mcoker, line 71: I'm a little confused about what it means to "use chart styles". Can you elaborate on that a bit? Just so it's clear (it probably is, just want to make sure) - `@patternfly/react-charts` in light theme works just fine without `@patternfly/patternfly` or any CSS from `@patternfly/patternfly`. The styles are all basically hard-coded into the chart SVG's.

packages/documentation-site/patternfly-docs/content/design-guidelines/charts/about/about.md
```diff
@@ -65,3 +65,15 @@ On click
...
+
+To utilize default chart styles, you don't need to import anything else. But, to support dark-themed charts, you must also import the stylesheet that contains dark theme styles. To do so, add this line before importing your main application component:
```
mcoker, line 73: Also confused about "utilize default chart styles" - not sure what "utilize" means exactly. Same with "to support dark-themed charts" - not sure what "support" means. If I simplifed it, I'd say charts works with light theme out of the box, no additional styles or anything are needed. However, for charts to work in dark theme, you'll need to import `@patternfly/patternfly/patternfly-charts.css`

packages/documentation-site/patternfly-docs/content/design-guidelines/charts/about/about.md
```diff
@@ -65,3 +65,15 @@ On click
...
+
+`import '@patternfly/patternfly/patternfly-charts.css';`
+
+Once you import this file, you'll have access to [all chart variables](https://www.npmjs.com/package/@patternfly/patternfly?activeTab=code). Beyond dark theme, you could use these variables to match the style of other UI elements to your chart styles.
```
mcoker, line 77: IMO it isn't really intuitive that "all chart variables" links to `@patternfly/patternfly` on NPM. I would probably be confused clicking on that, expecting it to take me to a list of chart variables or something like that. I might pair this sentence with the one below about filtering the all tokens table, but it's worth noting that chart variables and chart tokens aren't exactly the same.

mcoker (commented):
Hey @edonehoo! Sorry for the delay. I left some comments, but I'm not sure if they're helpful 😅 If not, I'm happy to meet up and we can probably chat through it pretty quickly.

packages/documentation-site/patternfly-docs/content/design-guidelines/charts/about/about.md
edonehoo: Ah nope, I was misunderstanding/assuming that you had to install @patternfly/patternfly in either case! This is ringing a bell for me now, ty

packages/documentation-site/patternfly-docs/content/design-guidelines/charts/about/about.md
edonehoo: Just to clarify, when you say "global tokens", do you mean design tokens? Or [this file](https://github.com/patternfly/patternfly/blob/main/src/patternfly/base/tokens/tokens-default.scss)? Which, this file is ultimately our design tokens..right?  

I should find time to create a cheat sheet for some of these terms React tokens, design tokens, variables, styles, etc. I get so mixed up 😅

edonehoo: Your simplified suggestion makes sense outside of me making sure I understand the verbiage, though! Gonna make updates based off this

packages/documentation-site/patternfly-docs/content/design-guidelines/charts/about/about.md
edonehoo: Okay this is great feedback! I was thinking that the patternfly-charts.css file over on NPM _was_ the chart variables, but I'm definitely treading water with variables, tokens, and friends lol

edonehoo: I think my remaining question is, where is a list of our chart variables?

packages/documentation-site/patternfly-docs/content/design-guidelines/charts/about/about.md
```diff
@@ -65,3 +65,14 @@ On click
 
 1. **Legend:** When a user clicks on a legend label, it becomes disabled and the color swatch is replaced with an eye-slashed icon.
 2. **Chart:** Data corresponding to the clicked legend label is hidden from view.
+
+## Develop with charts 
+
+Default styles in the [@patternfly/react-charts package](https://www.npmjs.com/package/@patternfly/react-charts) are aligned with our light theme. Charts work with PatternFly's light theme by default&mdash;you don't need to import anything else. 
+
+To support dark-themed charts, you must:
+1. Import the [@patternfly/patternfly package](https://www.npmjs.com/package/@patternfly/patternfly), so that you can use our global tokens.
```
mcoker, line 74: Are we wanting to provide documentation for an app that is not using PatternFly? Or are we assuming the charts will be included in a PatternFly app? If they aren't using PatternFly already, they'll need to import some other things.

mcoker (approved):
Left a question, but I'm also good with this as it is! I know this PR has been open a while so I don't want to hold it up any more than I already have 😅

nicolethoen (approved):

PR #4494, opened 2025-03-10
chore(Upgrade guide): removed verbiage on button aria-disabled

thatblindgeye:
Depends on https://github.com/patternfly/patternfly-react/pull/11619


thatblindgeye:
@edonehoo would you mind taking a look at this update? i originally put this PR up before one of yours went in that updated this page a bit and just pushed a tweak based on a convo Titani and I had yesterday.

tlabaj (approved):
lgtm

packages/documentation-site/patternfly-docs/content/get-started/upgrade.md
```diff
@@ -172,8 +172,8 @@ If you have previously implemented any breakpoint logic based on a pixel value,
 #### Button
 
 1. *Cannot find aria-disabled*
-    - **Cause:** We changed the `isDisabled` prop to assign a value for `disabled`, but none for `aria-disabled`. 
-    - **Fix:** Remove tests that look for `aria-disabled` in buttons.
+    - **Cause:** We changed button's `isDisabled` prop to assign a value for `disabled`, but none for `aria-disabled` - except for when the `component` prop is passed anything other than "button" as a value. Additionally, `aria-disabled` will now only render when true.
```
edonehoo, line 175:
```suggestion
    - **Cause:** We changed button's `isDisabled` prop to assign a value for `disabled`, but none for `aria-disabled`. The exception to this is when the `component` prop is passed anything other than "button" as a value. Additionally, `aria-disabled` will now only render when true.
```

packages/documentation-site/patternfly-docs/content/get-started/upgrade.md
```diff
@@ -172,8 +172,8 @@ If you have previously implemented any breakpoint logic based on a pixel value,
...
+    - **Fix:** Remove tests that look for `aria-disabled` in buttons when the expectation is for it to be false, or to match `isDisabled` when `component="button"`.
```
edonehoo, line 176:
```suggestion
    - **Fix:** Remove tests that look for `aria-disabled` in buttons when the expectation is either: 
        - For `aria-disabled` to be false.
        - For `aria-disabled` to match `isDisabled` when `component="button"`.
```
I'm not sure if I'm correctly interpreting the end of the sentence. I took it like there are 2 instances where you should remove these tests, but I can also interpret it like this (I'll just share options for both): 

> **Fix:** Either:
>     - Remove tests that look for `aria-disabled` in buttons, when the expectation is for `aria-disabled` to be false.
>     - Match `aria-disabled` to `isDisabled` when `component="button"`.

edonehoo (commented):
I left a couple of options for line 176, because I wasn't 100% sure if I was reading that right. If both are wrong, lmk 😅

PR #4508, opened 2025-03-24
docs(cards): Updates images for v6.

### LINKED ISSUE #4275: Guideline updates: Cards & Tiles

bekah-stephens:
Closes #4275


bekah-stephens:
@edonehoo I've updated the image in the "Border treatments" section, but the content needs to be tweaked.

Short term fix: change the section to be "Styling" and call out the difference between primary and secondary cards.
Longer term fix: look at moving this section within Variations, rather than Usage where it is now.

packages/documentation-site/patternfly-docs/content/design-guidelines/components/card/img/card-images.svg
edonehoo, line 1: I think there may (?) have been a chart in your second image that was in a different frame or something!

bekah-stephens: whoops! fixing it now

edonehoo (approved):
I left a comment about a small caption thing that could be tweaked, but these look great!

andrew-ronaldson (approved):
💰

nicolethoen (approved):

PR #4517, opened 2025-04-04
feat(ver): update chatbot

kmcfaul:


rebeccaalpert (approved):
This runs. We need to make sure ChatBot has the right 6.2. versions so the dropdowns work but styles look ok.

PR #4535, opened 2025-04-09
fix: css variables table has colors again & charts dependencies fixed

nicolethoen:


dlabrecq (approved):

dlabaj (approved):
tested and chart error is fixed. thank you.

Issue #4520, opened 2025-04-05
Guideline updates: List,  Simple list, Data list & Decription list


andrew-ronaldson:

edonehoo:
FYI, these updates should also close https://github.com/patternfly/patternfly-org/issues/4492

PR #4544, opened 2025-04-11
docs(list): Update unordered list image.

### LINKED ISSUE #4492: Unordered list design guidelines image is potentially confusing
The first image under https://www.patternfly.org/components/list/design-guidelines#ordered seems to be a potential cause for confusion, as the Wizard can cause a misinterpretation that we offer an unordered list with the wizard step styling, as happened in [this Slack thread](https://redhat-internal.slack.com/archives/C023VGW21NU/p1741262340377949).

I think we should replace that image with one of an unordered list outside of a Wizard.

edonehoo:
Closes #4492 for v5, v6 update handled via https://github.com/patternfly/patternfly-org/issues/4520


edonehoo:
@dlabaj not sure about the merge block message or build error

PR #4585, opened 2025-04-24
Revert "chore(deps): Temporarily allow for React 19 peer dependency"

kmcfaul:
Reverts patternfly/patternfly-org#4582


kmcfaul:
reverting in favor of bumping with the prerelease versions in a single PR

PR #4595, opened 2025-05-02
docs(upgrade): Fix link to `card` component

tiagobento:
Without the leading `/` the link is relative, thus:
- Doesn't work when clicked (but keeps adding to the URL); and
- Opens a non-existent page when opened in a new tab.

Btw, I gotta say from a first glance I already love the look & feel of PF6. Can't wait to upgrade the [tools repo of Apache KIE](https://github.com/apache/incubator-kie-tools) to it! We just finished upgrading to v5 🙏


nicolethoen (approved):
Thank you so much for your contribution!

PR #4600, opened 2025-05-06
docs(login page): Updates images for v6.

### LINKED ISSUE #4553: Guideline updates: Login page

bekah-stephens:
Closes #4553


andrew-ronaldson (changes requested):
Thanks, Bekah! i just see one change in two of the images. The background image of the pattern should be bottom-right with no spacing. So all the way to the edge of the image.

bekah-stephens:
> Thanks, Bekah! i just see one change in two of the images. The background image of the pattern should be bottom-right with no spacing. So all the way to the edge of the image.

fixing the screenshots, but does that change need to be made on the Desktop Login Page component? I pulled it straight from the library and the background image has that spacing

PR #4626, opened 2025-05-20
Iss4546

### LINKED ISSUE #4546: Docs - Update Drawer guidelines
**Description**
There’s currently only one screenshot with a description in the guidelines for this component (re: splitter/resizing)

**Recommendation / Fix**
Provide more examples / explanations here? Maybe a placement section (right, left, main content area)?

edonehoo:
Closes #4546


edonehoo:
IDK if this is a big deal, but I noticed that the "overlay drawer" image still pushes the page content to the side, which isn't 100% accurate. It should just overlay it, so the page content should extend to the full width of the screen, including the drawer. I couldn't figure out how to fix it in Figma, so I can coordinate with whoever else can!

bekah-stephens:
> IDK if this is a big deal, but I noticed that the "overlay drawer" image still pushes the page content to the side, which isn't 100% accurate. It should just overlay it, so the page content should extend to the full width of the screen, including the drawer. I couldn't figure out how to fix it in Figma, so I can coordinate with whoever else can!

I'm not sure the best way to fix the component itself, but did a quick fix and updated the screenshot. I would say pull in the fixed screenshot for the guidelines and then add the component problem to the audit doc.

edonehoo:
@bekah-stephens ty! Updated and added to the audit sheet

andrew-ronaldson (approved):

PR #4630, opened 2025-05-27
docs(typography): Adds guidance for line length.

### LINKED ISSUE #4288: Typography & Content docs to add guidance for max line length
Currently the typography and content pages do not provide design guidelines for a max line length. There is an common design principle that lines of text should not span large content areas but I would like to investigate this more and provide clear guidance for our users.

edonehoo:
Closes #4288


andrew-ronaldson (approved):
Incredible

PR #4639, opened 2025-05-30
feat: Enhanced theme switcher with system preference detection and tri-state selection

srambach:
## 🎨 Enhanced Theme Switcher Implementation

This PR implements a comprehensive theme switcher that provides users with three choices: **System**, **Light**, and **Dark** themes, with proper persistence and synchronization across the entire application.

### ✨ Features

1. **🖥️ Automatic System Preference Detection**
   - Detects user's `prefers-color-scheme` setting automatically
   - Dynamically responds to system theme changes in real-time
   - Defaults to system preference on first visit

2. **🎯 Tri-State Theme Selection**
   - **System**: Follow user's operating system preference
   - **Light**: Always use light theme regardless of system setting
   - **Dark**: Always use dark theme regardless of system setting

3. **💾 Persistent User Preferences**
   - Saves user's theme choice in localStorage
   - Persists across page reloads and browser sessions
   - Gracefully handles missing localStorage (SSR-safe)

4. **🔄 Perfect Synchronization**
   - Single source of truth for theme state
   - Synchronized between header switcher and full-screen examples
   - No more conflicting theme instances

#### New Shared Theme Hook (`useTheme.js`)
- Centralized theme management with React hooks
- Handles system preference detection and changes
- Manages localStorage persistence
- Applies theme classes to document root

#### Enhanced UI Components
- **Header**: Replaced toggle with Select component showing current mode and descriptions
- **Full-screen examples**: Consistent Select component with same options
- **Icons**: Desktop, Sun, and Moon icons to clearly indicate each mode

#### Architecture Improvements
- Moved theme management to `SideNavLayout` level for single source of truth
- Eliminated duplicate theme logic between components
- Props-based state sharing to prevent synchronization issues

### 🐛 Bug Fixes

- ✅ Fixed system preference affecting UI when explicit theme is selected
- ✅ Eliminated need for page reload after theme changes
- ✅ Resolved theme synchronization issues between components
- ✅ Improved SSR compatibility with proper client-side hydration

### 🎭 User Experience

- **Intuitive Selection**: Clear labels and descriptions for each theme mode
- **Visual Feedback**: Icons immediately show which mode is active
- **Consistent Behavior**: Theme choice works identically across all pages
- **Responsive to System**: Automatically adapts when user changes OS theme (in System mode)

### 📦 Files Changed

- `packages/documentation-framework/hooks/useTheme.js` (new)
- `packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js`
- `packages/documentation-framework/components/example/example.js`

### 🧪 Testing

This implementation has been tested with:
- ✅ Manual theme switching between all three modes
- ✅ System preference changes (OS dark/light mode toggle)
- ✅ Page reloads maintaining user preference
- ✅ Full-screen example synchronization
- ✅ Browser sessions and localStorage persistence

### 🎯 Addresses

This implementation builds upon and completes the work started in PR #4367, providing the full tri-state theme selection that was originally requested while fixing the synchronization issues that prevented the earlier implementation from being merged.

---

**Ready for review!** 🚀

### DIFF
--- a/packages/documentation-framework/hooks/useTheme.js
+++ b/packages/documentation-framework/hooks/useTheme.js
@@ -9,6 +9,14 @@ const THEME_MODES = {
 const THEME_STORAGE_KEY = 'theme-preference';
 const DARK_MODE_CLASS = 'pf-v6-theme-dark';
 
+/**
+ * Custom hook for managing theme state with system preference detection
+ * @returns {Object} Theme state and controls
+ * @returns {string} themeMode - Current theme mode ('system'|'light'|'dark')
+ * @returns {Function} setThemeMode - Function to change theme mode
+ * @returns {string} resolvedTheme - The actual applied theme ('light'|'dark')
+ * @returns {Object} THEME_MODES - Available theme mode constants
+ */
 export const useTheme = () => {
   const getStoredThemeMode = () => {
     if (typeof window === 'undefined' || !window.localStorage) return null;
@@ -21,16 +29,27 @@ export const useTheme = () => {
   };
 
   const getResolvedTheme = (mode) => {
-    if (typeof window === 'undefined') return 'light';
+    // SSR-safe check for window and matchMedia
+    if (typeof window === 'undefined' || !window.matchMedia) return 'light';
+    
     if (mode === THEME_MODES.SYSTEM) {
-      return window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
+      try {
+        return window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
+      } catch (error) {
+        // Fallback if matchMedia fails
+        console.warn('matchMedia not supported, defaulting to light theme');
+        return 'light';
+      }
     }
     return mode;
   };
 
   const updateThemeClass = (resolvedTheme) => {
-    if (typeof window === 'undefined') return;
+    if (typeof window === 'undefined' || !document) return;
+    
     const htmlElement = document.querySelector('html');
+    if (!htmlElement) return;
+    
     if (resolvedTheme === 'dark') {
       htmlElement.classList.add(DARK_MODE_CLASS);
     } else {
@@ -56,9 +75,16 @@ export const useTheme = () => {
 
   // Listen for system preference changes
   useEffect(() => {
-    if (typeof window === 'undefined') return;
+    // Enhanced SSR-safe check
+    if (typeof window === 'undefined' || !window.matchMedia) return;
 
-    const mediaQuery = window.matchMedia('(prefers-color-scheme: dark)');
+    let mediaQuery;
+    try {
+      mediaQuery = window.matchMedia('(prefers-color-scheme: dark)');
+    } catch (error) {
+      console.warn('matchMedia not supported, skipping system theme detection');
+      return;
+    }
     
     const handleSystemThemeChange = (e) => {
       if (themeMode === THEME_MODES.SYSTEM) {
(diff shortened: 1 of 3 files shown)


packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -70,10 +75,35 @@ const HeaderTools = ({
...
-    document.querySelector('html').classList.toggle('pf-v6-theme-dark', darkThemeToggleClicked);
-    setIsDarkTheme(darkThemeToggleClicked);
+  const handleThemeChange = (_event, selectedMode) => {
+    setThemeMode(selectedMode);
+    setIsThemeSelectOpen(false);
+  };
+
+  const getThemeDisplayText = (mode) => {
+    switch (mode) {
+      case THEME_MODES.SYSTEM:
```
rebeccaalpert, line 85: Same as above.

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -70,10 +75,35 @@ const HeaderTools = ({
...
+      case THEME_MODES.DARK:
+        return 'Dark';
+      default:
+        return 'System';
+    }
+  };
+
+  const getThemeIcon = (mode) => {
+    switch (mode) {
+      case THEME_MODES.SYSTEM:
```
rebeccaalpert, line 98: Same as above.

rebeccaalpert (approved):
I think you and your robot friend did a good job. I read through and it seems understandable. I just left a few nitpicky comments about the switches, but you're free to ignore. It won't cause any problems if it's left as-is.

Gotta figure out my back-up plan for when our robot overlords put me out of work. :)

rebeccaalpert (approved):

edonehoo:
I would recommend "System theme", "Light theme", "Dark theme" as the toggle text to reflect what's selected or pre-selected

Or you could just have "Theme" and not show selections in the toggle [(like Atlassian does)](https://atlassian.design/)

packages/documentation-framework/layouts/sideNavLayout/sideNavLayout.js
```diff
@@ -94,20 +120,49 @@ const HeaderTools = ({
...
+                selected={themeMode}
+                onSelect={handleThemeChange}
+                onOpenChange={(isOpen) => setIsThemeSelectOpen(isOpen)}
+                toggle={(toggleRef) => (
+                  <MenuToggle
+                    ref={toggleRef}
+                    onClick={() => setIsThemeSelectOpen(!isThemeSelectOpen)}
+                    isExpanded={isThemeSelectOpen}
+                    icon={getThemeIcon(themeMode)}
+                    aria-label="Theme selection"
```
thatblindgeye, line 135: This would apply to the example js file as well, but for the aria-label, instead of just this string it'd be more informative if we include the current theme selection so that the toggle is announced as "Dark theme, Theme selection".

A wai aria example of a combobox results in a similar accessible name structure (though it kinda gets it by default because their toggle is a combobox role, so it has a `value` attribute that gets included).

We'd need to do something like ```aria-label={`${getThemeDisplayText(themeMode)}, Theme selection`}``` (though creating a variable for the getThemeDisplayText would be better so that we can use that variable here and on line 137)

packages/documentation-framework/hooks/useTheme.js
```diff
@@ -0,0 +1,127 @@
...
+
+  return {
+    themeMode,
+    setThemeMode,
+    resolvedTheme,
+    THEME_MODES
+  };
+};
+
+export { THEME_MODES };
```
thatblindgeye, line 127: since useTheme is already being exported and returns this, we could avoid exporting this. If we want to export this on its own, though, we should just update line 3 to `export const THEME_MODES ...`

packages/documentation-framework/hooks/useTheme.js
```diff
@@ -0,0 +1,127 @@
...
+ * Custom hook for managing theme state with system preference detection
+ * @returns {Object} Theme state and controls
+ * @returns {string} themeMode - Current theme mode ('system'|'light'|'dark')
+ * @returns {Function} setThemeMode - Function to change theme mode
+ * @returns {string} resolvedTheme - The actual applied theme ('light'|'dark')
+ * @returns {Object} THEME_MODES - Available theme mode constants
+ */
+export const useTheme = () => {
+  const getStoredThemeMode = () => {
+    if (typeof window === 'undefined' || !window.localStorage) return null;
```
thatblindgeye, line 22: This is fairly nitpicky and more to follow existing syntax, but rather than the inline return, just wrapping the curly braces around:

```suggestion
    if (typeof window === 'undefined' || !window.localStorage) {
    return null;  
  }
```

Which would apply to other returns in this hook as well, but again fairly nitpicky.

thatblindgeye (changes requested):
In addition to the comment below, one non-blocker I'm noticing is when the page first loads, the "System" default can be seen before updating to the theme selection in local storage (at least when "Light" is the stored theme). It happens really quickly and probably only noticeable if you're looking for it, but if it is something we'd want to take care of a followup issue would be fine (but also if we want to see if astro will handle it we can do that sa well).

sg00dwin (approved):
Great work Sarah!

PR #4671, opened 2025-06-18
docs(icons): Adds more guidance for removal and deletion.

### LINKED ISSUE #4659: Guidelines update: Remove vs Delete
Related to [file uploads](https://www.patternfly.org/components/file-upload/simple-file-upload) and the [Chatbot image upload.](https://github.com/patternfly/patternfly-design-kit/issues/807)

Wonder if the action button and tooltip terminology for deleting a file should be "remove" or "delete" or "clear" for file uploads? [Our ux terminology guidelines](https://www.patternfly.org/ux-writing/terminology) seems like remove might be the right term (the example on the site even says "remove file".)

the icon guidelines say the minus circle is used to remove items vs trash means to delete. Slack & gmail both use "remove" when uploading a file. Whatever we choose, the icon guidelines will need updated to reflect that

I am inclined to use remove as the action paired with a trash can icon or a tooltip that says "remove" but I will defer to Erin for a recommendation and guidelines update

edonehoo:
Closes #4659 

This is a small update to address the issue -- I think this is a good solution to cover the additional use cases without removing guidance that products are likely already following

but more work could come in follow-up issues if we want:
- Adding icons within the UX writing terminology table
- Adding a new doc to Patterns to document general removal and deletion patterns/scenarios

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/design-guidelines/styles/icons/icons.js
+++ b/packages/documentation-site/patternfly-docs/content/design-guidelines/styles/icons/icons.js
@@ -531,14 +531,14 @@ export const iconsData = [
     "Name": "fa-minus",
     "React_name": "MinusIcon",
     "Type": "Action",
-    "Contextual_usage": "Indicates the ability to remove an item",
+    "Contextual_usage": "Indicates the ability to remove an item (such as a table row), decrease (in number input controls), or minimize",
   },
   {
     "Style": "fas",
     "Name": "fa-minus-circle",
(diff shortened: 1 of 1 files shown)


kaylachumley (approved):
sounds good! do we have a follow up issue already for adding the icon table to the terminology page? I know Pat was pro going that direction but wasn't sure if we ever made an issue post that discussion

edonehoo:
@kaylachumley just opened one, thanks!

PR #4675, opened 2025-06-23
docs(wizard): Updates images for v6.

### LINKED ISSUE #4576: Guideline updates: Wizard

bekah-stephens:
Closes #4576


kaylachumley (commented):
not sure if its just me but all of the images look blurry on my side! besides that the rest looks good

bekah-stephens:
> not sure if its just me but all of the images look blurry on my side! besides that the rest looks good

hmm not blurry on my end - what about you @edonehoo ?

edonehoo (approved):

nicolethoen:
I agree with @kaylachumley - they look blurry to me. I think the images are a little large be default and getting like squashed a bit causing the images to look blurry? I might have a poorer quality monitor than some other people?

edonehoo:
@nicolethoen do you think anything about this css could be causing that? is the max width scaling down too much? I don't think we have control over the original svg size that Figma exports 🤔

image

PR #4686, opened 2025-07-02
chore(timestamp): image updates

andrew-ronaldson:
Closes https://github.com/patternfly/patternfly-org/issues/4571


packages/documentation-site/patternfly-docs/content/design-guidelines/components/timestamp/timestamp-elements.svg
bekah-stephens, line 1: Generally a little confused by this screenshot. Is the second example just a tooltip version? I'm not sure the #5 annotation makes sense since it's calling out the dotted line (and there isn't one on the tooltip)

andrew-ronaldson: I thought the same. The 5 on the dotted underline indicates there is a tooltip and 5 is also the tooltip?

bekah-stephens: maybe we can just update the #5 annotation? @edonehoo

edonehoo: yeah the current image is weird imo. Could you update the image so that the tooltip is shown above the timestamp? And then the #5 annotation just points to the tooltip

Like this, if that makes sense? 

image

bekah-stephens (approved):
✨👌🏻

PR #4691, opened 2025-07-02
fix(Lockfile): Remove PF deps from ChatBot

rebeccaalpert:
Causes dropdowns to break in Org due to mismatched versions.


nicolethoen:
Will products have this same issue?

PR #4702, opened 2025-07-11
docs(motion): Expands motion guidelines with recent updates.

edonehoo:
Closes https://github.com/patternfly/patternfly-design/issues/1401

This was as deep as I understand as far as technical/dev details. If we want more specifics, rather than linking to implemented examples, will need some more information compiled


andrew-ronaldson (commented):
Couple of questions and follow up thoughts for later.

- Where you talk about opt-ins do we want to mention/link to the codemods?
- Not sure if it is the staging site but the example links of the components on the dev page don't jump to the right section. The links are all correct so not sure why it's being a pain.
- Design should create example gifs of the timing/easing/etc. sections to break up the text and give a sense of how things work. I'll make an issue for that and I think it can be done with @bekah-stephens' visual library work.

andrew-ronaldson (approved):

dlabaj (approved):

Issue #4709, opened 2025-07-18
Docs framework - theme switcher updates


mcoker:
In talking with @andrew-ronaldson let's POC
* An updated "system" icon that conveys something about light/dark/theme stuff.
* Remove the theme toggle text and only show the text in the expanded menu.

Resources:
* https://ux.redhat.com/ - toggle group with cool icons
    <img width="313" height="106" alt="Image" src="https://github.com/user-attachments/assets/48e77dc8-5053-4fbd-98b1-03800539b4af" />
* https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_security_for_kubernetes/4.8#Top%20use%20cases - basically the menu we want to try in a POC.
    <img width="266" height="212" alt="Image" src="https://github.com/user-attachments/assets/5d47cf02-5663-4b87-83b0-b6c85daea2fd" />

PR #4710, opened 2025-07-18
fix(docs-framework): updates to theme switcher menu/icons

mcoker:
fixes https://github.com/patternfly/patternfly-org/issues/4709

cc @andrew-ronaldson - feel free to push updates to this PR as you see fit!


andrew-ronaldson (commented):
The only thing I would update is adding a `pf-v6-u-font-size-md` modifier so the icons are bigger in the button. The icon size in the menu is fine so just the menu toggle.

PR #4725, opened 2025-07-30
docs(table): Updates design guidelines for v6.

### LINKED ISSUE #4687: Docs - Update table guidelines.
Follow-up to https://github.com/patternfly/patternfly-org/pull/4676

A variety of general edits are needed. Just full audit and review the content for grammar, spelling, consistency/accuracy, linking, and structure.

edonehoo:
closes #4687

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/design-guidelines/components/table/table.md
+++ b/packages/documentation-site/patternfly-docs/content/design-guidelines/components/table/table.md
@@ -6,123 +6,128 @@ section: components
 import '../components.css';
 
 ## Elements
-The elements mentioned below are similar for a table with compact or default spacing. This example shows a table with a compact spacing.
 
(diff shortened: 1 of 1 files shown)


bekah-stephens (changes requested):
The example screenshots for compact and default spacing are the same content, just different row heights. I think this was fine out of context, but they don't really match the "Example" text (the 3 bullet points after each screenshot). I'm wondering if we should at least update the "Default spacing" screenshot to reflect what those bullet points say so that it's visually apparent why you would want to use that spacing over the compact.

edonehoo:
@bekah-stephens good point there! I tried to spice up the compact/default examples, can you lmk if they're more in line with what seems helpful? Or just lmk if you have other ideas for what they may look like!

bekah-stephens (approved):

PR #4729, opened 2025-08-01
docs(navigation): Updates design guidelines for v6.

### LINKED ISSUE #4666: Docs - Update navigation guidelines.
Follow-up to https://github.com/patternfly/patternfly-org/pull/4658#pullrequestreview-2936113208

- Replace spelled out numbers with numerics 
- Update grouped nav description to align with image
- Update two-level/three-level nav description and headers 
- Fly-out > Flyout? React docs call it Flyout (confirm which is right)
- Remove specifics in secondary horizontal nav description, which aren’t relevant to the image
- For secondary horizontal nav, add subheadings for each image (vertical vs horizontal primary)

edonehoo:
Closes #4666


packages/documentation-site/patternfly-docs/content/design-guidelines/components/navigation/img/nav-elements.svg
bekah-stephens, line 1: should we have the hamburger menu on the horizontal navigation example? the annotation and use is for vertical navigation

Issue #4622, opened 2025-05-16
Docs - Update spacers guidelines


edonehoo:
Expand guidance with recent changes 

From Lucia (reach out for more details) "We did recently add semantic spacer tokens for global page chrome inset and global gutters. we could add that info on our spacing page to include that and provide info on how they are applied/link to the layouts & page component."

PR #4730, opened 2025-08-01
docs(spacers): Documents recently added layout spacers.

edonehoo:
Works towards #4622

But remaining pr is needed to update tokens table description to close the og issue


mcoker (approved):
LGTM! Left a comment about layouts since "layout" as a concept are different from "layouts" as a product offering (our grid, gallery, flex, stack, etc layouts), in case that wasn't clear.

lboehling (approved):
nice!

PR #4731, opened 2025-08-04
feat: Highlight new animations article

nicolethoen:


edonehoo (approved):
if we could also fix the blurry images in this pr that would be nice, but this looks good either way

nicolethoen:
I don't know the answer to the blurriness right now - I tried to update it to the original size of the image in the article and it looked blurrier 😢

PR #4735, opened 2025-08-06
add cdn to html

202420505:


dlabaj:
Not adding any cdn link.

Issue #4677, opened 2025-06-24
Implement PatternFly Website Navigation Redesign


edonehoo:
**Problem statement** 
UX Designers and Developers struggle to consistently, successfully, and confidently locate information on PatternFly.org. Data from the PatternFly Navigation Study shows that they are successful in locating information 70.7% of the time, after 27 seconds on average, with a confidence rating of 3.62/5, and 52% of successes being indirect.  

**Goal / Impact statement**
A new PF.org navigation will make it easier for our users to confidently use the main site navigation and page context to successfully find the resources they need. It could improve task finding success rates to >76%, with a finding time of <20 seconds, and a confidence rating of >4/5. 
An updated PF.org navigation structure will better support content design and docs scalability. 

**Acceptance criteria**
A new PF.org navigation is merged to staging. 
(if possible) The org file structure in GitHub is aligned with the new nav structure.

**Risks**
If any technical limitations prevent planned nav changes.
We probably will want to clean up / align the file organization within GitHub to match the new nav structure (for easier contribution).

**Relevant links**
[Research study] https://issues.redhat.com/browse/UXDR-3908 

**Open questions**
Can we feasibly add nav section group headers and dividers? Are there accessibility/mobile considerations?

PR #4754, opened 2025-08-26
feat(sideNav): Adds support for grouped nav with headers and dividers.

edonehoo:
Closes 4750

This PR is only meant to add the functionality, actually adding in nav groups (I've tested with success locally) will come in a follow up pr for https://github.com/patternfly/patternfly-org/issues/4677?issue=patternfly%7Cpatternfly-org%7C4753

### DIFF
--- a/packages/documentation-framework/components/sideNav/sideNav.js
+++ b/packages/documentation-framework/components/sideNav/sideNav.js
@@ -4,14 +4,21 @@ import {
   Label,
   Nav,
   NavList,
+  NavGroup,
   NavExpandable,
   PageContextConsumer,
   capitalize,
   Flex,
-  FlexItem
+  FlexItem,
+  Divider
 } from '@patternfly/react-core';
 import { css } from '@patternfly/react-styles';
 import { Location } from '@reach/router';
+
+const DIVIDER_STYLES = {
+  marginTop: 'var(--pf-t--global--spacer--md)',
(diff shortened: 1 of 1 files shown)


edonehoo:
Feels like cursor did some extra things -- I asked it to review its own work and make sure any efficiency/accessibility problems are addressed, but lmk if it went to far. Tested with the nav array and it seems to still work fine

wise-king-sullyman (commented):
I would maybe re-approach this in a way that doesn't require as much logic being added in, i.e. having the grouped nav sections get defined as something like 

```
sideNavItems: [
      { title: 'Learn', items: [
        { section: 'About us' },
        { section: 'Releases' },
        { section: 'get-started' },
      ], hasDivider: true},
      { title: 'Design and develop', items: [
        { section: 'components' },
        { section: 'patterns' },
        { section: 'extensions' },
        { section: 'foundations' },
        { section: 'accessibility' },
        { section: 'AI' },
        { section: 'content' },
        { section: 'developer-guides' },
      ], hasDivider: true},
      { title: 'Connect', items: [
        { section: 'get-involved' },
        { section: 'get-help' },
      ]},
    ],
```
    
That would let us just handle it on the `sideNav.js` side by sorting them with something like 

```
const groupedItems = navItems.filter((entry) => entry.title);
const ungroupedItems = navItems.filter((entry) => !entry.title);
```

WDYT?

PR #4758, opened 2025-08-28
feat(org): pull in react & core docs to astro site

kmcfaul:
Closes https://github.com/patternfly/patternfly-doc-core/issues/81 & https://github.com/patternfly/patternfly-doc-core/issues/80

 Need to open issues for the following:
- https://github.com/patternfly/patternfly-doc-core/issues/145. Navigation sidebar doesn't `overflow: scroll` properly (the sidebar astro island doesn't always get the correct css applied)
- https://github.com/patternfly/patternfly-doc-core/issues/144. Tabs need to be supported in other sections like `Patterns`/`Layouts` . Right now multiple nav items are being created when they should be a single page with tabs.
- Some react docs are still using inline components so will not be rendered (Alert, TBD more)
- Some md (now mdx) files use inline components that will not be rendered (getting-started/upgrade, TBD more)
    - Might need to provide a suite of wrapped components beyond `LiveExample` for `mdx` usage like `Button` / `Alert`
    - Specifically for `getting-started/upgrade`, icons (or other JSX objects) passed as props is a no-go pattern in astro. We should figure out the best way to address this or work around it.
- Would be nice to figure out how to host a PR preview for the astro site.


mcoker:
Just a reminder for myself - to build/start this locally, run

```
yarn
yarn build:doc-core
yarn start:doc-core
```

PR #4761, opened 2025-08-28
feat(theme): Add high contrast switcher feature flag

wise-king-sullyman:
Followup from https://github.com/patternfly/patternfly-org/pull/4749, adds a feature flag for the HC switcher in case we don't want to publish the website with it yet but want to be able to use it for development and such


mcoker (approved):
Tested locally, works great!

PR #4766, opened 2025-09-03
fix: enable HC switcher, add beta label

### LINKED ISSUE #4759: How to show high contrast options on org
Since we're releasing as alpha/beta

1) Do we want to include the HC toggle in the theme switcher on org?
  * If not, should there be any way to enable HC on org, or would you need to go to one of our dev workspaces to toggle HC? (assuming we add it to other dev workspaces)
  * If so, should we label it as beta/alpha? 
2) Related, does anyone have objections to adding it to our dev workspaces in at least core and react?
3) Do we want to have any sort of POC/demo that's clearly marked alpha/beta?

Anything else to consider with our upcoming release? cc @lboehling

### LINKED ISSUE #4765: Add beta label to high contrast section of theme switcher
To match other beta labels, use the blue/filled/compact label beside the group title in the menu.

mcoker:
fixes https://github.com/patternfly/patternfly-org/issues/4765
fixes https://github.com/patternfly/patternfly-org/issues/4759

Questions:
* Is there a way I can pass this text + component to `<SelectGroup label="...">` and have it generate the menu title element/classes so I don't have to hard code it? According to the [<MenuGroup> code](https://github.com/patternfly/patternfly-react/blob/7a5d71d36e67a772c9c4990f80597d1ea602fa88/packages/react-core/src/components/Menu/MenuGroup.tsx#L33-L39), I assumed if I pass a function, it would  include the function within `<Wrapper />` but it doesn't appear to.
* I'm not sure what heading level to use - `<h2>` seemed appropriate considering it's a top-level heading for the page, not necessarily related to the content (cc @thatblindgeye)
* Should we add a label for the dark theme stuff? This is what it might look like (cc @lboehling) 
    Screenshot 2025-09-02 at 8 11 06 PM

### DIFF
--- a/packages/documentation-framework/components/themeSelector/themeSelector.js
+++ b/packages/documentation-framework/components/themeSelector/themeSelector.js
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
 import {
+  Content,
   Select,
   SelectGroup,
   SelectList,
@@ -11,8 +12,11 @@ import {
   ToggleGroupItem,
   Icon,
   Divider,
-  Label
+  Label,
+  Popover,
+  Button
 } from '@patternfly/react-core';
+import { HelpIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
 import { useTheme, THEME_TYPES } from '../../hooks/useTheme';
 
 const SunIcon = (
@@ -47,8 +51,37 @@ const DesktopIcon = (
 );
 
 const HCGroupLabel = () => {
+  const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+
   return (
-    <h2 className="pf-v6-c-menu__group-title">High contrast &nbsp;<Label color="blue" isCompact>Beta</Label></h2>
+    <h2 className="pf-v6-c-menu__group-title">
+      High contrast{' '}
+      <Popover
+        headerContent={"Under development"}
+        headerComponent="h3"
+        bodyContent={
+          "We are still working to add high contrast support across all PatternFly components and extensions. This beta allows you to preview our progress."
+        }
+        footerContent={
+          <Button icon={<ExternalLinkAltIcon />} component="a" isInline variant="link" href="https://medium.com/patternfly" target="_blank">
+            Learn more (replace with link to blog post)
+          </Button>
+        }
+        isVisible={isPopoverOpen}
+        shouldClose={() => setIsPopoverOpen(false)}
+        aria-label="More info about high contrast"
+        appendTo={() => document.body}
+      >
+        <Button
+          variant="plain"
+          hasNoPadding
+          icon={<HelpIcon />}
+          onClick={() => setIsPopoverOpen(!isPopoverOpen)}
+          aria-label="High contrast help"
+        />
+      </Popover>{' '}&nbsp;
+      <Label color="blue" isCompact>Beta</Label>
+    </h2>
   )
 };
 
@@ -116,7 +149,7 @@ export const ThemeSelector = ({ id }) => {
         preventOverflow: true
       }}
     >
-      <SelectGroup>
+      <SelectGroup label={'Color scheme'} labelHeadingLevel="h2">
         <SelectList aria-label="Light/Dark theme switcher">
           <SelectOption value={colorModes.SYSTEM} icon={DesktopIcon} description="Follow system preference">
             System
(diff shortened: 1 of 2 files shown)


thatblindgeye (commented):
- For the heading level, Austin's original PR I think had it as an h1. I'm not sure if any heading level will be great - h1 is slightly better because then it doesn't come across as a sub-section of some other h1 on the page (on component pages for example, "High contrast" comes off as a sub-section of the component's CSS variables heading). An h2 might work if there were visible text above the dark/system/light modes area since that renders an h1, just right now an empty h1 is being rendered because we don't pass a label in. Visually, though, we've had these menu groups with headings/labels act as independent groups rather than "This is the label/heading for the entire menu, and this label/heading is a sub-section".
- For the function bit, I think the React implementation isn't actually setup properly to accept a function? It's typed as only a React Node, but even then we aren't rendering it correctly if a React Node is passed in since the check to wrap stuff in that title class only accepts a function or string. Do we recall the reason for only wrapping the label in this menu groujp class in this way? Passing a Title in doesn't provide the same styling as just a string so either that can't be it or we missed updating some CSS to take that into account.

edonehoo:
My 2 cents: what if we add a help popover beside the beta label that either links to Medium article or incoming Theming docs? Like this: 

image

image

Maybe with something like this copy: 

> Under development
> We are still working to add high contrast support across all PatternFly components and extensions. This beta allows you to preview our progress.
> 
> [Learn more](Medium article)

I think it's clear enough without it, but would help link to additional context more conveniently 🤷‍♀️

packages/documentation-framework/components/themeSelector/themeSelector.js
```diff
@@ -45,6 +50,41 @@ const DesktopIcon = (
   </svg>
 );
 
+const HCGroupLabel = () => {
+  const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+
+  return (
+    <h2 className="pf-v6-c-menu__group-title">
+      High contrast{' '}
+      <Popover
```
thatblindgeye, line 59: This may depend on discussion of my PR comment above, but if we do decide to keep heading levels within this menu, could we structure this in a way so that the popover button and beta label aren't part of the heading? This is what the heading ends up looking like in the a11y tree:

image

Which feels odd for some reason. I think if it was just the High Contrast text and the Beta label it'd look a little better.

thatblindgeye (commented):
File comment below will depend on this comment for conversation:

I'm wondering if we should allow div and span to be passed as the "title/heading" elements for MenuGroup. Because those are all being appended to the document body, the heading levels are tough to land on which level they should be, or it may not end up making sense.

For example, the Popover heading level is an h3, but it doesn't have any h2 ancestors (nor h1), so it ends up being a random h3 within the page. This Popover may actually work as an h1 being passed as the heading element instead, because at least with VoiceOver, wwhen traversing the rotor menu that's the only heading that shows up since the Popover has the modal attribute. Can confirm in NVDA the Popover heading is also treated as the only heading accessible when trying to navigate with keyboard shortcuts to next/previous headings, so there is a case for this being an h1.

For the MenuGroups, however, their h2 level really only works on the main page of Org since they act as subsections of the h1 there, but on component pages they end up coming off as subsections of the component itself. Thinking about whether it would make more sense to make the current headings in the theme menu plain divs, and just link them to the Select groups via aria-labelledby so that the Select lists are properly named.

packages/documentation-framework/components/themeSelector/themeSelector.js
```diff
@@ -104,7 +144,7 @@ export const ThemeSelector = ({ id }) => {
       shouldFocusToggleOnSelect
       onOpenChangeKeys={['Escape']}
     >
-      <SelectGroup>
+      <SelectGroup label={'Color scheme'} labelHeadingLevel="h2">
```
thatblindgeye, line 147: Also depending if anything changes from above comment, but since we're giving a visible label here let's also update the aria-label of the SelectList to match this visible label.

thatblindgeye (approved):

edonehoo (changes requested):
let's instead link to the incoming theming doc? re: https://github.com/patternfly/patternfly-org/pull/4799

edonehoo (approved):

lboehling (approved):
nice!

kmcfaul (approved):
lgtm. 

@dlabaj Is the Cloudflare build failing an issue?

dlabaj (approved):

PR #4775, opened 2025-09-10
chore: move theme switcher to bottom/left

mcoker:
Moves the theme switcher to the bottom/left by default on full page examples. There is currently an override from chatbot that moves that menu to the bottom/left that shows up on org and can now be removed - https://github.com/patternfly/chatbot/pull/675

### DIFF
--- a/packages/documentation-framework/components/example/example.css
+++ b/packages/documentation-framework/components/example/example.css
@@ -72,8 +72,8 @@
 
 .ws-full-page-utils {
   position: fixed;
-  right: 0;
-  bottom: 0;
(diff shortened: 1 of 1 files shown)


mcoker:
To test, you need to disable these `!important` styles that are coming from chatbot. Just verify that the menu is still in the bottom left after disabling these two styles

Screenshot 2025-09-10 at 9 20 20 AM

kmcfaul (approved):

thatblindgeye (approved):

wise-king-sullyman (approved):

PR #4779, opened 2025-09-11
Potential theme selector fix

srambach:
@wise-king-sullyman I tried seeing if Cursor could see a problem with the theme switcher that could be causing the behavior where the selector shows "Display" but it's really set to light or dark. This is what it came up with - full answer below. I can't verify the solution (I can't reproduce it locally and also don't know React 😝 ) but was hoping you'd take a 👀 please and see if any of it looks useful. I tried adding CPU/network throttling in case that triggered it if it's a timing thing, but no luck.

------------------------
Looking at the code, I can see a potential bug in the theme selector logic. Let me analyze the issue:

[1 tool called]

I found the bug! The issue is in the `getThemeDisplayText` function in `themeSelector.js`. Let me trace through the logic:

1. **In `useTheme.js`**: The default mode is `COLOR_MODES.SYSTEM` (which equals `'system'`)
2. **In `themeSelector.js`**: The `getThemeDisplayText` function has a bug in its switch statement

Looking at lines 66-75 in `themeSelector.js`:

```javascript
const getThemeDisplayText = (mode) => {
  switch (mode) {
    case colorModes.LIGHT:
      return 'Light';
    case colorModes.DARK:
      return 'Dark';
    default:
      return 'System';
  }
};
```

The problem is that `colorModes.SYSTEM` is `'system'` (from `useTheme.js` line 4), but the switch statement is checking against `colorModes.LIGHT` and `colorModes.DARK` only. When the mode is `'system'`, it falls through to the `default` case and returns `'System'` (with capital S), but the actual stored value is `'system'` (lowercase).

However, there's another issue: the `selected` prop in the Select component is using `themeMode` directly, but the SelectOption values are using `colorModes.SYSTEM`, `colorModes.LIGHT`, etc. Let me check if there's a mismatch:

[2 tools called]

Now I see the issue! The problem is that `colorModes` comes from the `useTheme` hook, which returns `theme.modes` (line 173 in `useTheme.js`). This means `colorModes.SYSTEM` should equal `'system'`, but there might be a timing issue or the values aren't being properly passed.

Let me check what `colorModes` actually contains by looking at how it's used:

[1 tool called]

I found the bug! The issue is that `colorModes` in the theme selector is `theme.modes` which is `COLOR_MODES` from the hook, but the `selected` prop is using `themeMode` which is the actual stored value (`'system'`, `'light'`, `'dark'`). 

The Select component expects the `selected` value to match one of the `SelectOption` values. Since `SelectOption` values are using `colorModes.SYSTEM` (`'system'`) but the `selected` prop is using the raw `themeMode` value, there's a mismatch.

Let me fix this:

[1 tool called]

Actually, let me also check if there's an issue with the `getThemeDisplayText` function. The issue might be that when `themeMode` is undefined or null during initial load, it's not handling it properly:

[1 tool called]

Let me also fix the `getThemeIcon` function to be consistent:

[1 tool called]

Now let me check for any linting errors:

[2 tools called]

Perfect! I found and fixed the bugs in the theme selector. Here's what was causing the issue:

### 1. **Selected Value Mismatch**
**Problem**: The `selected` prop was using `themeMode` directly, but during initial load, `themeMode` might be `undefined` or not yet initialized, causing the Select component to not properly match with the SelectOption values.

**Fix**: Added fallback to `colorModes.SYSTEM`:
```javascript
selected={themeMode || colorModes.SYSTEM}
```

### 2. **Switch Statement Logic**
**Problem**: The `getThemeDisplayText` and `getThemeIcon` functions had incomplete switch statements that didn't explicitly handle the `colorModes.SYSTEM` case, relying only on the `default` case.

**Fix**: Added explicit `case colorModes.SYSTEM:` before the `default` case in both functions:
```javascript
case colorModes.SYSTEM:
default:
  return 'System'; // or DesktopIcon
```

## Root Cause Analysis

The issue you described (toggle showing "System" but Light being selected) was happening because:

1. **Initial Load**: During SSR or initial client-side render, `themeMode` might be `undefined`
2. **Display Logic**: The toggle was showing "System" because the display text function was working correctly
3. **Selection Logic**: But the Select component's `selected` prop was `undefined`, so it wasn't properly matching any SelectOption value
4. **Visual Mismatch**: This created the visual inconsistency where the toggle showed one thing but the dropdown selection showed another

## What the Fixes Accomplish

- **Consistent State**: The Select component now always has a valid selected value
- **Proper Fallback**: If `themeMode` is undefined, it defaults to `colorModes.SYSTEM`
- **Explicit Handling**: Both display functions now explicitly handle the SYSTEM case
- **Visual Consistency**: The toggle display and dropdown selection will now always match

This should resolve the issue where the page loads with the selector showing "System" but the dropdown shows "Light" as selected.

### DIFF
--- a/packages/documentation-framework/components/themeSelector/themeSelector.js
+++ b/packages/documentation-framework/components/themeSelector/themeSelector.js
@@ -69,6 +69,7 @@ export const ThemeSelector = ({ id }) => {
         return 'Light';
       case colorModes.DARK:
         return 'Dark';
+      case colorModes.SYSTEM:
       default:
         return 'System';
     }
@@ -80,6 +81,7 @@ export const ThemeSelector = ({ id }) => {
         return SunIcon;
       case colorModes.DARK:
         return MoonIcon;
+      case colorModes.SYSTEM:
       default:
         return DesktopIcon;
     }
@@ -89,7 +91,7 @@ export const ThemeSelector = ({ id }) => {
     <Select
       id={id}
       isOpen={isThemeSelectOpen}
-      selected={themeMode}
+      selected={themeMode || colorModes.SYSTEM}
       onSelect={handleThemeChange}
       onOpenChange={(isOpen) => setIsThemeSelectOpen(isOpen)}
       toggle={(toggleRef) => (
@@ -103,6 +105,11 @@ export const ThemeSelector = ({ id }) => {
       )}
       shouldFocusToggleOnSelect
       onOpenChangeKeys={['Escape']}
+      popperProps={{ 
+        position: 'right',
+        enableFlip: true,
+        preventOverflow: true
+      }}
     >
       <SelectGroup>
         <SelectList aria-label="Light/Dark theme switcher">
--- a/packages/documentation-framework/components/themeSelector/themeSelector.js
+++ b/packages/documentation-framework/components/themeSelector/themeSelector.js
@@ -69,7 +69,6 @@ export const ThemeSelector = ({ id }) => {
         return 'Light';
       case colorModes.DARK:
         return 'Dark';
-      case colorModes.SYSTEM:
       default:
         return 'System';
     }
@@ -81,7 +80,6 @@ export const ThemeSelector = ({ id }) => {
         return SunIcon;
       case colorModes.DARK:
         return MoonIcon;
-      case colorModes.SYSTEM:
       default:
         return DesktopIcon;
     }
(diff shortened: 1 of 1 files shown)


rebeccaalpert (approved):
I think this is safe given the approvals here: https://github.com/patternfly/patternfly-org/pull/4787. I'll close mine and assign the issue to you. :)

PR #4787, opened 2025-09-15
fix(ThemeSelector): Adjust popperProps

rebeccaalpert:
Prevent theme selector from overflowing container. OK if we want to drop in favor of https://github.com/patternfly/patternfly-org/pull/4779; Sarah has the fix there too. I'd just want to assign the ticket to her so she gets credit.

### DIFF
--- a/packages/documentation-framework/components/themeSelector/themeSelector.js
+++ b/packages/documentation-framework/components/themeSelector/themeSelector.js
@@ -103,6 +103,7 @@ export const ThemeSelector = ({ id }) => {
       )}
       shouldFocusToggleOnSelect
       onOpenChangeKeys={['Escape']}
+      popperProps={{ position: 'right', preventOverflow: true, enableFlip: true }}
     >
       <SelectGroup>
         <SelectList aria-label="Light/Dark theme switcher">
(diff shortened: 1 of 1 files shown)


kmcfaul (approved):

packages/documentation-framework/components/themeSelector/themeSelector.js
```diff
@@ -103,6 +103,7 @@ export const ThemeSelector = ({ id }) => {
       )}
       shouldFocusToggleOnSelect
       onOpenChangeKeys={['Escape']}
+      popperProps={{ position: 'right', preventOverflow: true, enableFlip: true }}
```
mcoker, line 106: Should we default to right when there's probably room to show it on the left?

mcoker: Oh sorry I was specifically referring to the one in the masthead, which is the most prominent one. On org it's left aligned currently, but right aligned in this PR:

Org
Screenshot 2025-09-16 at 2 56 31 PM

This PR
Screenshot 2025-09-16 at 2 56 54 PM

rebeccaalpert: Right! The release dropdown is right positioned already so I made it match. It looks a little funky to me otherwise, but I ultimately don't really care.

Screenshot 2025-09-16 at 4 10 47 PM

mcoker: Oh gotcha! I didn't realize you were comparing it to the version switcher. Checked with @lboehling and she's good with this change and prefers right aligned things here in general (search bar slides out from the right, too) 🚀

PR #4790, opened 2025-09-16
fix(theming): removed theme usage in footer

thatblindgeye:
This fix helps resolve a bug where having Light or Dark selected as the color scheme would be ignored when updating the OS level setting.


wise-king-sullyman (approved):

mcoker:
> @andrew-ronaldson Updated to apply a height of 32px to the svg, which is just about what the old img itself was in height (plus like 0.112 in change or so).

This is cool but a tiny nit. Looks like the old logo was using `width="145px"` to get its shape (the height was discarded, probably because `img` gets `height: auto` from the CSS). Setting `width="145px"` might get you closer to the old image size? 

https://github.com/patternfly/patternfly-org/blob/22946798a156249fac50b57f7d0a7ce8208d59a4/packages/documentation-framework/components/footer/footer.js#L297-L302

thatblindgeye:
Looks like this has been resolved by #4817

PR #4793, opened 2025-09-18
Fix narrow code editor width issue in Label component demos

### LINKED ISSUE #4777: Bug - Label - Code editor is too small
**How do you reproduce the problem?**
The code editor is too small in https://www.patternfly.org/components/label/react-demos making readability difficult. 

**Expected behavior**
The code editor should be wide enough just like other examples on the website.

**Screenshots**
Current Behaviour
Image

Expected Behaviour
Image

nicolethoen:
## Summary
- Fixes narrow code editor width issue on Label component React demos page
- Ensures the `.ws-code-editor` element takes full width of its container
- Adds CSS rules to override any restrictive width constraints

## Changes
- Added CSS rules to set `width: 100%` and `max-width: none` for `.ws-code-editor`
- Added specific rule for nested `.pf-v6-c-code-editor` to ensure full width
- Applied changes to `packages/documentation-framework/components/example/example.css`

## Test plan
- [x] Build the project successfully (`yarn build`)
- [x] Verify CSS changes are applied to the documentation framework
- [ ] Manually test on Label component React demos page to confirm wider code editor
- [ ] Check that other component demo pages are not adversely affected

Fixes #4777

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/documentation-framework/components/example/example.css
+++ b/packages/documentation-framework/components/example/example.css
@@ -2,6 +2,16 @@
   --ws-code-editor--tooltip--MaxWidth: 16ch;
 }
 
+/* Fix for narrow code editor width issue #4777 */
+.ws-code-editor {
+  width: 100%;
+  max-width: none;
+}
+
+.ws-code-editor .pf-v6-c-code-editor {
+  width: 100%;
+}
+
 .ws-code-editor:not(.ws-example-code-expanded) > .pf-v6-c-code-editor__header::before {
   --pf-v6-c-code-editor__header--before--BorderBottomWidth: 0;
 }


mcoker (commented):
This is one way to fix it, but IMO the issue is that the main content area's width is defined by the width of the content and not a consistent width. I think it should be a consistent width, regardless the size of components and docs and stuff in the main content area.

You can see here that the width of the main content is `max-width: min(100%, 825px)`, which effectively means `width: 100%; max-widht: 825px;`. 

That would look a lot better if it were `width: 825px; max-width: 100%;`, or `width: min(100%, 825px)`

Also this may overlap with changes @srambach is making in https://github.com/patternfly/patternfly-org/pull/4802

PR #4802, opened 2025-09-25
fix(TOC): fixes toc showing incorrectly at exactly 1450px

### LINKED ISSUE #4747: Bug - PF.org table of contents responsiveness
At certain window sizes (for me at least) our website table of contents is moving to the bottom of the page content instead of collapsing into the expandable section at the top of the page. It's a very small blip, but if you happen to have your window at that size its confusing

Using Chrome on Mac
Happened for me at 1461px x 1114px

Image

srambach:
Fixes #4747 

This fixes the small blip where the menu shows incorrectly at exactly 1450px. 

One known problem: if the TOC is expanded when in the dropdown (mobile) mode, then the `.pf-m-expanded` class will be added. If the window is resized to be larger than 1450px, the TOC will be displayed vertically - however, that class is never removed. So the next time the user resizes the window to be smaller than 1450px, the expanded class is still there and the menu will show as expanded.

### DIFF
--- a/packages/documentation-framework/components/tableOfContents/tableOfContents.css
+++ b/packages/documentation-framework/components/tableOfContents/tableOfContents.css
@@ -46,7 +46,7 @@
   display: none;
 }
 
-@media (min-width: 1451px) {
+@media (min-width: 1450px) {
   .ws-toc {
     width: 260px;
     max-height: calc(100vh - 76px);


srambach:
@rebeccaalpert You were right about the other media query; that fixed the difference in spacing. All the numbers were off by 1px and that's based on our 2xl breakpoint because the jumplinks have the `.pf-m-non-expandable-on-2xl` class; 2xl is 90.625rem, or 1450px.

I also adjusted the offset to add a condition  because our page also reconfigures slightly at the md breakpoint of 768px, so we needed a different offset value between 768 and 1450.

Finally, @lboehling I tightened up the spacing around the TOC toggle from a -md block padding to a -sm because it seemed so dang big ;-) Happy to undo that if you don't like it!

One last comment..the background color on the mobile version makes it appear that you can click anywhere, but you can really only click on the button, which is only as long as the text. I toyed with making the button full width or turning off the background but left it as is, so if you think it would be good to change that and have ideas, lmk!

rebeccaalpert (approved):
She's a beaut!

PR #4897, opened 2025-12-17
chore(ReactFlow): updated guidance for a11y

thatblindgeye:


packages/documentation-site/patternfly-docs/content/AI/Compass/ReactFlow.md
```diff
@@ -48,4 +48,14 @@ The React Flow package `@xyflow/react` is compatible with PatternFly through cus
...
\ No newline at end of file
+```
+
+### Accessibility considerations
+
+While React Flow is built with accessibility in mind, you should always check that your implementation when paired with PatternFly is accessible via mouse, keyboard, and other assistive technologies like screen readers. Some things to keep in mind include:
+
+- Providing an accessible name to the `<ReactFlow>` component, if its default role of "application" is kept or if it is given another semantic role that requires an accessible name.
+- Due to the nature of React Flow and howw you are able to click and drag nodes, if your draggable nodes contain other actions inside of them (kebab toggles, links, other action buttons), you should ensure those actions are large enough to prevent misclicks.
+  -  [WCAG 2.5.8 - Target size (Minimum)](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum) requires at least 24z24 CSS pixels. While they also state exceptions, because a React Flow node itself can be clicked for some action, you should aim for sticking to this minimum, especially if implementing an action where clicking a node quickly can open a details drawer.
```
edonehoo:
```suggestion
  -  [WCAG 2.5.8: Target size (Minimum)](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum) requires targets to be at least 24 by 24 CSS pixels. While there are exceptions, you should still aim for this minimum because a React Flow node itself can be clicked for some actions. This is especially important for an action like opening a details drawer upon a quick click to a node.
```

packages/documentation-site/patternfly-docs/content/AI/Compass/ReactFlow.md
```diff
@@ -48,4 +48,14 @@ The React Flow package `@xyflow/react` is compatible with PatternFly through cus
...
+- In general ensure that any actions that can be taken via mouse only can be taken via keyboard as well, whether that be actions inside a node or being able to drag a node around to reposition it.
+  - Dragging nodes via keyboard should be built into React Flow itself out of the box, but always double check in case.
```
edonehoo:
```suggestion
  - Dragging nodes via keyboard should be built into React Flow itself by default, but always double check to confirm.
```

PR #4915, opened 2026-01-23
docs(terminology): Add guidance around trademarked terms.

### LINKED ISSUE #4632: Add trademark guidance to UX writing guide.
Update UX writing guide to add relevant guidance for using other companies terms/trademarked phrases. Pull whatever from here is actually relevant for PF: 

"It is ok to use our competitors trademarks so long as:
It is used accurately to describe the other party’s own goods and
services, and all of the following conditions are met:
- It is not misleading;
- It compares goods or services meeting the same needs or intended
for the same purpose;
- It objectively compares one or more material, relevant, verifiable
and representative features of those goods or services,
- It does not create confusion in the marketplace between Red Hat and
a VMWare;
- It does not discredit or denigrate the trademarks, trade names or
other distinguishing signs of a competitor;
- It does not take unfair advantage of the trademark or other
distinguishing sign of a competitor; and
- It does not present goods or services as imitations or replicas of
goods or services bearing a trademark.
We may not use a competitor’s trademark as an attention-getting device. We
should use them in the body text only and avoid using them in headlines,
titles, and graphics. We always need permission to use a third party's
logo."

edonehoo:
Closes #4632 

Link to docs: https://pf-org--pr-4915-site.surge.sh/content-design/grammar/terminology

### DIFF
--- a/packages/documentation-site/patternfly-docs/content/content-design/grammar/terminology.md
+++ b/packages/documentation-site/patternfly-docs/content/content-design/grammar/terminology.md
@@ -24,7 +24,7 @@ When using trademarks:
 
 | Before | After | 
 | :---: | :---: |
-| Unlike the overcomplicated and overpriced Competitor®, our solution is both easy to use and affordable. | While Competitor® provides robust support for virtualization, our solution offers more integrated support for containerized environments. |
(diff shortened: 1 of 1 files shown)


edonehoo:
From Allison: Approved, but also with suggestion to add a Do / Do not example - so I'll add this

PR #4934, opened 2026-02-19
fix(theme switch): change unified theme class name

srambach:
I changed just the class name `pf-v6-theme-unified` to `pf-v6-theme-redhat` to match the core implementation.
I left "unified" as the label and variable names for the time being until a final name for the theme is set.


mcoker (approved):
LPTM

PR #4963, opened 2026-03-13
docs(components): Standardize approach to accessibility docs via design guidelines.

### LINKED ISSUE #4906: docs(design-guidelines): Consider alternative solution for components without design guidelines.
Right now, there are some components with no design guidelines, simply due to the fact that there really aren't any to offer. For example: https://www.patternfly.org/components/backdrop/design-guidelines. We received feedback via the org survey that it's confusing when there's just a link to the a11y docs under an Accessibility heading.

Do we need to have a design guidelines tab in these cases? Could we add an H1 to clarify this more? Or an alert? Or is it fine as is?

Explore a few ideas and take note of the components this applies to

edonehoo:
Closes #4906 

When a component's design guidelines only had the Accessibility section, deleted the design guidelines as they offered no valuable information.
Also ensured that all components with separate accessibility tab link to it from design guidelines should they exist.


phcox:
Just re-reviewed everything. Looks good!

Issue #4621, opened 2025-05-15
Update AI principles and guidelines.


edonehoo:
Iterate on https://github.com/patternfly/patternfly-org/pull/4545#event-17673659992 to incorporate additional suggestions from Kri:

- Could we offer a downloadable PDF/file instead of web-based checklist?
- Adjust: "Value and need: AI is aligned with user needs and values and will be continuously refined based on user feedback and ethical considerations."
    - it's a little generic. "Maybe a better checklist item is that the user need and value proposition has been documented for every AI product? Kind of like how in UXD we want to always have a problem statement that we're solving for before we start a project."
- When discussing "determine if AI adds value":
    - We aren't differentiating between gen AI and other AI. Perhaps offer guidance that distinguishes potential use cases for each, encouraging exploration? While gen AI still has accuracy issues, other structured types of AI may be better for tasks with lower risk tolerance, for ex.

Issue #4792, opened 2025-09-17
Docs - Update conversation design guidelines.


edonehoo:
Brand recently shared voice information for the openshift chatbot. I think we could incorporate some of this into our conversation design guidelines: [https://docs.google.com/document/d/16hzI3XFwwJK1X1-wM_4sPLm82Pql4TCzTHsnY3aW_uk/edit?tab=t.0#heading=h.eib3kop4eu95](https://docs.google.com/document/d/16hzI3XFwwJK1X1-wM_4sPLm82Pql4TCzTHsnY3aW_uk/edit?tab=t.0#heading=h.eib3kop4eu95) and [https://docs.google.com/spreadsheets/d/1z05E60i4rASU9TH1vWD0I17NO5WzcjgapLnFldKS_OA/edit?gid=1925215603#gid=1925215603](https://docs.google.com/spreadsheets/d/1z05E60i4rASU9TH1vWD0I17NO5WzcjgapLnFldKS_OA/edit?gid=1925215603#gid=1925215603) (examples)

Also, make sure images and content are updated with our current recs. Add more depth to any areas where we now have more knowledge.

---

**Jira Issue:** [PF-2709](https://redhat.atlassian.net/browse/PF-2709)

edonehoo:
closed by [docs(AI): Updates AI resources with latest guidance. #4967](https://github.com/patternfly/patternfly-org/pull/4967)

PR #4967, opened 2026-03-18
docs(AI): Updates AI resources with latest guidance.

edonehoo:
Closes https://github.com/patternfly/patternfly-org/issues/4621 and https://github.com/patternfly/patternfly-org/issues/4792


edonehoo:
@phcox ty for the review! I made those typo/copy changes

Do the images still look grainy to you? They look okay on my end, but there might be a browser/monitor difference that I could look into. 

Definitely agree re the robot icons, but they're not in Figma yet (unless I just can't find them), so this PR will need to wait for those to be uploaded. Similarly, I'll add in an image of the color options for the avatar version once those are finalized and available

phcox:
Looking at the images again, everything is fine except for the Figma screen captures. They seem pretty washed out and low-res.

All the text changes LGTM.

If you want to get this wrapped up, I bet we could find the robot images outside of Figma for you to use.

edonehoo:
@phcox added in the robot icon images! I also updated the Figma screenshots, but I suspect they'll still be blurry. I'm not sure how much more hi res I can get them. They're still readable on my end for whatever reason, so lmk how it looks for you. If it's that bad, I can just remove them and leave the text-based instructions

phcox:
Everything looks good to me. The images look fine - they're definitely readable, just not super crisp. I wouldn't remove them if I were you. And the new robot icons look good. Thanks!

PR #4996, opened 2026-04-28
ci(workflows): use org reusable issue and preview checks

nicolethoen:
Replace custom issue labeling with the org reusable workflow and gate Surge preview deploys behind the org team-membership reusable check.

Made-with: Cursor


.github/workflows-src/partials/build-astro.yml
```diff
@@ -9,5 +9,5 @@
   with:
     path: "packages/documentation-site/.cache"
     key: ${{ runner.os }}-${{ hashFiles('yarn.lock') }}
-- run: yarn ci:build:doc-core
+- run: yarn build:doc-core
```
wise-king-sullyman: What was the motivation behind this change?

nicolethoen: I believe it's because this src partial which should have been used to generate the workflows/pr-preview.yml did not match. 

So i'll double check with you - is the command in the workflows/pr-preview.yml correct? https://github.com/patternfly/patternfly-org/blob/9091ff1de25df42822a51b4a22a7dabc084c4f98/.github/workflows/pr-preview.yml#L58

If it's correct, then this change (made by cursor) is aligning them in case the workflows are ever regenerated from the src again.

If they are not both incorrect, I can revert this and fix the actual workflow file.

wise-king-sullyman: Ah, yeah it should be ci:build:doc-core for the astro builds

wise-king-sullyman (approved):

PR #5000, opened 2026-04-30
chore(ai design language): image change

andrew-ronaldson:
Replaced an image that showed users how to use rh icons. Also updated text to match description.


edonehoo (approved):
nice stuff

PR #5044, opened 2026-06-26
fix(docs-framework): fixes isStickyStuck rendering bugs

### LINKED ISSUE #5035: Bug - triggering dynamic tabs refreshes org page
To reproduce:
Open an example code editor on a component page and then scroll enough so the tabs' sticky state changes. The page will refresh, closing the editor and losing any active edits.

---

**Jira Issue:** [PF-4275](https://redhat.atlassian.net/browse/PF-4275)

mcoker:
fixes https://github.com/patternfly/patternfly-org/issues/5035

### DIFF
--- a/packages/documentation-framework/hooks/useIsStuck.js
+++ b/packages/documentation-framework/hooks/useIsStuck.js
@@ -2,16 +2,16 @@ import { useState, useLayoutEffect, useRef } from 'react';
 
 export const useIsStuck = (stickyElementId) => {
   const [isStuck, setIsStuck] = useState(false);
-  const lastValueRef = useRef(false);
+  const prevIsStuck = useRef(false);
 
   useLayoutEffect(() => {
     const scrollElement = document.getElementById('ws-page-main');
     const stickyElement = document.getElementById(stickyElementId);
 
     if (!scrollElement || !stickyElement) {
-      if (lastValueRef.current !== false) {
+      if (prevIsStuck.current !== false) {
         setIsStuck(false);
-        lastValueRef.current = false;
+        prevIsStuck.current = false;
       }
       return;
     }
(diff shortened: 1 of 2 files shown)


thatblindgeye (approved):

packages/documentation-framework/templates/mdx.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
```
jcmill: Is this `useMemo` being used? I could totally be blind.

mcoker: Yah right here https://github.com/patternfly/patternfly-org/blob/01ba63c5cf4d0ebd07bb6717ed8a2a537be32b95/packages/documentation-framework/templates/mdx.js#L28

packages/documentation-framework/hooks/useIsStuck.js
```diff
@@ -1,25 +1,37 @@
-import { useState, useLayoutEffect } from 'react';
+import { useState, useLayoutEffect, useRef } from 'react';
 
 export const useIsStuck = (stickyElementId) => {
   const [isStuck, setIsStuck] = useState(false);
+  const lastValueRef = useRef(false);
```
kmcfaul: Think this can be named `prevIsStuck` since it's being used as a check to see if a new state update is necessary.

packages/documentation-framework/templates/mdx.js
kmcfaul: I think `useMemo` and `React.memo` are different. You can remove `useMemo` or replace it with `memo` and use it directly instead of `React.memo`.

packages/documentation-framework/templates/mdx.js
```diff
@@ -336,6 +343,7 @@ export const MDXTemplate = ({ title, sources = [], path, id, componentsData }) =
               {sources
                 .map((source, index) => {
                   source.index = index;
+                  source.id = id;
```
kmcfaul: Not sure what this is added for, can probably be removed?

kmcfaul (changes requested):

packages/documentation-framework/templates/mdx.js
mcoker: oh my bad! Removed

jcmill (approved):
🥞

kmcfaul (approved):

PR #5046, opened 2026-06-30
feat(misc): separate CSSVariables into its own tab

kmcfaul:
Authored by Cursor AI, pending more review for cleanliness

### DIFF
--- a/packages/documentation-framework/routes.js
+++ b/packages/documentation-framework/routes.js
@@ -82,6 +82,7 @@ const sourceOrder = {
   'react-deprecated': 2.1,
   html: 3,
   'html-demos': 4,
+  'css-variables': 5,
   'design-guidelines': 99,
   'accessibility': 100,
   'upgrade-guide': 101,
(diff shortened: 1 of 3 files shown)


mcoker:
Some pages on org are so slow and this PR improves them quite a bit. A before/after of DOM nodes for a few components pages that are particularly slow

| component | before | after 
| :--- | :--- | :---
| tabs | 13579 | 5324 
| table | 18111 | 9469 
| button | 28847 | 3558

rebeccaalpert (commented):
This looks good to me.

For pages with more than one table like http://localhost:8003/components/alert/css-variables, we may want to add a TOC or something so it's easier to see, but that's a scope increase here.

rebeccaalpert (approved):

mcoker (commented):
I'm noticing extra filter boxes on some components. Looks like it's referencing the `cssPrefix` metadata in example files to build the table from? If so, we could just go through our example files and fix ones with incorrect values, and/or we could exclude the heading/table if the table is empty.

https://pf-org--pr-5046-site.surge.sh/components/button/css-variables#prefixed-with-pf-d-button
* https://github.com/patternfly/patternfly/blob/cfe5a9e62bbfaa7852e08f741a527fad63d2a1de/src/patternfly/demos/Button/examples/Button.md?plain=1#L4


https://pf-org--pr-5046-site.surge.sh/components/table/css-variables#prefixed-with-pf-c-table
* https://github.com/patternfly/patternfly-react/blob/1e942d80aa3715e39c370cb370c9aced072db4e2/packages/react-table/src/deprecated/components/Table/examples/Table.md?plain=1#L3

thatblindgeye (approved):

PR #5048, opened 2026-07-01
feat(ver): update to new versions

rebeccaalpert:


thatblindgeye (approved):

packages/documentation-site/package.json
```diff
@@ -18,20 +18,20 @@
     "screenshots": "pf-docs-framework screenshots"
   },
   "dependencies": {
-    "@patternfly/chatbot": "6.7.0-prerelease.4",
+    "@patternfly/chatbot": "6.6.0",
```
kmcfaul: Do we want to keep this as the 6.7 prerelease? Versions.json would also need this version specified if so.

rebeccaalpert: All set; thank you!

kmcfaul (approved):

PR #5075, opened 2026-07-14
fix(docs-framework): let examples shrink to fit window width

### LINKED ISSUE #5074: Bug - docs framework - examples do not shrink below 825px
If you go to any example page and shrink the viewport width, the examples never go below 825px due to the `min-width` here https://github.com/patternfly/patternfly-org/blob/6a7a04eccaf23ef9ed9f27940923add860e9f181/packages/documentation-framework/templates/mdx.css#L29-L33

---

**Jira Issue:** [PF-4441](https://redhat.atlassian.net/browse/PF-4441)

mcoker:
fixes https://github.com/patternfly/patternfly-org/issues/5074

In https://github.com/patternfly/patternfly-org/pull/4802, the example area was given the change below, which added `min-width: 825px`. That's keeping the examples from shrinking below 825px

Screenshot 2026-07-14 at 6 01 11 PM

I'm pretty sure that was to fix a bug on example pages that have narrow example content and no text that would make the example area fill the content width like this. AFAIK it only showed up on one example page, but I don't remember which example it was.

Screenshot 2026-07-14 at 6 06 33 PM

That said, `flex-grow: 1` fixes that specific bug ^, so we can revert back to the old max-width and fix the bug caused by the new min-width.

### DIFF
--- a/packages/documentation-framework/templates/mdx.css
+++ b/packages/documentation-framework/templates/mdx.css
@@ -27,7 +27,6 @@ p.pf-v6-c-content--p.ws-p {
 }
 
 .ws-example-page-wrapper {
-  min-width: 825px;
-  max-width: 1200px;
+  max-width: min(100%, 1200px);
   flex-grow: 1;
 }


thatblindgeye (approved):
Assuming it would have still been at exactly 1450px width, the original ToC bug doesnt seem to happen with this update (ToC goes from the collapsible view to the always-visible-on-the-right view when going from 1449px to 1450px width)

jcmill (approved):
LGTM 👾

Issue #5037, opened 2026-06-24
Themes & Modes: Org documentation


patternfly-jira-sync:
---

**Jira Issue:** [PF-4331](https://redhat.atlassian.net/browse/PF-4331)

patternfly-jira-sync:
Closed via Jira sync - Jira issue PF-4331 was closed.

PR #5102, opened 2026-08-14
docs(theming patterns): add best practices

andrew-ronaldson:
Will close to https://github.com/patternfly/patternfly-org/issues/5037

Quick link to the edits https://pf-org--pr-5102-site.surge.sh/foundations-and-styles/theming#theme-controls


andrew-ronaldson:
Notes on this draft:

- **Theming** needs its own section. We should specify that if you are a RH product then use Felt. If users are building open source they should default to PatternFly theme. Should we show **Theming** in products to end-users OR just show them Felt only.
- Too many instances of _Best practice_ in the doc. Consider a different title for Theming control section.
- Replace product examples with generic image. Update in Figma guidelines file.

packages/documentation-site/patternfly-docs/content/foundations-and-styles/styles/theming/theming.md
```diff
@@ -119,6 +119,49 @@ The following table outlines the availability and compatibility of PatternFly fe
...
+- Color scheme is the control most users will reach for, with theme and contrast mode as secondary, expanded options.
+- Your product doesn't already have a dedicated Settings or User preference page, and you don't want to introduce one solely for theming.
+
+### Theme control guidelines
+
+Regardless of which pattern you use, keep the following guidelines in mind:
+
+- **Default to the system setting:** Provide a "System" option for color scheme and contrast mode, and select it by default, so your product respects the user's OS- or browser-level preferences until they choose to override them.
+- **Persist user selections:** Store each user's choice, for example in `localStorage` or their account settings, so their preferences are remembered across sessions.
+- **Apply changes immediately:** Update the UI as soon as a user makes a selection, without requiring a separate save or submit action.
```
phcox, line 160: I agree with this guideline, but will there ever be a case when these controls are placed in a User Preferences modal that requires a Save or Apply action for other settings? If so, do we still want theme updates to happen immediately? I don't see a huge problem with it, but it could get weird that some things require a Save or Apply and others don't.

phcox (approved):
Looks good. Just made a single comment that I don't think should prevent this from being merged.

PR #4127, opened 2024-07-09
fix: clean up community and trainging card css

nicolethoen:
Closes https://github.com/patternfly/patternfly-org/issues/4021
Closes https://github.com/patternfly/patternfly-org/issues/4023

A major improvement to training cards: https://patternfly-org-pr-4127-site.surge.sh/training/html/
but open to more suggestions

PR #5103, opened 2026-08-25
chore(docs): Adjust spacing

### LINKED ISSUE #4257: Bug - Fix spacing issue that appears in React and HTML docs
New lines aren't formatted properly in these docs, looks like its due to p.pf-v6-c-content--p.ws-p in a css file

Example (there aren't any new lines when there should be a few paragraph breaks here):

image

---

**Jira Issue:** [PF-297](https://redhat.atlassian.net/browse/PF-297)

rebeccaalpert:
We used to determine the gutter based on whether items were component docs - this brings that back so we can adjust where margins are used.

I don't know the full history, but it was removed here: https://github.com/patternfly/patternfly-org/pull/4127/changes#diff-da135efaa2b01f501f1ac196874e6c6ae745345196f59a4581445639bf5384c4R94

I don't know if there's a reason why I shouldn't bring it back - please tell me!

Fixes https://github.com/patternfly/patternfly-org/issues/4257

Assisted-by: Cursor

### DIFF
--- a/packages/documentation-framework/components/example/example.css
+++ b/packages/documentation-framework/components/example/example.css
@@ -2,6 +2,10 @@
   --ws-code-editor--tooltip--MaxWidth: 16ch;
 }
 
+.ws-heading.ws-example-heading {
+  margin-block-end: -0.5rem;
+}
+
 .ws-code-editor:not(.ws-example-code-expanded) > .pf-v6-c-code-editor__header::before {
   --pf-v6-c-code-editor__header--before--BorderBottomWidth: 0;
 }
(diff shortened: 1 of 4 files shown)


kmcfaul (commented):
Looks good to me overall, only saw two spacing issues:

- Tables that precede other content don't have gap.
- Props table section doesn't have a gap if its preceded by something other than a code editor example. Maybe worth putting a one-off class on the Props table `h2`, or targeting the last `h2` on the page for additional padding.

kmcfaul (approved):
lgtm!

